PR #196, opened 2021-07-21
chore: improve bech32 validation in PrettyAddress

### LINKED ISSUE #189: defense-in-depth: data validation: bech32
just a nit noticed while doing #181 auditing -- not obvious whether it has security impact in practice.

we use bech32 w/ `oasis` prefix, but `isValidAddress` https://github.com/oasisprotocol/oasis-wallet-web/blob/22fad9dec2f3b8fe728925f8904f27fa4b923680/src/app/lib/helpers.ts#L24 or `assertValidAddress` https://github.com/oasisprotocol/oasis-wallet-web/blob/22fad9dec2f3b8fe728925f8904f27fa4b923680/src/app/state/transaction/saga.ts#L197 does not guard https://github.com/oasisprotocol/oasis-wallet-web/blob/22fad9dec2f3b8fe728925f8904f27fa4b923680/src/app/components/PrettyAddress/index.tsx#L15 so it's not obvious that there is no way to introduce malformed bech32 addresses, e.g., `oasis1...1foo` and `oasis1...1bar`, where the `...` are identical and valid bech32 characters, would be displayed sans the malformed suffix by `PrettyAddress` and are visually identical and confusing.  it seems likely that eventually malformed addresses would be caught, but it's not obvious that it will always be caught.

it would be better defense-in-depth if even UI code like `PrettyAddress` validated its preconditions about its arguments and threw an exception (which i think `bech32.decode` does), even though for error reporting / blame assignment the same check should be done as close to the source of the data as possible, so that the check in `PrettyAddress` is unreachable in normal execution.

Esya:
Fixes #189

### DIFF
--- a/src/app/components/PrettyAddress/index.tsx
+++ b/src/app/components/PrettyAddress/index.tsx
@@ -3,6 +3,7 @@
  * PrettyAddress
  *
  */
+import { isValidAddress } from 'app/lib/helpers'
 import React, { memo } from 'react'
 
 interface Props {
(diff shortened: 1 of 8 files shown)


src/app/components/PrettyAddress/index.tsx
```diff
@@ -13,17 +14,20 @@ interface Props {
...
-    const parts = addr.split('1')
+  if (!props.address || props.address === '') {
+    return <></>
+  }
+
+  if (isValidAddress(props.address)) {
+    const parts = props.address.split('1')
 
     const hrp = parts[0]
     const publicKey = parts[1].match(/.{1,4}/g)?.join(' ')
```
lukaw3d, line 26: This looks buggy if address contains "1", e.g. `oasis1qq1vzcvxn1js1unsch1me1xz1kr11vcasv1d1eq1`

lukaw3d: `'oasis1qq1vzcvxn1js1unsch1me1xz1kr11vcasv1d1eq1'.match(/(oasis1|.{1,4})/g).join(' ')` seems to work

lukaw3d: (updated)

lukaw3d: Oh, nevermind, Bech32 doesn't have `1` in data part

PR #259, opened 2021-09-14
feat: add main network

Esya:


src/config.ts
```diff
@@ -0,0 +1,14 @@
+export const config = {
+  mainnet: {
+    grpc: 'https://grpc.oasis.dev',
+    explorer: 'https://monitor.oasis.dev',
+  },
+  testnet: {
+    grpc: 'https://testnet.grpc.oasis.dev',
+    explorer: 'https://monitor.oasis.dev/api/testnet',
```
pro-wh, line 8: :information_source: original value was `https://testnet-api.oasismonitor.com`

pro-wh (approved):
to expand to having more than two networks, several ternary operators are replaced with a configuration dictionary.

however, as part of this PR, the URLs used are not only moved to a configuration file, they are also changed. the changes are consistent with a message I saw on our slack.

the code changes look fine too.

Issue #274, opened 2021-09-21
Hide or blur original mnemonic when confirming it (Firefox)


matevz:
When clicking *Open my Wallet* you are required to enter specific word from the mnemonic to prove that you really saved it somewhere safe. But currently this doesn't make sense, because the complete mnemonic is visible in the background anyway:
![image](https://user-images.githubusercontent.com/83672/134135282-ae3bb136-c1ca-4e8d-889a-acf73ba04bed.png)

I suggest we blur the original mnemonic in the background when showing the mnemonic confirmation dialog.

lukaw3d:
Firefox users :scream:

matevz:
It seems FF has some performance hickups with the `backdrop-filter` implementation and is currently disabled, but it will be implemented some time in the future: https://bugzilla.mozilla.org/show_bug.cgi?id=1578503

Let's just wait for mozilla folks to fix the issue in the browser and don't mess our code with this. Closing the issue.

matevz:
@lukaw3d thanks for your effort though :wink:

tjanez:
> It seems FF has some performance hickups with the `backdrop-filter` implementation and is currently disabled, but it will be implemented some time in the future: https://bugzilla.mozilla.org/show_bug.cgi?id=1578503

Ok, thanks for researching that.

> Let's just wait for mozilla folks to fix the issue in the browser and don't mess our code with this. Closing the issue.

I disagree. I think it's better to create a work-around. You never know when the upstream issue will be fixed :slightly_smiling_face:.

You can create an issue that will remove the "if firefox then ... else ..." conditional when the upstream issue is solved though.

PR #283, opened 2021-09-21
Blur mnemonic using filter (Firefox doesn't support backdropFilter)

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/274

### DIFF
--- a/src/app/pages/CreateWalletPage/index.tsx
+++ b/src/app/pages/CreateWalletPage/index.tsx
@@ -47,6 +47,10 @@ export function CreateWalletPage(props: CreateWalletProps) {
     }
   }, [createWalletActions, dispatch])
 
+  //@TODO Remove when firefox supports backdropFilter (used inside MnemonicValidation)
+  // https://github.com/oasisprotocol/oasis-wallet-web/issues/287
+  const blurMnemonicInFirefox = showConfirmation ? { filter: 'blur(5px)' } : {}
+
   return (
     <>
       {showConfirmation && (
@@ -57,7 +61,7 @@ export function CreateWalletPage(props: CreateWalletProps) {
         />
       )}
       <Grid gap="small" pad="small" columns={size === 'small' ? ['auto'] : ['2fr', '2fr']}>
-        <Box background="background-front">
+        <Box background="background-front" style={blurMnemonicInFirefox}>
           <MnemonicGrid mnemonic={mnemonic} />
           <Box margin="xsmall" pad="small" background="background-contrast" style={{ wordSpacing: '14px' }}>
             <strong data-testid="generated-mnemonic">{mnemonic.join(' ')}</strong>


src/app/pages/CreateWalletPage/index.tsx
```diff
@@ -57,7 +57,7 @@ export function CreateWalletPage(props: CreateWalletProps) {
         />
       )}
       <Grid gap="small" pad="small" columns={size === 'small' ? ['auto'] : ['2fr', '2fr']}>
-        <Box background="background-front">
+        <Box background="background-front" style={{ filter: showConfirmation ? 'blur(5px)' : 'none' }}>
```
lukaw3d, line 60: In Chrome, backdropFilter works [MnemonicValidation/index.tsx#L66](https://github.com/oasisprotocol/oasis-wallet-web/blob/e465051dd4b11a5d160f44f459cbace50c8665fb/src/app/components/MnemonicValidation/index.tsx#L66), so this extra filter will just double-blur mnemonic.

In Firefox this will only blur the mnemonic, because it is easiest to maintain bluring elements inside CreateWalletPage

matevz: `backdrop-filter` is also supported by firefox. Do you have any reasoning why this doesn't work? I wouldn't like to differentiate the codebase between the browsers at this stage...

lukaw3d: Doesn't seem to be supported
- https://caniuse.com/css-backdrop-filter
- https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter
- My firefox screenshot

lukaw3d: I don't think this is very important for ux, so I'm also happy with old behaviour

matevz: It seems FF has some performance hickups with the `backdrop-filter` implementation and is currently disabled, but it will be implemented some time in the future: https://bugzilla.mozilla.org/show_bug.cgi?id=1578503
Closing the issue.

tjanez (approved):

src/app/pages/CreateWalletPage/index.tsx
```diff
@@ -47,6 +47,9 @@ export function CreateWalletPage(props: CreateWalletProps) {
     }
   }, [createWalletActions, dispatch])
 
+  //@TODO Remove when firefox supports backdropFilter (used inside MnemonicValidation)
+  const blurMnemonicInFirefox = showConfirmation ? { filter: 'blur(5px)' } : {}
```
lukaw3d, line 51: Slightly refactored and added TODO

tjanez: Thanks :+1:

matevz: Maybe also add link to https://bugzilla.mozilla.org/show_bug.cgi?id=1578503 ?

Issue #326, opened 2021-10-15
Better handle pasting a mnemonic that contains a new line


tjanez:
For example, copy & pasting the following (test) mnemonic:
```
abandon abandon abandon abandon abandon abandon
abandon abandon abandon abandon abandon about
```

results in in new line being treated as "" and two consecutive mnemonic words concatenated into one:
![Screenshot from 2021-10-15 18-12-05](https://user-images.githubusercontent.com/2518544/137519441-ed53fee0-e880-4306-ac0a-ec0e51ab2dcd.png)

A users needs to manually find where the new line was and insert a space in between.

PR #331, opened 2021-10-20
Handle pasting mnemonic with newlines

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/326


lukaw3d:
> @pro-wh hope this doesn't break anyone's recovery for some weird reason

Hmmm, maybe it is too dangerous. I don't know enough standard tooling around ethereum for this. bip39 doesn't have a useful function.

`cat bip-0039/*.txt | grep -e "\\s"` is empty at least

PR #333, opened 2021-10-20
Keep pasted newlines in mnemonic

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/326
Supersedes https://github.com/oasisprotocol/oasis-wallet-web/pull/331

image


pro-wh:
now I wonder if this is more dangerous than the last one.

I checked if there's any cleaning done inside bip39, and it seems they only normalize weird unicode things, not whitespace. so in my reading, these whitespace differences will affect the seed.

I think there's more documentation 'out there' that treats the mnemonic as a list of words rather than a whitespace-specific string

pro-wh:
oh wait I hadn't even looked at the screenshot. this makes it a user error if they're not neatly space-separated

Issue #407, opened 2021-11-04
Support for multiple block explorers


gw0:
The Oasis Wallet currently depends on the external service `oasismonitor.com` (provided by a node operator). Recently, this service went down and this caused a lot of confusion/frustration (see https://github.com/oasisprotocol/oasis-wallet-web/issues/403).

To prevent such issues other wallets typically support multiple block explorers and let the user choose which one to use. I suggest that the Oasis Wallet should support at least 2 block explorer services.

tjanez:
Superseded by https://github.com/oasisprotocol/oasis-wallet-web/issues/693, implemented in https://github.com/oasisprotocol/oasis-wallet-web/pull/704.

Note that the backend choice is not exposed to the users but added as a build time option for wallet operators.

Issue #403, opened 2021-11-04
Wallet should still work if oasismonitor.com is down


matevz:
If the external service such as oasismonitor.com is down, wallet should still offer basic functionality like fetching the wallet balance, receiving and sending ROSE:
- [ ] Make a current full-page fatal error just a warning.
- [ ] Find out which functionalities don't require oasismonitor.com and gracefully hide functionalities that cannot work.
- [ ] Make errors regarding external services more descriptive. Current messages are cryptic and don't tell you that oasismonitor.com is down.

PR #408, opened 2021-11-04
AccountPage: Use wallet balance from oasis node directly instead of explorer

matevz:
Related to #403 and #407. Currently, the wallet balance is fetched from the block explorer and if the service went down, the balance was 0. This PR uses the balance reported by the local oasis node only.

### DIFF
--- a/src/app/pages/AccountPage/index.tsx
+++ b/src/app/pages/AccountPage/index.tsx
@@ -21,7 +21,7 @@ import { TransitionGroup } from 'react-transition-group'
 import { useAccountSlice } from '../../state/account'
 import { selectAccount } from '../../state/account/selectors'
 import { BalanceDetails } from '../../state/account/types'
-import { selectAddress, selectStatus, selectWallets } from '../../state/wallet/selectors'
+import { selectActiveWallet, selectAddress, selectStatus, selectWallets } from '../../state/wallet/selectors'
 import { ActiveDelegationList } from '../StakingPage/Features/DelegationList/ActiveDelegationList'
 import { DebondingDelegationList } from '../StakingPage/Features/DelegationList/DebondingDelegationList'
 import { ValidatorList } from '../StakingPage/Features/ValidatorList'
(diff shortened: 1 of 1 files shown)


src/app/pages/AccountPage/index.tsx
```diff
@@ -79,15 +79,15 @@ export function AccountPage(props: Props) {
   const selectedNetwork = useSelector(selectSelectedNetwork)
   const { active } = useSelector(selectTransaction)
   const wallets = useSelector(selectWallets)
+  const activeWallet = useSelector(selectActiveWallet)
 
   const balanceDelegations = stake.delegations.reduce((acc, v) => acc + Number(v.amount), 0)
-  const balanceDebondingDelegations = stake.debondingDelegations.reduce((acc, v) => acc + Number(v.amount), 0)
   const balance: BalanceDetails = {
-    available: account.liquid_balance ?? 0,
+    available: parseFloat(activeWallet?.balance.available ?? '') ?? 0,
```
pro-wh, line 86: here we go with the javascript type model
```suggestion
    available: parseFloat(activeWallet?.balance.available ?? '0'),
```

src/app/pages/AccountPage/index.tsx
```diff
@@ -79,15 +79,15 @@ export function AccountPage(props: Props) {
...
     delegations: balanceDelegations, //@TODO oasis-explorer : account.debonding_delegations_balance ?? 0,
-    debonding: balanceDebondingDelegations, //@TODO oasis-explorer : account.delegations_balance ?? 0,
-    escrow: account.escrow_balance ?? 0,
-    total: (account.liquid_balance ?? 0) + balanceDelegations + balanceDebondingDelegations,
+    debonding: parseFloat(activeWallet?.balance.debonding ?? '') ?? 0, //@TODO oasis-explorer : account.delegations_balance ?? 0,
+    escrow: parseFloat(activeWallet?.balance.escrow ?? '') ?? 0,
+    total: parseFloat(activeWallet?.balance.total ?? '') ?? 0,
```
pro-wh, line 90: can you confirm that these have the same meaning? it looks like activeWallet?.balance.xxx are delegations _to_ the account and debonding and total. had the previous code used the delegations _from_ the account?

pro-wh (commented):
I vaguely follow the flow back to a `nic.stakingAccount` call in saga.ts, although I don't really have experience with the frameworks used in this project.

the general/"available"/"liquid" balance handling seems fine

src/app/pages/AccountPage/index.tsx
pro-wh: ah the cypress tests are encountering this

matevz: Yeah, good point. I'll need to check thoroughly what those values are (to or from). Cypress tests are checking the total balance written in bold only and this also seems to be zero for some reason :shrug:

matevz: In the mean time, https://github.com/oasisprotocol/oasis-wallet-web/pull/408/commits/954a7e905a0e05b449df01e81824c5a4b4de88ef now computes `available` balance from the oasis-node and leave other balances taken from oasis monitor (cypress should pass). I think the `available` balance should always be taken from oasis-node since this balance will be checked, when you to send your tokens.

pro-wh (approved):
yeah the general balance should work

src/app/pages/AccountPage/index.tsx
```diff
@@ -79,11 +79,12 @@ export function AccountPage(props: Props) {
   const selectedNetwork = useSelector(selectSelectedNetwork)
   const { active } = useSelector(selectTransaction)
   const wallets = useSelector(selectWallets)
+  const activeWallet = useSelector(selectActiveWallet)
 
   const balanceDelegations = stake.delegations.reduce((acc, v) => acc + Number(v.amount), 0)
   const balanceDebondingDelegations = stake.debondingDelegations.reduce((acc, v) => acc + Number(v.amount), 0)
   const balance: BalanceDetails = {
-    available: account.liquid_balance ?? 0,
+    available: parseFloat(activeWallet?.balance.available ?? '0'), // Use local oasis-node available balance.
```
lukaw3d, line 87: This would need to check `activeWallet?.address === account.address`
- If user searches for address without opening a wallet: activeWallet is undefined
- If user opens a wallet and searches for different address, activeWallet is user's wallet

lukaw3d (changes requested):

lukaw3d:
I think this was effectively superseded by https://github.com/oasisprotocol/oasis-wallet-web/pull/737

PR #412, opened 2021-11-04
non-major features

pro-wh:
- a new codepath for #411. no automated tests. missing french translation
- failed block explorer requests were unhelpfully showing up as "undefined," an improvement to that

fixes #411

part of #403

PR #413, opened 2021-11-05
non-fatal explorer interaction: validators, account, transactions

pro-wh:
on top of #412 

for #403

### DIFF
--- a/src/app/pages/AccountPage/Features/TransactionHistory/index.tsx
+++ b/src/app/pages/AccountPage/Features/TransactionHistory/index.tsx
@@ -31,9 +31,15 @@ export function TransactionHistory(props: Props) {
 
   return (
     <Box gap="small" margin="none">
-      {transactionsError ? (
-        <p>Couldn't load transactions. List may be empty or out of date. {transactionsError}</p>
-      ) : null}
+      {transactionsError && (
+        <p>
+          {t(
+            'account.transaction.loadingError',
+            "Couldn't load transactions. List may be empty or out of date.",
+          )}{' '}
+          {transactionsError}
+        </p>
+      )}
       {allTransactions.length ? (
         transactionComponents
       ) : (
--- a/src/app/pages/AccountPage/Features/TransactionHistory/index.tsx
+++ b/src/app/pages/AccountPage/Features/TransactionHistory/index.tsx
@@ -9,7 +9,11 @@ import * as React from 'react'
 import { useTranslation } from 'react-i18next'
 import { useSelector } from 'react-redux'
 
-import { selectAccountAddress, selectTransactions } from '../../../../state/account/selectors'
+import {
+  selectAccountAddress,
+  selectTransactions,
+  selectTransactionsError,
+} from '../../../../state/account/selectors'
 
 interface Props {}
 
@@ -19,6 +23,7 @@ interface Props {}
 export function TransactionHistory(props: Props) {
   const { t } = useTranslation()
   const allTransactions = useSelector(selectTransactions)
+  const transactionsError = useSelector(selectTransactionsError)
   const address = useSelector(selectAccountAddress)
   const transactionComponents = allTransactions.map((t, i) => (
     <Transaction key={i} transaction={t} referenceAddress={address} />
@@ -26,6 +31,9 @@ export function TransactionHistory(props: Props) {
 
   return (
     <Box gap="small" margin="none">
+      {transactionsError ? (
+        <p>Couldn't load transactions. List may be empty or out of date. {transactionsError}</p>
+      ) : null}
       {allTransactions.length ? (
         transactionComponents
       ) : (
--- a/src/app/pages/AccountPage/index.tsx
+++ b/src/app/pages/AccountPage/index.tsx
@@ -123,24 +123,24 @@ export function AccountPage(props: Props) {
       {address && address !== '' && (
         <>
           <AccountSummary
-            address={account.address}
+            address={address}
             balance={balance}
             walletAddress={walletAddress}
             walletIsOpen={walletIsOpen}
           />
           <Box background="background-front" margin={{ vertical: 'small' }} direction="row" gap="small" wrap>
             <NavItem
-              route={`/account/${account.address}`}
+              route={`/account/${address}`}
               label={t('account.subnavigation.transactions', 'Transactions')}
             />
             <NavItem
-              route={`/account/${account.address}/active-delegations`}
+              route={`/account/${address}/active-delegations`}
               label={t('account.subnavigation.activeDelegations', 'Active delegations ({{count}})', {
                 count: stake.delegations.length,
               })}
             />
             <NavItem
-              route={`/account/${account.address}/debonding-delegations`}
+              route={`/account/${address}/debonding-delegations`}
               label={t('account.subnavigation.debondingDelegations', 'Debonding delegations ({{count}})', {
                 count: stake.debondingDelegations.length,
               })}
--- a/src/app/pages/AccountPage/index.tsx
+++ b/src/app/pages/AccountPage/index.tsx
@@ -117,9 +117,12 @@ export function AccountPage(props: Props) {
           </Box>
         </Layer>
       )}
-      {account.accountError ? (
-        <p>Couldn't load account. Information may be missing or out of date. {account.accountError}</p>
-      ) : null}
+      {account.accountError && (
+        <p>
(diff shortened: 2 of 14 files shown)


pro-wh:
I tried moving the catch up a few layers, to a point where we could decide what to do: continue with the previous list of validators and put a note in the UI that the API request failed

however, our `onError` handler to show the fatal error popup is still running, even when we catch the promise rejection. before our catch runs, in fact. is my reading of the redux-saga error handling tutorial https://redux-saga.js.org/docs/basics/ErrorHandling wrong? that page says a few things to the effect that the onError hook is a last resort, not that it's the first thing that runs and that `catch`es are worthless. so now what?

pro-wh:
the catch was working. there was a similar error that really was uncaught

cleaned up experimenal commits

added various state and UI to show the latest error. not translated yet

and still no test

but people can start reviewing now

pro-wh:
changing base branch so that we don't see other branches' commits

src/app/pages/StakingPage/Features/ValidatorList/index.tsx
```diff
@@ -93,6 +99,9 @@ export const ValidatorList = memo((props: Props) => {
   return (
     <Box pad="medium" background="background-front">
       {t('common.validators', 'Validators')}
+      {updateValidatorsError ? (
+        <p>Couldn't load validators. List may be empty or out of date. {updateValidatorsError}</p>
```
pro-wh, line 103: frontend devs please review

src/app/state/staking/saga.ts
```diff
@@ -84,7 +84,14 @@ function* loadDebondingDelegations(publicKey: Uint8Array) {
 
 function* refreshValidators() {
   const { accounts } = yield* call(getExplorerAPIs)
-  const validators = yield* call([accounts, accounts.getValidatorsList], {})
+  let validators
+  try {
+    validators = yield* call([accounts, accounts.getValidatorsList], {})
+  } catch (e) {
+    console.error('get validators list failed, continuing without updated list.', e)
```
pro-wh, line 91: this would be our first use of console in production. please discuss

we could alternatively build more comprehensive error rendering :shrug:

src/app/state/staking/saga.ts
```diff
@@ -84,7 +84,14 @@ function* loadDebondingDelegations(publicKey: Uint8Array) {
...
+    yield* put(stakingActions.updateValidatorsError('' + e))
```
pro-wh, line 92: could perhaps use `e.stack` here if:

1. we would know that e instanceof Error
2. we render in a `<pre>`

src/vendors/explorer/apis/AccountsApi.ts
```diff
@@ -381,14 +381,10 @@ export class AccountsApi extends runtime.BaseAPI {
...
     async getValidatorsList(requestParameters: GetValidatorsListRequest): Promise<Array<ValidatorRow>> {
-      try {
         const response = await this.getValidatorsListRaw(requestParameters);
         return await response.value();
-      } catch (e) {
-        // TODO: Properly handle this error - should show a warning or so! See #403
-        return [];
-      }
     }
+
```
pro-wh, line 387: this came from reverting the hotfix commit. should we keep it deleted?

src/app/state/staking/saga.ts
```diff
@@ -84,7 +84,14 @@ function* loadDebondingDelegations(publicKey: Uint8Array) {
 
 function* refreshValidators() {
   const { accounts } = yield* call(getExplorerAPIs)
-  const validators = yield* call([accounts, accounts.getValidatorsList], {})
+  let validators
+  try {
+    validators = yield* call([accounts, accounts.getValidatorsList], {})
```
pro-wh, line 89: by the way, why do we use `yield* effect`?

pro-wh:
added catches for the account info request and the transactions list

we should check that later code can actually survive with those unset

src/app/state/staking/saga.ts
pro-wh: third option: launch the error into the sun

pro-wh:
this breaks the

> Transactions
Active delegations (x)
Debonding delegations (x)

links

src/app/state/staking/saga.ts
```diff
@@ -85,7 +85,14 @@ function* loadDebondingDelegations(publicKey: Uint8Array) {
 
 function* refreshValidators() {
   const { accounts } = yield* call(getExplorerAPIs)
-  const validators = yield* call(() => accounts.getValidatorsList({ limit: 500 }))
```
pro-wh, line 88: cc @lukaw3d on the style change from using a lambda back to `call` with context and args

lukaw3d: Ah, fix in https://github.com/oasisprotocol/oasis-wallet-web/pull/442

src/app/state/staking/saga.ts
matevz: How about reporting errors with `put()` method https://redux-saga.js.org/docs/basics/ErrorHandling/ ? I imagined we should show a time-limited popup, if such call fail for example.

src/vendors/explorer/apis/AccountsApi.ts
matevz: If catching the error above works fine, then go ahead!

src/app/pages/StakingPage/Features/ValidatorList/index.tsx
matevz: I think this error should be translated using `t()`?

src/app/pages/AccountPage/index.tsx
```diff
@@ -117,6 +117,9 @@ export function AccountPage(props: Props) {
           </Box>
         </Layer>
       )}
+      {account.accountError ? (
+        <p>Couldn't load account. Information may be missing or out of date. {account.accountError}</p>
```
matevz, line 121: I think this error should be translated using `t()`?

src/app/pages/AccountPage/Features/TransactionHistory/index.tsx
```diff
@@ -19,13 +23,17 @@ interface Props {}
...
+  const transactionsError = useSelector(selectTransactionsError)
   const address = useSelector(selectAccountAddress)
   const transactionComponents = allTransactions.map((t, i) => (
     <Transaction key={i} transaction={t} referenceAddress={address} />
   ))
 
   return (
     <Box gap="small" margin="none">
+      {transactionsError ? (
+        <p>Couldn't load transactions. List may be empty or out of date. {transactionsError}</p>
```
matevz, line 35: I think this error should be translated using `t()`?

matevz (changes requested):

src/app/pages/StakingPage/Features/ValidatorList/index.tsx
pro-wh: ah yeah the outer part should

src/app/state/staking/saga.ts
pro-wh: there's actually a `put` right under this https://github.com/oasisprotocol/oasis-wallet-web/pull/413/files#diff-2fbfbfd97023b036563132e6139c482756aa57759d11641920c1020d9e98391cR43-R47

this log is in addition to the normal user-facing error handling done that way

src/app/pages/StakingPage/Features/ValidatorList/index.tsx
pro-wh: added

src/app/pages/AccountPage/index.tsx
pro-wh: added

src/app/pages/AccountPage/Features/TransactionHistory/index.tsx
pro-wh: added

pro-wh:
@matevz you're clear to take another look! thanks for the comments

pro-wh:
screenshot:

- address is from a validator entity on mainnet, but it's not important.
- to make this screenshot, I have the code tweaked to treat any HTTP status as failed. 200 is success, which the backend incidentally succeeded
- after #408 the 'available' balance would show up, as we'd be fetching it from the grpc-web node
- the fact that it displays zeros is unnecessarily scary. it's really unknown

image

matevz (approved):
LGTM. I see you're becoming a saga master :)

pro-wh:
thanks for reviewing!

Issue #267, opened 2021-09-20
Account switching issue - wrong account's details are displayed


tjanez:
Steps to reproduce:
- Open two different accounts (I used 2 different mnemonics).
- Switch between them with the account switcher button in the top-right corner.
- Account details are not updated and after switching, the wrong account's details are displayed.

Example screenshot (the chosen account is `oasis1qrs594jx6ytzgqp92uxqlr8n5qdcrpemv5xg4wn4` but they one displayed is `oasis1qqdcs4gj5g0kykdcy9s32snh8a58swgkhyj6kzdy`):
![Account switching issue - wrong account details are displayed](https://user-images.githubusercontent.com/2518544/134064878-85507b35-cca8-4927-a89d-e25ca70f4a1a.png)

PR #419, opened 2021-11-09
Account switching issue - wrong account's details are displayed

buberdds:
Closes https://github.com/oasisprotocol/oasis-wallet-web/issues/267.

### DIFF
--- a/src/app/state/wallet/saga.test.ts
+++ b/src/app/state/wallet/saga.test.ts
@@ -1,3 +1,4 @@
+import { select } from 'redux-saga/effects';
 import { push } from 'connected-react-router'
 import { expectSaga } from 'redux-saga-test-plan'
 import * as matchers from 'redux-saga-test-plan/matchers'
@@ -6,7 +7,8 @@ import { RootState } from 'types'
 
 import { walletActions } from '.'
 import { transactionActions } from '../transaction'
-import { getBalance, rootWalletSaga, walletSaga } from './saga'
+import { getBalance, rootWalletSaga, walletSaga, selectWallet } from './saga'
+import { selectActiveWallet } from './selectors'
 import { Wallet, WalletState } from './types'
 
 describe('Wallet Sagas', () => {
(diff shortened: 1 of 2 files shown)


pro-wh (approved):
this should do for now, thanks for the fix

although now that I really look at this, as well as the original code, it seems unidiomatic for us to be defining this kind of "URL follows selected wallet" kind of invariant by coding it imperatively in this "select a wallet then push URL" way. you've moved it into the consequence of a `takeEvery`, at least. doing the `push` in a 'later' `takeEvery(walletActions.walletSelected, ...)` might be nice too, unless that's taking things too far.

@Esya feel free to weigh in too if you have the time to look

buberdds:
Yeah I was trying to fix the issue without larger refactor as I am not familiar with the app yet.

After reading your comment I think you are correct. It all should work other way around where wallet dropdown/switcher would use router NavLink instead of dispatching current redux action. The problem is that current main wallet url handles different app states one for a default preview and one for selected wallet and selected wallet requires opening a wallet first. Maybe app need a different url for selected wallet where in a case of missing wallet user would go through "Open wallet" step?

pro-wh:
gonna try rebasing and merging

pro-wh:
alright, merged!

> The problem is that current main wallet url handles different app states one for a default preview and one for selected wallet and selected wallet requires opening a wallet first. Maybe app need a different url for selected wallet where in a case of missing wallet user would go through "Open wallet" step?

Thanks for the additional analysis on this. I'll take this into account if we get into discussing a larger rework of this. Using a different URL for users' wallets is something we'll consider. I wonder if we might alternatively try to unify more of the functionality actually to reduce the differences though. It should only come down to whether or not you can sign transactions, after all.

tjanez:
> Thanks for the additional analysis on this. I'll take this into account if we get into discussing a larger rework of this. Using a different URL for users' wallets is something we'll consider. I wonder if we might alternatively try to unify more of the functionality actually to reduce the differences though. It should only come down to whether or not you can sign transactions, after all.

@pro-wh, can you file this as an issue so it doesn't get lost?

pro-wh:
sure thing

Issue #275, opened 2021-09-21
Use default cursor on Amount and Block panes


matevz:
Moving mouse over Amount or Block panes shows the pointer cursor, but it's not really clickable:
![image](https://user-images.githubusercontent.com/83672/134136985-aca0878c-1ca6-4f12-b7e8-48f8c5b4fc60.png)

We should just use the default cursor there.

PR #423, opened 2021-11-10
Use default cursor on Amount and Block panes

buberdds:
Closes https://github.com/oasisprotocol/oasis-wallet-web/issues/275.

### DIFF
--- a/src/app/components/Transaction/InfoBox.tsx
+++ b/src/app/components/Transaction/InfoBox.tsx
@@ -1,30 +1,15 @@
 import { Box, Text } from 'grommet'
 import * as React from 'react'
-import { useHistory } from 'react-router'
 
 interface DetailProps {
   icon?: React.ReactNode
   label: string
   value: string | React.ReactNode
-  link?: string
 }
 
 export function InfoBox(props: DetailProps) {
-  const history = useHistory()
-  const boxClicked = () => {
-    if (props.link) {
-      history.push(props.link)
-    }
-  }
-
   return (
-    <Box
-      direction="row"
-      gap="small"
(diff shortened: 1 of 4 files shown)


src/app/components/Transaction/InfoBox.tsx
```diff
@@ -22,7 +22,7 @@ export function InfoBox(props: DetailProps) {
       direction="row"
       gap="small"
       hoverIndicator={{ color: 'background-contrast' }}
-      onClick={() => boxClicked()}
+      onClick={props.link ? boxClicked : undefined}
```
lukaw3d, line 25: Ah, nice!

I'd remove `if (props.link) {` from `boxClicked` too, but then typescript will complain.

And I'd replace onClick with wrapping NavLink around Box (real link with right-click and middle-click), but making it a conditional-wrapper results in another component.

Maybe this is better solved one level higher:
```diff
  <InfoBox icon={<Money color="brand" />} label={t('common.amount', 'Amount')} value={amount} />
+ <NavLink to={`/account/${otherAddress}`}>
    <InfoBox
      icon={<ContactInfo color="brand" />}
      label={designation}
      value={<ShortAddress address={otherAddress} />}
-     link={`/account/${otherAddress}`}    
    />
+ </NavLink>


- link?: string
- onClick={props.link ? boxClicked : undefined}
```

lukaw3d (approved):

src/app/components/Transaction/InfoBox.tsx
buberdds: looks much better, thanks! will bump PR today

PR #426, opened 2021-11-11
Bump swagger and API to latest upstream version

matevz:
Fixes #235:
- synchronizes typescript bindings with the latest [oasis-explorer openapi specs swagger.yml](https://github.com/everstake/oasis-explorer/blob/master/swagger/swagger.yml)
- adds new README.md with instructions how to generate fresh typescript bindings from swagger
- fixes fetching correct commission fees from the new API

PR #430, opened 2021-11-13
hotfix: Return empty list of validators, if oasismonitor.com is down

pro-wh:
need to reapply from #406 following #426 when we regenerated the API stubs


tjanez (approved):
Nice catch, @pro-wh !

Could you add a note to https://github.com/oasisprotocol/oasis-wallet-web/blob/develop/src/vendors/explorer/README.md that this change needs to be made every time TypeScript bindings for Oasis Explorer are re-generated?

pro-wh:
yup

Issue #282, opened 2021-09-21
Show more than top 50 validators when staking/delegating


tjanez:
Currently, only the top 50 validators are shown when a user wants to stake/delegate their tokens.

The Mainnet currently has 110 validators to whom one could delegate tokens.

matevz:
Fixed in #282, thanks @lukaw3d

PR #435, opened 2021-11-16
state/staking: show first 500 validators instead of 50

lukaw3d:
Mitigates https://github.com/oasisprotocol/oasis-wallet-web/issues/282

But https://monitor.oasis.dev/data/validators?limit=500 unexpectedly returns 134 rows (18 inactive), not 110 :shrug:

### DIFF
--- a/src/app/state/staking/saga.ts
+++ b/src/app/state/staking/saga.ts
@@ -1,6 +1,6 @@
 // import { take, call, put, select, takeLatest } from 'redux-saga/effects';
 // import { stakingActions as actions } from '.';
-import { address as oasisAddress, quantity } from '@oasisprotocol/client'
+import { address as oasisAddress, consensus, quantity } from '@oasisprotocol/client'
 import {
   StakingDebondingDelegationInfo,
   StakingDelegationInfo,
@@ -84,8 +84,15 @@ function* loadDebondingDelegations(publicKey: Uint8Array) {
 }
 
 function* refreshValidators() {
+  // Fetch active validator set size.
+  const nic = yield* call(getOasisNic)
+  const consensusParams = yield* call([nic, nic.schedulerConsensusParameters], consensus.HEIGHT_LATEST)
+
+  // Fetch validators.
   const { accounts } = yield* call(getExplorerAPIs)
-  const validators = yield* call(() => accounts.getValidatorsList({ limit: 500 }))
+  const validators = yield* call([accounts, accounts.getValidatorsList], {
+    limit: consensusParams.max_validators,
+  })
   const currentEpoch = yield* select(selectEpoch)
 
   const payload: Validator[] = validators
--- a/src/app/state/staking/saga.ts
+++ b/src/app/state/staking/saga.ts
@@ -1,6 +1,6 @@
 // import { take, call, put, select, takeLatest } from 'redux-saga/effects';
 // import { stakingActions as actions } from '.';
-import { address as oasisAddress, consensus, quantity } from '@oasisprotocol/client'
+import { address as oasisAddress, quantity } from '@oasisprotocol/client'
 import {
   StakingDebondingDelegationInfo,
   StakingDelegationInfo,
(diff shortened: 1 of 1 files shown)


matevz:
I think https://monitor.oasis.dev/data/validators?limit=size_of_current_active_set should be the correct approach. 134 seems correct, because oasis monitor shows all the validators, even those with smaller stakes which don't get any rewards.

src/app/state/staking/saga.ts
```diff
@@ -85,7 +85,7 @@ function* loadDebondingDelegations(publicKey: Uint8Array) {
 
 function* refreshValidators() {
   const { accounts } = yield* call(getExplorerAPIs)
-  const validators = yield* call([accounts, accounts.getValidatorsList], {})
+  const validators = yield* call(() => accounts.getValidatorsList({ limit: 500 }))
```
matevz, line 88: Set `limit: 110` as this is the current active validator set.

matevz (changes requested):

matevz:
I added support for fetching current active validator set size. Feel free to review the change, test it and merge it.

matevz (approved):

src/app/state/staking/saga.ts
```diff
@@ -84,8 +84,15 @@ function* loadDebondingDelegations(publicKey: Uint8Array) {
...
+  // Fetch active validator set size.
+  const nic = yield* call(getOasisNic)
+  const consensusParams = yield* call([nic, nic.schedulerConsensusParameters], consensus.HEIGHT_LATEST)
+
+  // Fetch validators.
   const { accounts } = yield* call(getExplorerAPIs)
-  const validators = yield* call([accounts, accounts.getValidatorsList], {})
+  const validators = yield* call([accounts, accounts.getValidatorsList], {
+    limit: consensusParams.max_validators,
+  })
```
lukaw3d, line 95: Does max_validators include inactive validators?

Otherwise this isn't exactly correct: When fetching 500, status=inactive validators were at indexes `[86, 98, 113, 114, 118, 119, 120, 121, 123, 124, 125, 126, 127, 130, 131, 132, 133]`, so this loses active validators at 108 and 109

matevz: `max_validators` contains current maximum allowed size of active validator set. If a validator is not active, there is some expiration period before it gets evicted from the active validator set, so it may be a valid assumption the validator is still in the active set although inactive currently. The same goes for active validators - they may be registered as validators, but may not be in the active validator set yet.

So I was unable to find out how oasis-explorer generates the list of validators here https://github.com/everstake/oasis-explorer/blob/master/dao/clickhouse/validators.go#L13-L17 and how it updates it here https://github.com/everstake/oasis-explorer/blob/master/dao/clickhouse/validators.go#L207 and combining it with metadata registry of entities here https://github.com/oasisprotocol/metadata-registry. I suggest we open a separate issue regarding those borderline cases. For now I would simply return the top-k validators where k=max_validators parameter and hide the rest.

matevz: I discussed this with @kostko and the correct approach would be first to call `schedulerGetValidators(height)` which gives you the actual validator set (no inactive validators!) and then intersect that list with the one from oasis-explorer to obtain detailed information (icon, name, escrow, fee etc.).

lukaw3d: Nope, there are 116 (and block explorer API doesn't support filtering by active status)

pro-wh:
if there are entities with validator-role-enabled nodes correctly entered into the registry but that don't have enough stake to be elected into the validator committee ("active"), then seems like it's fine to list them as well. we do hear about people asking how to provide the initial stake to start a new validator.

marking which validators are already active would be a welcome enhancement though.

tjanez:
> if there are entities with validator-role-enabled nodes correctly entered into the registry but that don't have enough stake to be elected into the validator committee ("active"), then seems like it's fine to list them as well. we do hear about people asking how to provide the initial stake to start a new validator. we don't intend to make that difficult.
> 
> marking which validators are already active would be a welcome enhancement though.

Agreed.

src/app/state/staking/saga.ts
tjanez: > I discussed this with @kostko and the correct approach would be first to call schedulerGetValidators(height) which gives you the actual validator set (no inactive validators!)

I would prefer to list all validators, i.e. all entities that have a validator node registered, even if they are not in the current validator set.

But it would be nice to somehow visually distinguish those that are not active versus active ones.

lukaw3d: 116 validators with status=active are marked as green, vs 18 red.
image

If I understand, you are proposing another symbol for bottom 6 status=active validators?

lukaw3d: For now, I'll merge this by fetching 500 (with Matevž's permission)

Issue #436, opened 2021-11-16
Validator fee rounding issue


lukaw3d:
![image](https://user-images.githubusercontent.com/3758846/141881566-c76fec08-2aa7-4141-8989-ac8d7d04ef1f.png)

lukaw3d:
We should probably use `BigNumber` around the code containing
```js
10 ** 9
/ 100_000
* 100
parseFloat(
Number(
Math.floor(
Math.round(
```

PR #437, opened 2021-11-16
Reproduce rounding issue with 7% validator fee

lukaw3d:
Reproduces https://github.com/oasisprotocol/oasis-wallet-web/issues/436

### DIFF
--- a/src/app/pages/StakingPage/Features/DelegationList/__tests__/__snapshots__/ActiveDelegationList.test.tsx.snap
+++ b/src/app/pages/StakingPage/Features/DelegationList/__tests__/__snapshots__/ActiveDelegationList.test.tsx.snap
@@ -604,7 +604,7 @@ Object {
                     id="cell-fee-undefined"
                     role="gridcell"
                   >
-                    10%
+                    7.000000000000001%
                   </div>
                 </div>
               </div>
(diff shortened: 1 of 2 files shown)


src/app/pages/StakingPage/Features/DelegationList/__tests__/__snapshots__/ActiveDelegationList.test.tsx.snap
```diff
@@ -604,7 +604,7 @@ Object {
                     id="cell-fee-undefined"
                     role="gridcell"
                   >
-                    10%
+                    7.000000000000001%
```
lukaw3d, line 607: <

matevz:
Interesting bug :thinking: While you're at it, can you also round the values in the Fee column in validator list to 2 decimals or so (validators could  use some crazy fee just to spoil the UI) and maybe align the text to right :wink:

pro-wh:
let's round to the commission denominator. that's thousandths of a percent https://github.com/oasisprotocol/oasis-core/blob/v21.3.6/go/staking/api/commission.go#L19

pro-wh (approved):
let's merge this so that we'll have a diff in the snapshot if something fixes this

lukaw3d:
Okay :shrug:

PR #442, opened 2021-11-17
Fix code style

lukaw3d:
Fixes code style change in https://github.com/oasisprotocol/oasis-wallet-web/pull/435


pro-wh (approved):
oh thanks, I was wonder about this part

matevz (approved):

PR #458, opened 2021-11-24
StakingPage: Support sorting by validator name

matevz:
Fixes sorting by name on the validator list page. If validator name is not present, sort by oasis address.

image

### DIFF
--- a/src/app/pages/StakingPage/Features/ValidatorList/index.tsx
+++ b/src/app/pages/StakingPage/Features/ValidatorList/index.tsx
@@ -66,14 +66,15 @@ export const ValidatorList = memo((props: Props) => {
     },
     {
       name: t('validator.name', 'Name'),
+      selector: 'name',
       cell: datum =>
-        datum.name ? (
-          datum.name
-        ) : (
+        datum.name ?? (
           <Text data-tag="allowRowEvents">
             <ShortAddress address={datum.address} />
           </Text>
         ),
+      sortable: true,
+      sortFunction: (row1, row2) => (row1.name ?? row1.address).localeCompare(row2.name ?? row2.address),
     },
     {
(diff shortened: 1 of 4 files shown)


src/app/pages/StakingPage/Features/ValidatorList/index.tsx
```diff
@@ -74,6 +75,11 @@ export const ValidatorList = memo((props: Props) => {
             <ShortAddress address={datum.address} />
           </Text>
         ),
+      sortable: true,
+      sortFunction: (row1, row2) =>
+        row1.name
+          ? row1.name.localeCompare(row2.name ?? row2.address)
+          : row1.address.localeCompare(row2.name ?? row2.address),
```
lukaw3d, line 82:
```suggestion
      sortFunction: (row1, row2) => (row1.name ?? row1.address).localeCompare(row2.name ?? row2.address),
```
is more symmetric

lukaw3d (approved):

src/app/pages/StakingPage/Features/ValidatorList/index.tsx
pro-wh: agreed with this suggestion above

could you also confirm that `??` is appropriate? surrounding code, e.g. `datum.name ? ( ... ) : ( ... )`, suggests that `&&` would be consistent

matevz: Thanks, I applied the suggested change and also used `??` for `datum.name ? ( ... ) : ( ... )` above.

matevz:
I noticed sorting by name also didn't work in the Active delegations and Debonding delegations pages as well as sorting by Amount. Fixed those as well.

PR #486, opened 2021-11-29
Check that HDKey.generateMnemonic is bip39 valid

lukaw3d:
This is never expected to fail


pro-wh (approved):
fine with me

pro-wh:
https://github.com/bitcoinjs/bip39/blob/5faee2c17b2195f30b03cb125df68c20d7dd584b/src/index.js#L152

hm, any exception -> treat as invalid. I wonder if that ever does anything unexpected

lukaw3d:
> https://github.com/bitcoinjs/bip39/blob/5faee2c17b2195f30b03cb125df68c20d7dd584b/src/index.js#L152
> 
> hm, any exception -> treat as invalid. I wonder if that ever does anything unexpected

`new Array(100000).fill(0).every(a => validateMnemonic(hdkey.HDKey.generateMnemonic(256).split(' ').join(' ')))` didn't fail :shrug:

PR #487, opened 2021-11-30
Add README steps to run without local network

lukaw3d:


pro-wh (approved):
oh thanks, I hadn't known about this

PR #517, opened 2021-12-03
Copy stacktrace with separator newlines

lukaw3d:


pro-wh (approved):
this will be good for Firefox, where the .stack doesn't add its own copy of the message

interesting that we have two different preparations of the text: one plaintext for copying, one constructed in HTML--`<pre>`s and now `<br>`s--for display

src/app/components/FatalErrorHandler/index.tsx
```diff
@@ -61,13 +63,7 @@ export function FatalErrorHandler(props: Props) {
           }}
           pad={{ horizontal: 'small', vertical: 'xsmall' }}
         >
-          <Box>
-            <Text weight="bold" data-testid="fatalerror-message">
-              {fatalError.message}
-            </Text>
-          </Box>
-          <pre>{fatalError.sagaStack}</pre>
-          <pre>{fatalError.stack}</pre>
```
lukaw3d, line 70: > interesting that we have two different preparations of the text: one plaintext for copying, one constructed in HTML`<pre>`s and now `<br>`s for display

Good point - deduplicated it now. `<Box><Text weight="bold">message` looked nearly the same as `<pre>`, so I'd rather reduce complexity

Issue #521, opened 2021-12-06
Don't allow Google Translate to show wrong mnemonic


lukaw3d:
- Disable Google Translate on mnemonic - https://cloud.google.com/translate/troubleshooting
- Force users to type out whole mnemonic before creating a new wallet (instead of clicking translated missing words)

Related to https://github.com/oasisprotocol/oasis-wallet-web/issues/513

PR #527, opened 2021-12-07
Force new users to type whole mnemonic before creating a wallet

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/521

| Before | After |
| --- | --- |
|  |  |


pro-wh:
the coverage is mad that your test doesn't cover the `onClickOutside` :laughing:

pro-wh (approved):
alright that looks like an appropriate implementation of the forced enter-the-whole-mnemonic idea

matevz (changes requested):
Generally I like the way this was implemented.

Two details I'd like you to settle:
1. Add the cancel button similar to what we had before: 
image
2. Fix responsiveness (maybe jump to fullscreen and align elements vertically?).
   Before: 
image

   After:
   
image

lukaw3d:
localhost_3000_create-wallet

localhost_3000_create-wallet (1)

localhost_3000_create-wallet (2)

localhost_3000_create-wallet (3)

pro-wh:
dark theme :monocle_face:

matevz (approved):
Good work!

Just one minor request. Can you make the Confirm mnemonic popup true fullscreen (without any sight of the previous page near the borders) once the window width is smaller than 770px? This behavior was already there before your patch.

PR #537, opened 2021-12-13
OpenWallet: Update link to BitPie FAQ page

matevz:
Oasis Web Wallet issues have been moved to a more common FAQ page at https://docs.oasis.dev/general/manage-tokens/faq. This PR updates the link in the BitPie warning.

### DIFF
--- a/src/app/pages/OpenWalletPage/index.tsx
+++ b/src/app/pages/OpenWalletPage/index.tsx
@@ -79,9 +79,7 @@ export function SelectOpenMethod() {
           i18nKey="openWallet.bitpie.warning"
           t={t}
           defaults="❗ BitPie wallet users: You cannot import the mnemonic phrase directly from your BitPie wallet. <0>Check documentation for details.</0>"
-          components={[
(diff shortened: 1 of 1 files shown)


lukaw3d (approved):

src/app/pages/OpenWalletPage/index.tsx
```diff
@@ -79,9 +79,7 @@ export function SelectOpenMethod() {
           i18nKey="openWallet.bitpie.warning"
           t={t}
           defaults="❗ BitPie wallet users: You cannot import the mnemonic phrase directly from your BitPie wallet. <0>Check documentation for details.</0>"
-          components={[
```
tjanez, line 82: Should we keep the link to a particular question in case FAQ grows in the future?

lukaw3d: (I don't care too much, but if we do: Can anchor be stable if heading changes in the future?)

tjanez: Unfortunately, having a stable anchor is not possible with GitBook.

PR #538, opened 2021-12-15
i18n: Sync French translation

matevz:
This PR syncs French translation with upstream and also fixes some typos.

Kudos to Raphael Marc Landers :wink:

PR #541, opened 2021-12-15
CI: Require English translation file in sync with source code

matevz:
This PR adds a check to CI that all terms in source files are also present in the original English translation file which is a baseline for Transifex.

Developers should always run `yarn extract-messages`, if they changed any terms in the source code to also update the English translation file. Synchronizing other translations is done by Transifex which fetches any English updates and applies them to other languages.

Other changes:
- enables sorting of all terms in translation files to assure canonical form among i18next-scaner versions and source code refactorings in the future
- adds missing defaults for all terms in the source code

### DIFF
--- a/.github/workflows/mega-linter.yml
+++ b/.github/workflows/mega-linter.yml
@@ -44,6 +44,15 @@ jobs:
           node-version: 14.x
       - run: yarn install --frozen-lockfile
 
+      # Check if all i18n terms in code are synced with source (English) translation file.
+      - name: Check English translation is synced
+        run: |
+          rm src/locales/en/translation.json
+          yarn extract-messages
+          git diff --exit-code src/locales/en/translation.json
+          # Check passed, revert changes to any other translation so it doesn't get applied by mega-linter.
+          git restore .
+
       # Mega-Linter
       - name: Mega-Linter
         id: ml
--- a/internals/extractMessages/i18next-scanner.config.js
+++ b/internals/extractMessages/i18next-scanner.config.js
@@ -6,6 +6,7 @@ module.exports = {
   options: {
     debug: false,
     removeUnusedKeys: true,
+    sort: true,
     func: {
       list: ['t'],
       extensions: ['js', 'jsx'] // We dont want this extension because we manually check on transform function below
--- a/src/app/components/ErrorFormatter/index.tsx
+++ b/src/app/components/ErrorFormatter/index.tsx
@@ -18,15 +18,15 @@ export function ErrorFormatter(props: Props) {
   const message = props.message
 
   const errorMap: { [code in WalletErrors]: string | React.ReactElement } = {
-    [WalletErrors.UnknownError]: t('errors.unknown', { message }),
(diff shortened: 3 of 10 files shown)


src/locales/fr/translation.json
```diff
@@ -41,14 +41,18 @@
...
     }
   },
   "createWallet": {
     "newMnemonic": "Générer une nouvelle phrase secrète",
     "thisIsYourPhrase": "Voici votre phrase secrète",
     "instruction": "Sauvegardez votre phrase secrète <1>dans le bon ordre</1> et dans un endroit sécurisé, vous en aurez besoin pour ouvrir votre portefeuille à nouveau.",
     "doNotShare": "Ne partagez jamais votre phrase secrète, l'accès à votre phrase secrète permet d'accéder à votre portefeuille ainsi qu'à vos tokens.",
-    "confirmSaved": "J'ai sauvegardé ma phrase secrète."
+    "confirmSaved": "J'ai sauvegardé ma phrase secrète.",
+    "mnemonicMismatch": "Entered mnemonic does not match."
```
pro-wh, line 55: is this correct that the script puts english strings into the french json file?

pro-wh: wait is this mixed with #538?

matevz: Ah, crap, I accidentally committed the French translation. Disregard this.

matevz: Oh, interesting. I didn't push the changes, mega-linter did. It seems that all changes to the tree occurring in the mega-linter's build job is automatically applied. And when I added this translation file check, it is also grabbed and applied. Let me reconsider this...

memoriesadrift: Hi! 
While working on an unrelated PR, English strings were added into the French json upon running `yarn run extract-messages`
If this isn't desired, it can be avoided by adding a default value for keys in the i18n scanner config file, under `options`:
`defaultValue: (lang, namespace, key) => { return '_TRANSLATION_MISSING_' }`

pro-wh: @matevz I'll mark this as "changes requested" for now.

pro-wh (changes requested):
lint will add English text to French translation file. we said we'd reconsider

.github/workflows/mega-linter.yml
```diff
@@ -44,6 +44,15 @@ jobs:
           node-version: 14.x
       - run: yarn install --frozen-lockfile
 
+      # Check if all i18n terms in code are synced with source (English) translation file.
+      - name: Check English translation is synced
```
lukaw3d, line 48: Update since https://github.com/oasisprotocol/oasis-wallet-web/pull/753 + https://github.com/oasisprotocol/oasis-wallet-web/pull/799: translation paths are now typechecked, so typescript should already complain about any translation paths missing from en.json

pro-wh: this additionally ensures that the strings match

pro-wh (commented):
code lg

were there any changes to the translation other than the sorting?

lukaw3d (approved):

PR #542, opened 2021-12-17
i18n: Add missing full stop, base-64 -> Base64

matevz:
This PR fixes the import base64-encoded private key term in the original English translation.


pro-wh (approved):
yeah

Issue #533, opened 2021-12-10
Not possible to send certain fractional amounts of tokens


tjanez:
Trying to send certain fractional amounts of tokens, e.g. 2.0121 or 4.0242, results in the following error:

```
Unknown error : 2012100000.0000002 can't be converted to BigInt because it isn't an integer
```

```
Unknown error : 4024200000.0000005 can't be converted to BigInt because it isn't an integer
```

Here are the corresponding screenshots:
![Screenshot from 2021-12-10 21-48-51](https://user-images.githubusercontent.com/2518544/145639926-eab110de-fc0a-4ca7-aded-ade7f0531fe1.png)
![Screenshot from 2021-12-10 21-49-12](https://user-images.githubusercontent.com/2518544/145640038-1b29e0d1-6c91-44f9-957d-0e77f8b0ef38.png)

PR #543, opened 2021-12-21
Not possible to send certain fractional amounts of tokens

buberdds:
Closes https://github.com/oasisprotocol/oasis-wallet-web/issues/533.

### DIFF
--- a/src/app/lib/helpers.test.ts
+++ b/src/app/lib/helpers.test.ts
@@ -2,8 +2,8 @@ import { parseNumberToBigInt } from './helpers'
 
 describe('parseNumberToBigInt', () => {
   it('should return a value of type bigint', () => {
-    expect(parseNumberToBigInt(33333.3)).toEqual(33333300000000n)
-    expect(parseNumberToBigInt(2.0121)).toEqual(2012100000n)
-    expect(parseNumberToBigInt(5)).toEqual(5000000000n)
+    expect(parseNumberToBigInt(33333.3)).toEqual(33333_300_000_000n)
+    expect(parseNumberToBigInt(2.0121)).toEqual(2_012_100_000n)
+    expect(parseNumberToBigInt(5)).toEqual(5_000_000_000n)
   })
 })
--- a/src/app/lib/helpers.test.ts
+++ b/src/app/lib/helpers.test.ts
@@ -1,0 +1,9 @@
+import { parseNumberToBigInt } from './helpers'
+
+describe('parseNumberToBigInt', () => {
+  it('should return a value of type bigint', () => {
+    expect(parseNumberToBigInt(33333.3)).toEqual(33333300000000n)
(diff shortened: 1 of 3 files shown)


pro-wh (approved):
Good fix for that issue, thanks.

Small ideas for this PR:

- in the tests, group digits in the long big integers with `_`
- could we use Math.round() instead of toFixed() so that we don't have to take a trip through string type? should behave the same

The weakness of this is that we don't detect inputs that had too much precision which we coulnd't represent on chain, e.g. 5.1112223335. Ideally we'd still error out in those cases. But it's better to have this, and I'm in favor of merging it

buberdds:
Thanks for your input. Can you give me sample code for # 1 as I am not following your idea about grouping digits yet. 

As for precision issue, again thanks for sharing that thought. It sounds more like a validation not parsing to me. I actually started tackling another issue related to amount field so I will look into validation as well.

src/app/lib/helpers.test.ts
```diff
@@ -0,0 +1,9 @@
+import { parseNumberToBigInt } from './helpers'
+
+describe('parseNumberToBigInt', () => {
+  it('should return a value of type bigint', () => {
+    expect(parseNumberToBigInt(33333.3)).toEqual(33333300000000n)
+    expect(parseNumberToBigInt(2.0121)).toEqual(2012100000n)
+    expect(parseNumberToBigInt(5)).toEqual(5000000000n)
```
pro-wh, line 7:
```suggestion
    expect(parseNumberToBigInt(33333.3)).toEqual(33333_300_000_000n)
    expect(parseNumberToBigInt(2.0121)).toEqual(2_012_100_000n)
    expect(parseNumberToBigInt(5)).toEqual(5_000_000_000n)
```

something like this, so it's easier to eyeball where 10^9 is.

src/app/lib/helpers.ts
```diff
@@ -48,3 +48,5 @@ export function concat(...parts: Uint8Array[]) {
   }
   return result
 }
+
+export const parseNumberToBigInt = (value: number) => BigInt(Math.round(value * 10 ** 9))
```
lukaw3d, line 52: @pro-wh is this usually called converting to base units?

I'm not sure if that's accurate for `parseNumberToBigInt(action.payload.shares)`

pro-wh: Should it be something else? Lots of software, e.g. this and oasis scan, seem to be using giga shares.

lukaw3d: Not sure.

I rebased. Happy to merge

pro-wh: ok let's proceed to merge. there's been some talk of moving away from presenting shares or gigashares to end users completely, so we'll have a lot more design discussion when we do that

Issue #544, opened 2021-12-22
Expanding validator reloads all validator logos


lukaw3d:
Related to https://github.com/oasisprotocol/oasis-wallet-web/pull/457

(staging was redeployed with logos and CSP)
- open https://wallet.stg.oasisprotocol.org/account/oasis1qpzwzf4khcr957ut8wkxlkycqrm36lh8mgsl6g4h/stake
- open console / network
- expand a validator
- all logos get fetched again
- expand another validator
- all logos get fetched again

PR #457, opened 2021-11-24
StakingPage: Add support for validator icon (move from #298)

matevz:
Adds validators icons.

image

PR #545, opened 2021-12-22
Fix validator logos reloading on expand

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/544
Related https://github.com/oasisprotocol/oasis-wallet-web/pull/457

- Added native lazy loading to logos
- All validator rows were destroyed and recreated because `keyField` on rows and `id` on cols were missing
- Show validator logos in DelegationList

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -19,6 +19,7 @@
   "jest.showCoverageOnLoad": true,
   "typescriptHero.imports.insertSemicolons": false,
   "files.exclude": {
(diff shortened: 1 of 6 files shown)


pro-wh (approved):
was there an issue with the browser not caching these?

lukaw3d:
Caching depends on server where images are hosted. It seems everstake has nocache, and multiple error responses weren't cached

pro-wh:
tragic. thanks for the info

Issue #492, opened 2021-11-30
Add 100% button when sending or debonding ROSE


matevz:
Currently, when sending or debonding all your ROSE balance, it is difficult to enter the exact amount since it can be rounded and you will not type it in entirety. I suggest we add a *100%* button in the lower-right corner of the input field which populates the field with the maximum still valid ROSE value.

Example from bitstamp trading view: 
![image](https://user-images.githubusercontent.com/83672/144042076-d060c5e5-16ad-4664-b5ea-7f0f7c6026c8.png)

Example from binance withdrawal page:
![image](https://user-images.githubusercontent.com/83672/144042713-b29e0982-66fa-4aff-862a-f7e575e5c678.png)

lukaw3d:
We should only add a button for reclaiming 100% of stake for now. Transferring 100% and staking 100% will get complicated with fees

PR #565, opened 2022-01-03
Add 100% button when sending or debonding ROSE 

buberdds:
Draft PR that targets https://github.com/oasisprotocol/oasis-wallet-web/issues/492

Screenshot 2022-01-03 at 10 26 43

Screenshot 2022-01-03 at 10 18 56
(disabled due to 0 tokens on this account)

### DIFF
--- a/src/app/components/AmountTextInput/index.tsx
+++ b/src/app/components/AmountTextInput/index.tsx
@@ -1,0 +1,63 @@
+/**
+ *
+ * AmountTextInput
+ *
+ */
+import React from 'react'
+import { Box, Button, FormField, TextInput } from 'grommet'
+
+interface Props {
+  disabled: boolean
+  label?: string
+  min?: number
+  max?: number
+  placeholder?: string
+  handleMaxValue: () => void
+  inline?: boolean
+}
+
+const contentProps = {
+  border: { position: 'inner' },
+  round: 'xsmall',
+}
+
+export function AmountTextInput({
+  disabled,
+  label,
+  min = 0,
+  max,
+  handleMaxValue,
+  placeholder = '0',
+  inline = false,
+}: Props) {
+  return (
+    <>
+      <FormField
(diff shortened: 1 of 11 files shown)


src/app/components/AmountTextInput/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+}: Props) {
+  return (
+    <>
+      <FormField
+        contentProps={inline ? contentProps : {}}
+        htmlFor="amount-id"
+        label={label}
+        name="amount"
+        validate={{
+          regexp: /^\d*((\.|,)\d{0,9})?$/,
```
buberdds, line 41: How many decimal places are allowed?  for example `AmountFormatter` defaults to 15, but it's not aligned with `10 ** 9`

src/app/pages/AccountPage/Features/SendTransaction/index.tsx
```diff
@@ -42,7 +60,12 @@ export function SendTransaction() {
 
   return (
     <Box border={{ color: 'background-front-border', size: '1px' }} round="5px" background="background-front">
-      <Form>
+      <Form<SendTransactionFormState>
+        validate="submit"
```
buberdds, line 64: I went with full `Form` API.  Props like `validate` and `onSubmit` look good here. I am not sure about typed form, `onChange` and `value`.  Should I  discard these and  restore

```
  const [recipient, setRecipient] = useState('')
  const [amount, setAmount] = useState('')
  ```
pattern ?

src/app/components/AmountTextInput/index.tsx
pro-wh: should be 9, dunno why AmountFormatter mentions 15

buberdds: Ok, thanks. I will reopen PR once it's ready.

PR #570, opened 2022-01-07
Show a warning banner when no wallet is open

lukaw3d:
Some users have been confused when they don't see buttons to transfer when "opening" a wallet by URL, e.g.
https://wallet.oasisprotocol.org/account/oasis1qrh9yz4nwsngkj5w9w6hktn4rca8ycwm0gw8v2zy

| Before | After |
| --- | --- |
| image | image |


pro-wh (approved):
I'm in favor of adding this or something like this.

Side idea: could "Open your wallet" be too abstract here, as really the next step is to click "Home"? Or maybe we could include a link to https://wallet.oasisprotocol.org/open-wallet as part of this banner

tjanez:
> Or maybe we could include a link to https://wallet.oasisprotocol.org/open-wallet as part of this banner

I like this idea.

tjanez (approved):

lukaw3d:
Updated: image

PR #595, opened 2022-01-19
Differentiate transaction types with icons

lukaw3d:
| Before | After |
| --- | --- |
| localhost_3000_preview (1)  | localhost_3000_preview  |


pro-wh (approved):
I like the new look of these, thanks

Issue #501, opened 2021-12-01
UI breaks after expanding a validator quickly after reload


lukaw3d:
- open https://wallet.oasisprotocol.org/account/oasis1qq3xrq0urs8qcffhvmhfhz4p0mu7ewc8rscnlwxe/stake
- open console and simulate slower network `Network Throttling: Fast 3G`
- reload page
- quickly click `Staking Fund` (before GetEpoch request?)
- UI disappears

![Peek 2021-12-01 06-55](https://user-images.githubusercontent.com/3758846/144179745-e320cd20-c935-44ad-afc5-b583b86d5866.gif)

lukaw3d:
@matevz  Seems to be caused because props are readonly, and sort mutates original array
https://github.com/oasisprotocol/oasis-wallet-web/blob/e994b39d61e9362af7f30a7f27651fbc7862ec01/src/app/pages/StakingPage/Features/CommissionBounds/index.tsx#L54-L59

lukaw3d:
if I add `.slice()` to clone array before, there's no error but:

| click quickly after reload | click slowly after reload |
| --- | --- |
|  ![Screenshot from 2021-12-01 07-30-14](https://user-images.githubusercontent.com/3758846/144184406-509bd126-ed1a-4cd0-99a1-86a3ed29aca8.png)  |  ![Screenshot from 2021-12-01 07-30-26](https://user-images.githubusercontent.com/3758846/144184413-5c2fb3be-b938-4caf-b811-dd22740cd847.png) |

matevz:
I can't seem to reproduce this. Are there issues with oasis monitor again?

kinqsta:
I just have same problem

```
Http response at 400 or 500 level
The above error occurred in task ft
    created by takeEvery(network/selectNetwork, ft)
    created by ht
Tasks cancelled due to error:
takeEvery(network/selectNetwork, ft)
D@https://wallet.oasisprotocol.org/static/js/2.48cea8b1.chunk.js:2:1047878
Ft@https://wallet.oasisprotocol.org/static/js/2.48cea8b1.chunk.js:2:1070696
Ut/<@https://wallet.oasisprotocol.org/static/js/2.48cea8b1.chunk.js:2:1069926
Ue@https://wallet.oasisprotocol.org/static/js/2.48cea8b1.chunk.js:2:1056709
Be@https://wallet.oasisprotocol.org/static/js/2.48cea8b1.chunk.js:2:1056515
[...]

PR #611, opened 2022-01-26
Clone bounds before sorting (fixes error mutating readonly props)

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/501


pro-wh (approved):
o nice find

PR #618, opened 2022-02-14
Improve Ledger errors

lukaw3d:
## WebUSB is not supported

Before - `TransportWebUSB.create` caught TypeError
Screen Shot 2022-02-14 at 20 16 20

After - `TransportWebUSB.isSupported`
Screen Shot 2022-02-14 at 20 18 06

## Oasis App is closed

Before - `publicKey (0x05 0x01 0 0 path)` command failed against the open app:
wallet oasisprotocol org_open-wallet

After - `appName` is not "Oasis":
localhost_3000_open-wallet_ledger


pro-wh (approved):
looking good

PR #619, opened 2022-02-14
Fix waiting in async tests (wait for expect.resolves / rejects)

lukaw3d:
Previously mistakes in these tests would still pass and just print UnhandledPromiseRejectionWarning, Expected value, Received value


pro-wh (approved):
https://jestjs.io/docs/tutorial-async#rejects oh yeah huh

PR #690, opened 2022-02-22
Add 100% button when sending or debonding ROSE

buberdds:
Targets https://github.com/oasisprotocol/oasis-wallet-web/issues/492

Screenshot 2022-02-22 at 12 47 12
Screenshot 2022-02-22 at 12 48 26
Screenshot 2022-02-22 at 12 49 06


src/app/lib/helpers.ts
```diff
@@ -50,3 +50,5 @@ export function concat(...parts: Uint8Array[]) {
 }
 
 export const parseNumberToBigInt = (value: number) => BigInt(Math.round(value * 10 ** 9))
+
+export var parseBigIntToString = (value: string) => (Number(value) / 10 ** 9).toString()
```
pro-wh, line 54: could lose precision

src/app/components/ReclaimEscrowForm/index.tsx
```diff
@@ -59,29 +60,24 @@ export const ReclaimEscrowForm = memo((props: Props) => {
...
-            <TextInput
-              data-testid="amount"
-              id="amount-id"
-              name="amount"
-              placeholder={t('common.amount')}
-              type="number"
-              step="any"
-              min={0.0001}
+            <AmountTextInput
+              availableAmount={props.maxAmount}
```
pro-wh, line 66: potential confusion: min, max, value are in ROSE (or similar) units, availableAmount is in base units

actually are there use cases where this is not equal to the max?

buberdds: good point and thanks for help  I will circle back to the issue once oasis scan work is done 🖖

buberdds: - redundant availableAmount prop is removed
- min, max, value are passed to a new component in ROSE units

src/app/lib/helpers.ts
buberdds: I moved this from current ReclaimEscrowForm `Number(props.maxAmount) / 10 ** 9`  and we do this in other places as well ('AmountFormatter') . Can you give some examples so I can dig into that more?

pro-wh: ok if its moved from original code, I guess.

but fyi, for example `Number('10000000000000001')` ~10 mil ROSE goes wrong, due to limits of the number type

pro-wh (commented):
to the mathematical people, can we get an analysis of whether this would work? I think the goal is to be able to end up with exactly zero shares delegated, and for the chain to be able to delete that entry. (actually, is that the goal?) anyway, this method still goes through a measurement in ROSE, which we then convert to shares. not to mention, some of the calculation goes through floating point. will the rounding work out right? rounding up could lead to a failed transaction.

src/app/lib/helpers.ts
matevz: Could we finally push forward this feature? Did you consider this workaround:
1. increase number of decimals to 10 for the amount in ROSE field
2. 100% button takes the number of shares as the source information and then computes the amount in ROSE

matevz:
Um.. did we merge this? @buberdds?

buberdds:
I've closed this one for now due to long inactivity

PR #694, opened 2022-02-23
OasisScan: build scripts and config endpoints

buberdds:
- add oasisscan's URLs to config
- make api backend selectable during build process
- validator status - based on feedback from Tadej `status should  mean the same for both Oasis Monitor and Oasis Scan`

for local dev
`yarn start` (yarn start:oasismonitor alias)
`yarn start:oasisscan`

same for prod build
`yarn build` (yarn build:oasismonitor alias)
`yarn build:oasisscan`

Account and Validators endpoints work just fine when you run the app or switch between networks.


package.json
```diff
@@ -7,10 +7,14 @@
     "url": "https://github.com/Esya/oasis-wallet.git"
   },
   "scripts": {
-    "start": "react-scripts -r @cypress/instrument-cra start",
+    "start": "yarn start:oasismonitor",
+    "start:oasismonitor": "react-scripts -r ./internals/scripts/set-oasismonitor-backend-env.js -r @cypress/instrument-cra start",
+    "start:oasisscan": "react-scripts -r ./internals/scripts/set-oasisscan-backend-env.js -r @cypress/instrument-cra start",
```
lukaw3d, line 12: I would simplify to
```json
    "start": "react-scripts -r @cypress/instrument-cra start",
    "start:oasismonitor": "REACT_APP_BACKEND=oasismonitor yarn start",
    "start:oasisscan": "REACT_APP_BACKEND=oasisscan yarn start",
```

and `REACT_APP_BACKEND=oasismonitor` in .env, so `yarn start` still works

buberdds: Do we want to keep  "start:oasismonitor": task where it is a duplicate of a default one ?

lukaw3d: I think so. Default `start` uses REACT_APP_BACKEND from .env, but that can be overwritten by .env.local.

The explicit `start:oasismonitor` ensures REACT_APP_BACKEND is actually oasismonitor

PR #705, opened 2022-02-25
Replace RPC delegations with oasisscan API

lukaw3d:
Extends https://github.com/oasisprotocol/oasis-wallet-web/pull/704

### DIFF
--- a/src/vendors/monitor.ts
+++ b/src/vendors/monitor.ts
@@ -1,5 +1,7 @@
+import * as oasis from '@oasisprotocol/client'
+import { StakingDebondingDelegationInfo, StakingDelegationInfo } from '@oasisprotocol/client/dist/types'
 import { Account } from 'app/state/account/types'
-import { Validator } from 'app/state/staking/types'
+import { DebondingDelegation, Delegation, Validator } from 'app/state/staking/types'
 import { Transaction, TransactionType } from 'app/state/transaction/types'
 import {
   AccountsApi,
@@ -12,6 +14,7 @@ import {
   ValidatorCommissionScheduleRates,
   ValidatorRow,
 } from 'vendors/explorer'
+import { addressToPublicKey } from 'app/lib/helpers'
 
 import { sortByStatus } from './helpers'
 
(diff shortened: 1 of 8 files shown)


src/vendors/oasisscan.ts
```diff
@@ -103,3 +121,25 @@ export function parseTransactionsList(transactionsList: OperationsRow[]): Transa
...
+      shares: parseStringValueToInt(delegation.shares).toString(),
+      validatorAddress: delegation.validator_address,
+    }
+    return parsed
+  })
+}
+export function parseDebonding(debonding: DebondingDelegationRow[]): DebondingDelegation[] {
+  return debonding.map(debonding => {
+    const parsed: DebondingDelegation = {
+      amount: (parseStringValueToInt(debonding.shares) * 1).toString(),
```
lukaw3d, line 138: When debonding, are shares always equal to amount?
[Debonding API] only returns shares, unlike [delegations API]

[Debonding API]: https://api.oasisscan.com/mainnet/chain/account/debonding?address=oasis1qzah5wn48ekakmq5405qvcg4czp8hjvrcvcywvhp&size=500
[delegations API]: https://api.oasisscan.com/mainnet/chain/account/delegations?address=oasis1qzah5wn48ekakmq5405qvcg4czp8hjvrcvcywvhp&size=500

src/vendors/oasisscan.ts
```diff
@@ -42,7 +45,22 @@ export function getOasisscanAPIs(url: string | 'https://api.oasisscan.com/mainne
     return parseTransactionsList(transactionsList.data.list)
   }
 
-  return { getAccount, getAllValidators, getTransactionsList }
+  async function getDelegations(params: { accountId: string; nic: oasis.client.NodeInternal }): Promise<{
+    delegations: Delegation[]
+    debonding: DebondingDelegation[]
+  }> {
+    const delegations = await accounts.getDelegations({ address: params.accountId, size: 500 })
+    const debonding = await accounts.getDebondingDelegations({ address: params.accountId, size: 500 })
```
lukaw3d, line 53: TODO: warning about delayed data

src/vendors/oasisscan.ts
lukaw3d: related code in oasisscan:
- [delegations calculating amount](https://github.com/bitcat365/oasisscan-backend/blob/f1dafe3/src/main/java/romever/scan/oasisscan/service/AccountService.java#L246-L251)
- [debondings not calculating](https://github.com/bitcat365/oasisscan-backend/blob/f1dafe3/src/main/java/romever/scan/oasisscan/service/AccountService.java#L210)

lukaw3d: Since Oasisscan is slow about improving this, I added TODO comment. I would merge with this approximation :shrug:

Ready for review

src/vendors/monitor.ts
```diff
@@ -130,3 +151,51 @@ export function parseTransactionsList(transactionsList: OperationsRow[]): Transa
     return parsed
   })
 }
+
+function getSharePrice(pool: oasis.types.StakingSharePool) {
+  const balance = Number(oasis.quantity.toBigInt(pool.balance!)) / 10 ** 9
```
buberdds, line 156: Will refer question to @pro-wh, because he usually helps me with such stuff.  Can `getSharePrice` lose precision: parsing big int to number, or we are safe here, or we don't care about it here ?

src/vendors/oasisscan.ts
```diff
@@ -108,3 +126,26 @@ export function parseTransactionsList(transactionsList: OperationsRow[]): Transa
...
+      validatorAddress: delegation.validator_address,
+    }
+    return parsed
+  })
+}
+export function parseDebonding(debonding: DebondingDelegationRow[]): DebondingDelegation[] {
+  return debonding.map(debonding => {
+    const parsed: DebondingDelegation = {
+      // TODO: use amount field when it is available
+      amount: (parseStringValueToInt(debonding.shares) * 1).toString(),
```
buberdds, line 144: `parseStringValueToInt(debonding.shares) * 1).toString()` 
we multiple a string value to make it a number and then we cast it to string again. Do I understand it correctly?

src/vendors/monitor.ts
pro-wh: the share price is a rational quantity, and we've been using plain old number to represent those, approximately as you point out

src/vendors/oasisscan.ts
pro-wh: the `1` I think is a placeholder for when we get proper debonding pool share price information in the future

src/vendors/monitor.ts
lukaw3d: We should convert all such math into bignumber https://github.com/oasisprotocol/oasis-wallet-web/issues/436#issuecomment-982186445

even low values can cause ugliness that I would rather not solve by rounding

src/vendors/oasisscan.ts
lukaw3d: Made it more explicit to avoid confusion :)
```ts
    // TODO: use amount field, or share price when it is available. Until then,
    // using price=1 is inaccurate if debonding pool gets slashed.
    const sharePrice = 1
    const parsed: DebondingDelegation = {
      amount: (parseStringValueToInt(debonding.shares) * sharePrice).toString(),
```

lukaw3d:
Bump

buberdds (approved):

PR #718, opened 2022-02-25
Update Fatal Error message

buberdds:
Switch to new Support ticketing system

### DIFF
--- a/src/app/components/FatalErrorHandler/index.tsx
+++ b/src/app/components/FatalErrorHandler/index.tsx
@@ -46,8 +46,7 @@ export function FatalErrorHandler(props: Props) {
             <Trans
               i18nKey="fatalError.instruction"
               t={t}
-              components={[<Anchor href="https://github.com/oasisprotocol/oasis-wallet-web/issues" />]}
-              defaults="A fatal error has occurred and Oasis Wallet must stop. Try refreshing the page and reopening your wallets to see if the issue persists. Please file an issue to our <0>github project</0> and attach the report below."
+              components={[<Anchor href="mailto:wallet@oasisprotocol.org" />]}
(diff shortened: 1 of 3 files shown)


src/app/components/FatalErrorHandler/index.tsx
```diff
@@ -46,8 +46,8 @@ export function FatalErrorHandler(props: Props) {
             <Trans
               i18nKey="fatalError.instruction"
               t={t}
-              components={[<Anchor href="https://github.com/oasisprotocol/oasis-wallet-web/issues" />]}
-              defaults="A fatal error has occurred and Oasis Wallet must stop. Try refreshing the page and reopening your wallets to see if the issue persists. Please file an issue to our <0>github project</0> and attach the report below."
+              components={[<Anchor href="mailto:wallet@oasisprotocol.org" />]}
+              defaults="A fatal error has occurred and Oasis Wallet must stop. Try refreshing the page and reopening your wallets to see if the issue persists. Engage directly with our support team regarding issues via <0>wallet@oasisprotocol.org</0> email. Please attach the report below."
```
tjanez, line 50: A fatal error has occurred and Oasis Wallet must stop. Try refreshing the page and reopening your wallets to see if the issue persists. If the issue persists, please contact our support team via <0>wallet@oasisprotocol.org</0> email and attach the report below.

tjanez (commented):
Also, would it be possible to not have the message text duplicated for the translation or is that the case with all messages?

buberdds:
Turns out that we don't have to use `defaults` prop. Missing lang entry will default to `en/translation.json`. Duplication removed.

tjanez (approved):

PR #734, opened 2022-03-04
Dump validators list from oasisscan and use as fallback

lukaw3d:
Part of https://github.com/oasisprotocol/oasis-wallet-web/issues/751
Part of https://github.com/oasisprotocol/oasis-wallet-web/issues/749

Currently, oasisscan is down, and oasismonitor is not loading validators.

| Before | After |
| --- | --- |
| image  | localhost_3000_account_oasis1qp25k3hv6ysrm6wntjnt6x0j5du0ud4qqchplrf2_stake (1) |
|  |  localhost_3000_account_oasis1qp25k3hv6ysrm6wntjnt6x0j5du0ud4qqchplrf2_stake (2) |

### DIFF
--- a/src/app/state/staking/index.ts
+++ b/src/app/state/staking/index.ts
@@ -4,11 +4,17 @@ import { useInjectReducer, useInjectSaga } from 'utils/redux-injectors'
 
 import { stakingSaga } from './saga'
 import { DebondingDelegation, Delegation, StakingState, Validator, ValidatorDetails } from './types'
+import * as dump_validators from 'vendors/oasisscan/dump_validators.json'
 
 export const initialState: StakingState = {
   debondingDelegations: [],
   delegations: [],
-  validators: [],
+  validators: dump_validators.list.map(v => {
+    return {
+      ...v,
+      status: 'inactive', // In case validators are stale
+    }
+  }),
   updateValidatorsError: null,
   selectedValidatorDetails: null,
   selectedValidator: null,
--- a/src/app/state/staking/index.ts
+++ b/src/app/state/staking/index.ts
@@ -12,7 +12,7 @@ export const initialState: StakingState = {
   validators: dump_validators.list.map(v => {
     return {
       ...v,
-      status: 'inactive', // In case validators are stale
+      status: 'unknown',
     }
   }),
   updateValidatorsError: null,
--- a/src/app/state/staking/index.ts
+++ b/src/app/state/staking/index.ts
@@ -9,12 +9,15 @@ import * as dump_validators from 'vendors/oasisscan/dump_validators.json'
 export const initialState: StakingState = {
   debondingDelegations: [],
   delegations: [],
-  validators: dump_validators.list.map(v => {
-    return {
-      ...v,
-      status: 'unknown',
-    }
-  }),
+  validators: {
+    timestamp: dump_validators.dump_timestamp,
+    list: dump_validators.list.map(v => {
+      return {
+        ...v,
+        status: 'unknown',
+      }
+    }),
+  },
   updateValidatorsError: null,
   selectedValidatorDetails: null,
   selectedValidator: null,
@@ -31,7 +34,10 @@ const slice = createSlice({
     fetchAccount(state, action: PayloadAction<string>) {},
     updateValidators(state, action: PayloadAction<Validator[]>) {
       state.updateValidatorsError = null
-      state.validators = action.payload
+      state.validators = {
+        timestamp: Date.now(),
+        list: action.payload,
+      }
     },
     updateValidatorsError(state, action: PayloadAction<string>) {
       state.updateValidatorsError = action.payload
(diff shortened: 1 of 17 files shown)


src/vendors/smartstake/validators_dump.js
```diff
@@ -0,0 +1,6932 @@
...
+  oasisBlockHeight: '7510677',
+  notification: null,
+  statusSummary: { InConsensus: '111', NotInConsensus: '81', AllVal: 192 }
+}
+
+export const validators = validatorsJson.data.map((v, ix) => {
+  return {
+    name: v.entityName ?? undefined,
+    address: v.idAddress,
+    escrow: v.total * 10 ** 9,
```
lukaw3d, line 6928: I could also add `current_rate: v.commission != null ? v.commission / 100 : undefined` but that is a bit more sensitive if it is out of date

tjanez: In combination with a general notification like "Showing validator list as of <DATE-AND-TIME>" I think it would be fine to show it.

tjanez (approved):
I like this approach. I would perhaps just add a general notification like "Showing validator list as of <DATE-AND-TIME>".

But that can be done in a follow-up PR.

src/app/state/staking/index.ts
```diff
@@ -4,11 +4,12 @@ import { useInjectReducer, useInjectSaga } from 'utils/redux-injectors'
 
 import { stakingSaga } from './saga'
 import { DebondingDelegation, Delegation, StakingState, Validator, ValidatorDetails } from './types'
+import * as validatorsDump from 'vendors/smartstake/validators_dump'
 
 export const initialState: StakingState = {
   debondingDelegations: [],
   delegations: [],
-  validators: [],
+  validators: validatorsDump.validators,
```
Yawning, line 12: So with this users can end up delegating to a stale validator that may no longer exist?  At a minimum staking operations while this data is being provided to users should be limited to debonding.

NACK.

Yawning (changes requested):

src/app/state/staking/index.ts
tjanez: Fair point, agreed

src/app/state/staking/index.ts
```diff
@@ -4,11 +4,12 @@ import { useInjectReducer, useInjectSaga } from 'utils/redux-injectors'
 
 import { stakingSaga } from './saga'
 import { DebondingDelegation, Delegation, StakingState, Validator, ValidatorDetails } from './types'
+import * as validatorsDump from 'vendors/smartstake/validators_dump'
```
lukaw3d, line 7: Another issue: this list is inaccurate for testnet

src/app/state/staking/index.ts
lukaw3d: Mitigated by marking them all `inactive` for now, but I don't intend to merge image

buberdds: imo instead of changing initial state we should fallback in saga when a request fails. instead of dispatching `updateValidatorsError` action we can dispatch 
`yield* put(stakingActions.updateValidators(validatorsDump.validators))`

lukaw3d: One difference in that case (without extra handling): if getting validators first succeeds, but fails second time, result is:
- fallback on error: fallback
- fallback in initial state: first success

buberdds: ah true, is it possible that between the first and the second request validator could be updated? First success -> validator A works just fine -> something bad happens to validator A -> the second request to monitor fails  -> we show `fallback in initial state: first success` where validator A is Ok -> user delegates to a stale validator.

lukaw3d: Yeah, it is possible

src/vendors/smartstake/validators_dump.js
lukaw3d: Added
localhost_3000_account_oasis1qp25k3hv6ysrm6wntjnt6x0j5du0ud4qqchplrf2_stake (1)

src/app/state/staking/index.ts
lukaw3d: Added a warning before staking to validators with status inactive* or unknown:

localhost_3000_account_oasis1qp25k3hv6ysrm6wntjnt6x0j5du0ud4qqchplrf2_stake (2) 

---

\* Active/inactive meaning [oasisscan/swagger.yml#L306-L309](https://github.com/oasisprotocol/oasis-wallet-web/blob/8c7b9d5/src/vendors/oasisscan/swagger.yml#L306-L309):
> If "true", an entity has a node that is registered for being a validator, node is up to date, and has successfully registered itself. However, it may or may not be part of validator set (top <scheduler.params.max_validators> by stake).

Yawning: While the warning does make things clearer, I am still torn on this because it still forces people that delegate to something not working, into the debonding period.  I am still of the opinion that instead of a warning this should just flat out reject the operation.

Inevitably someone will ignore the warning, and later complain once their tokens get stuck in debonding.  While it is on them, forcing people to wait till things recover (in what should be an increasingly rare occurence in the future), is likely better.

If people really can't wait to delegate the command line tools will always work...

PR #736, opened 2022-03-05
Ignore API timeouts when checking transaction methods

lukaw3d:
Our CI was failing when oasisscan API wasn't responding: https://github.com/oasisprotocol/oasis-wallet-web/runs/5427977855?check_suite_focus=true#step:12:43

And cypress doesn't have a way to ignore timeouts :facepalm: so refactored into jest

### DIFF
--- a/cypress/integration/api-transaction-methods.ts
+++ b/cypress/integration/api-transaction-methods.ts
@@ -1,27 +1,48 @@
 import * as monitor from 'vendors/monitor'
 import * as oasisscan from 'vendors/oasisscan'
 
+function ignoreTimeoutError() {
+  // Note: This is strongly discouraged. If it stops working, we converted to
+  // jest in https://github.com/oasisprotocol/oasis-wallet-web/pull/736
+  cy.on('fail', error => {
+    if (error.name === 'CypressError' && error.message.includes('timed out waiting')) {
+      return false
+    }
+    throw error
+  })
+}
+
 describe('check all transaction methods are handled', () => {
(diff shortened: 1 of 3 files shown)


buberdds:
I would keep e2e tests separated from unit tests. Not Cypress expert, but it seems like custom error handler will work for us:

```
const skipTimeoutError = () => cy.on('fail', (error) => {
  if (error.name === 'CypressError' && error.message.includes('timed out waiting')) {
    return false  
  }
})

describe('check all transaction methods are handled', () => {
  it('by parsing latest transactions on oasismonitor', () => {
    skipTimeoutError()
    
    cy.request({
      url: 'https://monitor.oasis.dev/data/transactions?limit=2000',
      retryOnNetworkFailure: false,
      timeout: 5000
    })
```

and the same for scan

```
  it('by comparing list of methods from oasisscan', () => {
    skipTimeoutError()

    cy.request({
      url: 'https://api.oasisscan.com/mainnet/chain/methods',
      retryOnNetworkFailure: false,
      timeout: 5000
    })
 ```
 
When `timeout: 1` (for testing) a call to our BE will fail with `CypressError: `cy.request()` timed out waiting `1ms` for a response from your server.` (same as CI error), but test will continue to run and we will hit our `isOkStatusCode` check.

lukaw3d:
@buberdds Nice :open_mouth:  that is "strongly discouraged", but I like it

buberdds (approved):

PR #738, opened 2022-03-07
Update Grommet and Grommet icons

buberdds:
I went through Grommet changes and was testing the app manually. Snapshot changes are related to icon package (aria role fix/updated dark theme icon).


pro-wh (approved):
approving, although I didn't test on my side

and I guess this means we now have two versions of prop-types in our dependency tree, why is that? does something need precisely v15.6?

buberdds:
@pro-wh 
> and I guess this means we now have two versions of prop-types in our dependency tree, why is that? does something need precisely v15.6?

I did run `yarn-deduplicate` to clean up prop-types in lock file. I will test everything tomorrow EU morning again. At some point it looks like we will need to switch to yarn v2 which supports package deduplication by default when you run `yarn install`.

pro-wh:
interesting, thanks

PR #740, opened 2022-03-08
Fix modal light theme styles

buberdds:
Add missing `dark` boolean prop to handle light theme styles correctly.

Screenshot 2022-03-08 at 13 19 55
 
vs

Screenshot 2022-03-08 at 13 20 07

### DIFF
--- a/src/app/pages/OpenWalletPage/Features/FromLedger/index.tsx
+++ b/src/app/pages/OpenWalletPage/Features/FromLedger/index.tsx
@@ -100,7 +100,6 @@ export function FromLedgerModal(props: FromLedgerModalProps) {
   const { t } = useTranslation()
   const ledgerActions = useLedgerSlice().actions
   const walletActions = useWalletSlice().actions
-
   const ledger = useSelector(selectLedger)
   const error = ledger.error
   const selectedAccounts = useSelector(selectSelectedLedgerAccounts)
@@ -150,7 +149,7 @@ export function FromLedgerModal(props: FromLedgerModalProps) {
             }}
             background={{
               color: 'status-error',
-              opacity: 0.3,
+              opacity: 'weak',
             }}
             pad={{ horizontal: 'small', vertical: 'xsmall' }}
           >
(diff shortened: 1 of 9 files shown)


src/app/pages/OpenWalletPage/Features/FromLedger/index.tsx
```diff
@@ -150,6 +151,7 @@ export function FromLedgerModal(props: FromLedgerModalProps) {
             }}
             background={{
               color: 'status-error',
+              dark: theme === 'dark',
```
lukaw3d, line 154: Strange, I thought this would be inherited

lukaw3d: Seems like it tries to detect color darkness, but doesn't take into account the opacity `color: 'status-error', opacity: 0.3,`. I think we should
```diff
-      'status-error': '#ff4212',
+      'status-error': '#ff42124c', // 30% opacity

-  color: 'status-error', opacity: 0.3, [dark: theme === 'dark',]
+  color: 'status-error',
```

lukaw3d: (and same with other code involving `opacity:`)

buberdds: I think `status-error` works just fine, but inner components (here hyperlinks) are inheriting wrong theme which defaults to dark.

buberdds: Updated after a meeting
- use 'weak' workaround 
- unify opacity for all boxes

Change set is big, because I added ThemeProvider to one of a snapshot. Currently snapshots have default Grommet theme css rules not ours. I can remove that test update, but I am not sure it's worth having snapshots with css rules that we are not using. 


```
  // old snapshot
  background-color: rgba(255,170,21,0.2); // that's default Grommet  statusColors for example
  border-left: solid 3px #FFAA15;

  // snapshot with ThemeProvider
  background-color: rgba(244,206,75,0.25);
  border-left: solid 3px #f4ce4b;
```

lukaw3d (approved):

src/app/pages/OpenWalletPage/Features/FromLedger/index.tsx
lukaw3d: Nice

I don't know the main reason for snapshots with css, but it does help detect visual changes from updating dependencies :shrug:

PR #743, opened 2022-03-10
Use Parcel to build wallet

buberdds:
Targeting side branch as this will need more work (CI failure). I will also split missing pieces into separated tasks to address:

- check cypress


internals/set-parcel-build-variables.js
```diff
@@ -0,0 +1,11 @@
+const buildEnv = Object.create(process.env)
+
+const { buildTime, sha } = require('./getBuildData.js')
+const { csp } = require('./getCsp.js')
+
+buildEnv.REACT_APP_BYPASS_LOCAL = 1
+buildEnv.REACT_APP_BUILD_TIME = buildTime
```
lukaw3d, line 7: This name is confusing me :D  I'd prefer either:
- REACT_APP_BUILD_DATE, buildDate
- REACT_APP_BUILD_DATETIME, buildDatetime

internals/set-parcel-build-variables.js
```diff
@@ -0,0 +1,11 @@
+const buildEnv = Object.create(process.env)
+
+const { buildTime, sha } = require('./getBuildData.js')
+const { csp } = require('./getCsp.js')
+
+buildEnv.REACT_APP_BYPASS_LOCAL = 1
```
lukaw3d, line 6: (This could go into `.env.production`)

(And we can simplify `process.env.NODE_ENV && process.env.NODE_ENV !== 'production' && !process.env.REACT_APP_BYPASS_LOCAL` checks to `!process.env.REACT_APP_BYPASS_LOCAL`)

internals/prod-scripts.js
```diff
@@ -0,0 +1,4 @@
+const execSync = require('child_process').execSync
+const { buildEnv } = require('./set-parcel-build-variables.js')
+
+execSync('yarn parcel:build', { env: buildEnv, stdio: 'inherit' })
```
lukaw3d, line 4: One simpler, but not cross-platform solution would be:
```json
"build": "REACT_APP_META_CSP=`yarn --silent print-csp` REACT_APP_BUILD_VERSION=`git rev-parse HEAD` REACT_APP_BUILD_TIME=`date +%s` yarn parcel:build"
```

internals/set-parcel-build-variables.js
buberdds: oopos thanks, I left it here by accident actually

internals/prod-scripts.js
buberdds: It's not working on my OS. I would keep cross-platform support and it's a bit easier to maintain

internals/set-parcel-build-variables.js
buberdds: I think I don't want to have `REACT_APP_BYPASS_LOCAL` env at all here and especially in env.production. 

side note note related to Parcel migration: I would actually remove these conditions in `networkSaga` and `initialState` completely one day to limit build artefacts differences between envs.

buberdds:
Will merge changes into side branch and create PR that targets develop. Not sure why switching base branch did not trigger CI.

internals/set-parcel-build-variables.js
lukaw3d: We can make default behaviour equal to REACT_APP_BYPASS_LOCAL=1, but some tests run against local network - don't think we can avoid different build

buberdds: Yep at least in one case it will exist in out code where we add an item to a network dropdown. Anyway, can we keep behaviour the same compared to current build for now?

PR #746, opened 2022-03-11
Use Parcel to build wallet

buberdds:
Replace react-script with Parcel build. There are some places for improvements - added these to our internal board. Was aiming for 1:1 migration so on a daily basis commands are the same, CI is happy as well. Small code coverage change is related to removal of PUBLIC_URL which we were not using.

Bad:
- customisation of "no config tools" is a pain sometimes. Plugins are not widely adopted and some of them looks abandon
- npm trend does not look promising https://www.npmtrends.com/esbuild-vs-parcel-vs-rollup-vs-webpack

Good:
- bundle is much smaller this should result in better user exp (load, parsing etc)
- switching to another build tool in the future should be much simpler as we extracted cypress, jest, lint, code coverages reports
- we did not use many of react-scripts (like postCSS) tools so deps tree is now much smaller

@tjanez once everyone is happy we will have to coordinate releases to stage and prod including stable branch creation before merging into master


src/app/components/Sidebar/index.tsx
```diff
@@ -102,7 +102,7 @@ const SidebarHeader = (props: SidebarHeaderProps) => {
     >
       <Link to="/">
         <Box align="center" direction="row" gap="small">
-          <Avatar src={process.env.PUBLIC_URL + '/logo192.png'} size={sizeLogo[size]} />
+          <Avatar src="/logo192.png" size={sizeLogo[size]} />
```
lukaw3d, line 105: This seems like it shouldn't work. When bundling assets often get a hash filename. https://parceljs.org/recipes/image/#javascript says you need to wrap it into `new URL`

package.json
```diff
@@ -7,18 +7,21 @@
     "url": "https://github.com/Esya/oasis-wallet.git"
   },
   "scripts": {
-    "start": "react-scripts -r ./internals/scripts/set-app-version-env.js -r @cypress/instrument-cra start",
+    "clean": "rm -rf build/ .parcel-cache",
+    "parcel": "parcel --port 3000 --target web --dist-dir build",
```
lukaw3d, line 11: This script name is annoying for debugging
`yarn parcel --help` runs `parcel --port 3000 --target web --dist-dir build --help`

package.json
```diff
@@ -178,5 +179,27 @@
...
+    "styles": "./src/styles",
+    "config": "./src/config",
+    "store": "./src/store",
+    "reportWebVitals": "./src/reportWebVitals.ts",
+    "vendors": "./src/vendors"
+  },
+  "targets": {
+    "web": {
+      "context": "browser",
+      "source": "src/index.html"
```
lukaw3d, line 196: I think Parcel has a bug, or `source` in `targets` doesn't do anything.
1. `yarn build` Done in **36.04s**
2. `yarn clean && npx parcel build --target web --dist-dir build` Same
3. `yarn clean && npx parcel build --target web --dist-dir build .` Same
4. `yarn clean && npx parcel build --target web --dist-dir build src/index.html` Built in **339ms**
5. `touch ~/yarn.lock` (create yarn.lock in one of the ancestor folders)
6. `yarn build` Takes a **really long time or errors**, I guess related to how much stuff is in ancestor folder
7. `yarn clean && npx parcel build --target web --dist-dir build` Same
8. `yarn clean && npx parcel build --target web --dist-dir build .` Same
9. `yarn clean && npx parcel build --target web --dist-dir build src/index.html` Built in **308ms**
10. `rm ~/yarn.lock`

The cause of 6., 7., 8. is wrong entries and wrong projectRoot in @parcel/core/lib/resolveOptions.js.

Do you get a large performance boost by specifying entry in command, too?

package.json
buberdds: renamed to `parcel:dev` and `parcel:prod`

src/app/components/Sidebar/index.tsx
buberdds: Plugin wasn't enabled in dev as I was testing prod build mainly. Anyway, works for both envs now. Build is keeping original file names for PNG and it works similar to current react-script build were PNG files are handled by raw loader. Caching is handled by AWS cache invalidation.

package.json
buberdds: Was playing with it a little bit and indeed there is an issue. If we care about use case where someone runs `touch ~/yarn.lock` then let's drop source in target and move it to command line. For performance boost I did not notice anything all builds take ~27s.

lukaw3d (approved):

jest.config.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
```
tjanez, line 1: Add comment with link to Jest docs documenting config syntax.

tjanez (approved):
Great work, thanks @buberdds!

buberdds:
@tjanez after applying new branch names `Mega-Linter / Mega-Linter` is failing with 
```
git.exc.GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git fetch origin refs/heads/develop:refs/remotes/origin/develop
  stderr: 'fatal: couldn't find remote ref refs/heads/develop'
 ```

Let me know if I have to open a new PR or there is something else I can do, thx.

PR #753, opened 2022-03-11
Enable intellisense for translations keys

lukaw3d:
Poll: Does anyone already have code-completion for translation keys? e.g. does your IDE expand:
```js
    t('account.oth', '')
//  t('account.otherTransaction.designation', '')
```

Because VSCode doesn't for me. But the following code refers to intellisense, so I don't know what it really does :shrug: 
- https://github.com/oasisprotocol/oasis-wallet-web/blob/59751cc512c7e7ba2b725cf8ddabf61ba23533b4/src/locales/translations.ts#L9-L15
- https://github.com/oasisprotocol/oasis-wallet-web/blob/59751cc512c7e7ba2b725cf8ddabf61ba23533b4/src/locales/i18n.ts#L18-L19
- https://github.com/oasisprotocol/oasis-wallet-web/blob/59751cc512c7e7ba2b725cf8ddabf61ba23533b4/src/locales/types.ts#L7

This PR enables code-completion with types


buberdds (approved):
that works for me as well, nice. Let's nuke `src/locales/translations.ts` and `src/locales/types.ts` in this case

ysz (approved):
LGTM!

lukaw3d:
Ah, I found what it does. It allows specifying translation keys through an object path. [i18n.test.ts](https://github.com/oasisprotocol/oasis-wallet-web/blob/e52a8bdbcc9116460499d81ffe2247bd6fc4af3c/src/locales/__tests__/i18n.test.ts#L12)
```ts
import { translations } from 'locales/translations'
t(translations.menu.home)
```

I think we can safely use strings, now that they can be strongly typed, and I'll add a type-only test to ensure it is definitely strongly typed.

Issue #747, opened 2022-03-11
Ensure block explorer link goes to the correct block explorer


tjanez:
### Oasis Scan backend

Currently, If one switches to Testnet, the block explorer links result in 404 since they link to Oasis Scan Mainnet block explorer.

Example steps to reproduce:
1. Open https://wallet.oasisprotocol.org/account/oasis1qqk86skatdsuhq8a5qcyt9s8pf37etjrfqsdce5p
2. Switch to Testnet
3. Click on the link (little i icon in the bottom right corner) for the following txn:
![Screenshot from 2022-03-11 17-29-44](https://user-images.githubusercontent.com/2518544/157908125-a48ef641-c3fe-4e07-b9cd-23884b2cb1ae.png)
4. It links to https://www.oasisscan.com/transactions/813cbbed20f0b7594d873b99a3f98171a7c93ae0b10fcfc17a082fbdbfda27e0, but it should link to https://testnet.oasisscan.com/transactions/813cbbed20f0b7594d873b99a3f98171a7c93ae0b10fcfc17a082fbdbfda27e0.

### Oasis Monitor backend

Example steps to reproduce:
1. Open a local development wallet with Oasis Monitor backend
  ```
   REACT_APP_BYPASS_LOCAL=1 yarn run start:oasismonitor
   ```
2. Open http://localhost:3000/account/oasis1qqk86skatdsuhq8a5qcyt9s8pf37etjrfqsdce5p
3. Click on the link (little i icon in the bottom right corner) for a transaction on the Mainnet. It currently links to Oasis Scan but It should link to Oasis Monitor.

For Testnet, Oasis Monitor currently misses that transaction (probably due to an indexing issue?):

![Screenshot from 2022-03-11 17-40-32](https://user-images.githubusercontent.com/2518544/157909867-c17301fe-eaad-4f5d-bda5-505e6fb72123.png)

But it should link to the appropriate Oasis Monitor Testnet page, e.g. https://testnet.oasismonitor.com/operation/813cbbed20f0b7594d873b99a3f98171a7c93ae0b10fcfc17a082fbdbfda27e0

PR #755, opened 2022-03-15
Ensure block explorer link goes to the correct block explorer

buberdds:
Closes https://github.com/oasisprotocol/oasis-wallet-web/issues/747

- handle block explorer links correctly for mainnet and testnet

- added ts-jest dependency to handle mock types nicely. I will create ticket on our board to remove this after Parcel migration. Currently due to old `react-scripts` we are using old Jest version. In the new one mocked function was moved from ts-test/utils into Jest.

### DIFF
--- a/src/app/components/Transaction/__tests__/index.test.tsx
+++ b/src/app/components/Transaction/__tests__/index.test.tsx
@@ -6,7 +6,9 @@ import * as React from 'react'
 import { Router } from 'react-router'
 import { createMemoryHistory } from 'history'
 import { Provider, useSelector } from 'react-redux'
+import { mocked } from 'ts-jest/utils'
 import { configureAppStore } from 'store/configureStore'
+import { BackendAPIs, backend } from 'vendors/backend'
 
 import { Transaction } from '..'
 import * as transactionTypes from 'app/state/transaction/types'
@@ -23,6 +25,7 @@ jest.mock('react-i18next', () => ({
   },
 }))
 
+jest.mock('vendors/backend')
 jest.mock('react-redux', () => ({
   ...jest.requireActual('react-redux'),
   useSelector: jest.fn(),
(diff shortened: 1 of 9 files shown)


src/app/components/Transaction/__tests__/index.test.tsx
```diff
@@ -6,7 +6,9 @@ import * as React from 'react'
 import { Router } from 'react-router'
 import { createMemoryHistory } from 'history'
 import { Provider, useSelector } from 'react-redux'
+import { mocked } from 'ts-jest/utils'
```
lukaw3d, line 9: Nice! We can improve types around other `jest.mock(` with this, later

src/app/components/Transaction/index.tsx
```diff
@@ -42,6 +44,14 @@ type TransactionDictionary = {
...
   referenceAddress: string
   transaction: transactionTypes.Transaction
+  network: NetworkType
+}
+
+const getBlockExplorerHref = network => {
+  const subdomain = network === 'testnet' ? 'testnet.' : ''
+  return isOasisScanEnabled()
+    ? `https://${subdomain}oasisscan.com/transactions/`
+    : `https://${subdomain}oasismonitor.com/operation/`
```
lukaw3d, line 54: I think this would fit better in https://github.com/oasisprotocol/oasis-wallet-web/blob/b264e85094407a02f3907ad8fb2ff04721b924ea/src/config.ts#L3-L21

buberdds: updated

for local I am returning empty strings, not sure it's Ok

lukaw3d: Maybe `http://localhost:9001/data/transactions?operation_id=` :shrug:

lukaw3d: That could work with `http://localhost:9001/data/transactions?operation_id={{txHash}}` +  `blockExplorerLink.replace('{{txHash}}', encodeURIComponent(transaction.hash))`, instead of appending /${encodeURIComponent(transaction.hash)}

src/config.ts
```diff
@@ -1,31 +1,40 @@
 import { BackendAPIs } from 'vendors/backend'
 
+const oasisScanHostname = 'oasisscan.com'
+const oasisMonitorBlockExplorerHostname = 'oasismonitor.com'
+
 export const config = {
   mainnet: {
     [BackendAPIs.OasisMonitor]: {
       explorer: 'https://monitor.oasis.dev',
+      blockExplorer: `https://${oasisMonitorBlockExplorerHostname}/operation`,
```
lukaw3d, line 10: I'm not a fan of this change

buberdds: reverted 👍

src/app/components/Transaction/index.tsx
buberdds: done

PR #757, opened 2022-03-15
Sort validators by validator status property

### LINKED ISSUE #750: Sort validators by their active/inactive status first, then by stake
In the current version, validators are sorted by stake, regardless of whether they are currently active validators or not.

It would be better to present users with active validators first.

One option would be to sort validators by their active/inactive status first (after #748 is implemented) and then by stake.
Another option would be to have 2 subpages, one would show active validators and the other inactive ones.

buberdds:
Closes #750 

my preference here is to go with memoized selector, another option would be to use DataTable sortFunction prop. UX would be the same. 

Screenshot 2022-03-15 at 17 07 05
0


src/app/state/staking/selectors.ts
```diff
@@ -15,3 +16,17 @@ export const selectSelectedAddress = createSelector([selectStaking], state => st
...
 export const selectDebondingDelegations = createSelector([selectStaking], state => state.debondingDelegations)
+
+const ValidatorStatusPriority = {
+  active: 1,
+  inactive: 2,
+}
+
+const sortValidatorsByStatus = (validators: Validator[]) => {
+  const array = [...validators]
+  return array.sort((a, b) => ValidatorStatusPriority[a.status] - ValidatorStatusPriority[b.status])
```
lukaw3d, line 27: In other code we use this pattern before sorting:
```ts
    validators
      // Always clone before sort so it doesn't mutate source
      .slice()
      .sort((a, b) => b.escrow_balance - a.escrow_balance)
```

lukaw3d, line 27: I would have put this in parseValidatorsList, where we sort by escrow. And comment that it maintains sort by escrow because sort is stable.

lukaw3d: Scratch that. Let's move both sort(escrow).sort(status) to selectors.
(I don't think we will have API sorting validators and paginating them anytime soon, so this code doesn't need to be close to API)

lukaw3d (approved):

src/app/state/staking/selectors.ts
buberdds: Logic moved to `parseValidatorsList` to keep `rank` logic for Monitor intact

src/vendors/oasisscan.ts
```diff
@@ -53,23 +53,26 @@ export function parseAccount(account: AccountsRow): Account {
...
-        twitter_acc: v.twitter ?? undefined,
-        website_link: v.website ?? undefined,
-      },
-      rank: v.rank,
-    }
-    return parsed
-  })
+  return validators
+    .slice()
+    .sort((a, b) => Number(b.status) - Number(a.status))
```
lukaw3d, line 58: This isn't readable. Needs a comment or refactor.

src/vendors/__tests__/__snapshots__/oasisscan.spec.ts.snap
```diff
@@ -158,32 +158,46 @@ Array [
...
-      "website_link": "https://www.b.com",
+      "website_link": "https://www.binance.com",
     },
-    "name": "b",
+    "name": "BinanceStaking",
     "rank": 2,
     "status": "active",
   },
+  Object {
+    "address": undefined,
```
lukaw3d, line 175: Something broke

src/vendors/oasisscan.ts
buberdds: I will just add helpers too vendors and use one sort function for both BEs

src/vendors/__tests__/__snapshots__/oasisscan.spec.ts.snap
buberdds: ahh I blindly copied `https://api.oasisscan.com/mainnet/validator/list?pageSize=3` response., but that's not what I need in tests. all clear

lukaw3d: options:
- manually convert fields to snake case
- convert with ValidatorRowFromJSON
- remove `modelPropertyNaming=snake_case` from `openapi-generator-cli`

PR #763, opened 2022-03-23
Link from web wallet back to GitHub repo

buberdds:
I am not using Grommet Anchor component, because it applies different styles to an icon and it all looks completely different than other icons used in a side bar.

Screenshot 2022-03-23 at 15 56 51

### DIFF
--- a/src/app/components/Sidebar/index.tsx
+++ b/src/app/components/Sidebar/index.tsx
@@ -13,7 +13,16 @@ import {
   Text,
   Tip,
 } from 'grommet'
-import { FormDown, Home, Language, LineChart, Logout, Menu as MenuIcon, Money } from 'grommet-icons/icons'
+import {
+  Github,
+  FormDown,
+  Home,
+  Language,
+  LineChart,
+  Logout,
+  Menu as MenuIcon,
+  Money,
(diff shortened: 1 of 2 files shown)


src/app/components/Sidebar/index.tsx
```diff
@@ -157,6 +166,11 @@ const SidebarFooter = (props: SidebarFooterProps) => {
           </Box>
         </Menu>
       </Box>
+      <Box align="center" pad="small">
+        <a href="https://github.com/oasisprotocol/oasis-wallet-web" target="_blank" rel="noopener">
```
lukaw3d, line 170: We already have two links to source :D  
image

buberdds: lol I will validate this task with Tadej now :D

buberdds: so the answer is "Just have both for the time being".

lukaw3d: okay :shrug:

lukaw3d, line 170: on wider window:
image

lukaw3d (approved):

PR #498, opened 2021-11-30
Disable minification in production to get better stacktraces

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/495

PR #765, opened 2022-03-23
Use Parcel to build extension

buberdds:
Targeting side branch for extension work

- inject our CSP rules with extension related bits
- create separated task for unit tests (jest projects) and linters. We can run all of them or just a single project. Long term we can use this in CI where GH action/PR check could be triggered based on paths changes just to run necessary tests etc
- PR does not address e2e tests for extension. I think we will have to figure this out later. Cypress does not support other protocols than https/http, Playwright is limited to background page, here's sample for puppeteer with dynamic extension id https://github.com/oasisprotocol/oasis-wallet-web/commit/d4a784a9e73df9412ec415541e98fdddd2ba91a2, Metamask is using Selenium/Webdriver for example.


package.json
```diff
@@ -195,7 +201,18 @@
...
     "web": {
-      "context": "browser"
+      "context": "browser",
+      "distDir": "./build"
+    },
+    "ext-dev": {
+      "distDir": "./build-ext",
+      "sourceMap": {
+        "inline": true,
+        "inlineSources": true
```
lukaw3d, line 211: This is recommended by https://parceljs.org/recipes/web-extension/#hmr.
I had this comment workaround in proof of concept: https://github.com/oasisprotocol/wallet-poc/blob/redux-ext-poc/package.json#L13-L19

package.json
```diff
@@ -195,7 +201,18 @@
   },
   "targets": {
     "web": {
-      "context": "browser"
+      "context": "browser",
+      "distDir": "./build"
```
lukaw3d, line 205: ah, I missed this in https://github.com/oasisprotocol/oasis-wallet-web/pull/746: `build-without-minify.js` isn't used anymore and we didn't migrate.

We wanted better stack traces in bug reports: https://github.com/oasisprotocol/oasis-wallet-web/pull/498

package.json
```diff
@@ -88,6 +97,7 @@
     "husky": "7.0.4",
     "i18next": "20.6.1",
     "i18next-browser-languagedetector": "6.1.2",
+    "parcel-transformer-env-variables-injection": "0.1.0",
```
lukaw3d, line 100: Nice

internals/getCsp.js
```diff
@@ -3,8 +3,15 @@
 // - remove 'unsafe-inline' style by precomputing theme hash
 // - add report-uri to gather errors if anything was missed
 
+const cspEnhancement = `
+  frame-ancestors 
+    'self' 
+    https: http://localhost:* http://127.0.0.1:*;
+  `
```
lukaw3d, line 10: I wouldn't call it enhancement :D

lukaw3d (approved):

package.json
buberdds: would like to keep optimisation for prod env, but if that's must have there is a `--no-optimize` flag. I will create PR to target master branch.

internals/getCsp.js
buberdds: omg I forgot to push commit before merging. Anyway will change the variable name in a next extension PR.

PR #768, opened 2022-03-23
Hide link to bitpie private key export documentation

lukaw3d:


buberdds (approved):

matevz (changes requested):
I exported my private key from the Bitpie mobile wallet this morning, imported it into oasis cli and successfully sent some ROSE. The process described in the docs worked fine for me, so there needs to be something else. I suggest we leave the documentation as it is and investigate the error further (check bitpie wallet version, upgrade/downgrade, import the mnemonics again).

lukaw3d:
with further attempts, it seems to intermittently work :shrug:

Issue #526, opened 2021-12-07
Self-host Google fonts


matevz:
Google fonts downloaded by the web wallet leak privacy information to Google when folks use the wallet and from where. We should make the web wallet self-hosted including the fonts. The Google license permits this. This will also simplify and reduce some CSP rules.

PR #772, opened 2022-03-24
Self-hosted fonts

buberdds:
Closes https://github.com/oasisprotocol/oasis-wallet-web/issues/526

- Tested using Chrome/FF/Safari
- I've removed  `Inter` web font completely, because we are not using it in our app. This also deprecates usage of `fontfaceobserver` dependency in the app


buberdds:
I forgot about CSP rules

buberdds:
e2e test is failing due to missing `governance.SubmitProposal`. That is covered in https://github.com/oasisprotocol/oasis-wallet-web/pull/776

package.json
```diff
@@ -201,7 +199,8 @@
   "parcel-namer-rewrite": {
     "developmentHashing": true,
     "rules": {
-      "(.*).png": "$1.png"
+      "(.*).png": "$1.png",
+      "(.*).woff(.*)": "$1.woff$2"
```
lukaw3d, line 203: this looks expensive and too broad (component.woffles.js)

```json
"(.*).woff": "$1.woff",
"(.*).woff2": "$1.woff2"
```

but this will probably go away after combining with https://github.com/oasisprotocol/oasis-wallet-web/pull/778/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L218

lukaw3d (approved):

package.json
buberdds: I will port removal of  `parcel-namer-rewrite`. Extension side branch is still in experimental state.

Issue #751, opened 2022-03-11
Check stale validators with RPC


lukaw3d:
Related to https://github.com/oasisprotocol/oasis-wallet-web/pull/734 and https://github.com/oasisprotocol/oasis-wallet-web/issues/748#issuecomment-1065494444

For stale validators, RPC can get up-to-date information of which validators are active:
```js
const validators = nic.schedulerGetValidators(oasis.consensus.HEIGHT_LATEST)
  .then(a => a.sort((a,b) => Number(b.voting_power)-Number(a.voting_power)))

validators.forEach:
  const nodeAddress = await publicKeyToAddress(validator.id)

compare stale validators by node address, and set active/inactive
```

PR #776, opened 2022-03-25
Check stale validators with RPC

lukaw3d:
Resolves https://github.com/oasisprotocol/oasis-wallet-web/issues/751

Summary of new flow to fetch validators from comments:
- Skip if validators are already in store, and network didn't change
- Fetch validators from API
- If API fails on testnet/local, fall back to empty list
- If fetching dump_validators fails, fall back to empty list
- If API fails on mainnet, fall back to dump_validators, and refresh validators' status with RPC
- If RPC fails, fall back to dump_validators with unknown validators' status
- Fall back to dump_validators with refreshed validators' status from RPC


src/app/state/staking/saga.test.ts
```diff
@@ -118,4 +120,84 @@ describe('Staking Sagas', () => {
       ])
     })
   })
+
+  describe('Fetch validators fallbacks', () => {
```
lukaw3d, line 124: Strange. Tests were passing with `react-scripts test`

buberdds: react-script does not configure test timeouts as far as I am away. it's probably https://jestjs.io/docs/configuration#testtimeout-number `testTimeout`  or just jest.setTimeout(xyz) in setup, but in overall does that unit test is now making a  real call to RPC ?

lukaw3d: Nope, all mocked :shrug:

lukaw3d: I have a lead: if I upgrade to jest@27 on react-scripts I get the same error

lukaw3d: Apparently I must useRealTimers with expectSaga, so I'll need a different way to mock timestamp

buberdds: :/ seems like it is known issue in new Jest and sinon workaround works for me with your code. There is  legacy timer option too, but I did not look at it.
https://github.com/facebook/jest/issues/10221#issuecomment-983063883

lukaw3d: Just extracted into a `yield* call(now)` function. That's easy to mock :)

Much better:
```
      ✓ should load validators when switching network (3 ms)
      ✓ should use fallback on mainnet (33 ms)
```

src/app/state/staking/selectors.ts
```diff
@@ -6,12 +6,19 @@ import { initialState } from '.'
 const selectSlice = (state: RootState) => state.staking || initialState
 
 export const selectStaking = createSelector([selectSlice], state => state)
-export const selectValidators = createSelector([selectStaking], state => state.validators.list)
-export const selectValidatorsTimestamp = createSelector([selectStaking], state => state.validators.timestamp)
+export const selectValidators = createSelector([selectStaking], state => state.validators?.list ?? [])
+export const selectValidatorsTimestamp = createSelector(
+  [selectStaking],
+  state => state.validators?.timestamp ?? 0,
```
buberdds, line 12: When it can be undefined? Do we want to show 1970 in updateValidatorsError?

src/app/state/staking/saga.ts
```diff
@@ -82,16 +85,113 @@ function* loadDebondingDelegations(publicKey: Uint8Array) {
   return debondingDelegations
 }
 
-function* refreshValidators() {
+export function* maybeRefreshValidators() {
```
buberdds, line 88: Prefer a previous name without maybe, this one make me think that I have no idea what to expect.

buberdds (approved):

src/app/state/staking/saga.ts
lukaw3d: Alright :) Didn't know how to name it, after I added
```ts
  // Skip if validators are already in store, and network didn't change
  if (existingValidators.length > 0 && network === existingValidatorsNetwork) return
```

src/app/state/staking/selectors.ts
lukaw3d: 1970 should never appear with current code. It is undefined until first `maybeRefreshValidators` (initialState = { validators: null })

Annoying options:
1. keep default 0, so type is number
2. default to null, and force checking nullable everywhere we want to display timestamp. You have to check nullable even if we already check updateValidatorsError.
3. create selector for `state.validators`, so that it is a single variable for nullable checks. But it probably isn't best practice, not minimal for mocking, and redux wouldn't detect changes if subfield changed.

lukaw3d: Hm, wait, 2. is fine if we just use non-null assertion

PR #779, opened 2022-03-28
Dev dependencies updates: types, jest, testing-library

buberdds:
I will try to do some batch updates. Starting with simplest dev deps.

Other then just version bump I've removed `ts-jest` which is now not necessary as we are using the latest jest runner (part of parcel build switch)


pro-wh (approved):
oh cool, thanks for sorting out these new dependencies

PR #787, opened 2022-04-01
Use webext-redux in extension

buberdds:
To use `webext-redux` and reuse what we have in web in a new extension I think we need to:
- remove `connected-router`. While redux state was updated correctly,  popup routing was never affected by history push called from saga. Moreover, this lib is not working with newer router and react router team recommend to "not to keep your routes in your Redux store at all".
- redux-injectors are not compatible with Proxy story as they check store schema and Proxy is incompatible

It would be good if someone had time for similar research maybe we don't need such overhaul.


src/store/reducers.ts
```diff
@@ -3,19 +3,35 @@
...
+export function createReducer(injectedReducers = {}) {
   // Initially we don't have any injectedReducers, so returning identity function to avoid the error
-  return combineReducers({
+  const rootReducer = combineReducers({
     ...injectedReducers,
-    router: connectRouter(history),
+    account: accountReducer,
+    createWallet: createWalletReducer,
+    fatalError: fatalErrorReducer,
+    ledger: ledgerReducer,
```
lukaw3d, line 27: RootState type can now be non-optional
https://github.com/oasisprotocol/oasis-wallet-web/blob/5cebb40d55c0a51badaf23f46a9303781453e5a3/src/types/RootState.ts#L14-L23

src/app/useRouteRedirects.test.ts
```diff
@@ -0,0 +1,56 @@
...
+}))
+
+const mockedHistoryPush = jest.fn()
+jest.mock('react-router-dom', () => ({
+  useHistory: () => ({
+    push: mockedHistoryPush,
+  }),
+}))
+
+describe('useStartTrial', () => {
```
lukaw3d, line 18: Renamed to useRouteRedirects

lukaw3d (commented):
Nice!

src/utils/types/injector-typings.ts
```diff
@@ -1,22 +0,0 @@
-import { RootState } from 'types'
-import { Saga } from 'redux-saga'
-import { SagaInjectionModes } from 'redux-injectors'
-import { Reducer, AnyAction } from '@reduxjs/toolkit'
-
-type RequiredRootState = Required<RootState>
-
-export type RootStateKeyType = keyof RootState
```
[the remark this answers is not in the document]
buberdds: yep we need to fix `yarn checkTs` first https://github.com/oasisprotocol/oasis-wallet-web/issues/788

src/app/useRouteRedirects.tsx
```diff
@@ -0,0 +1,17 @@
...
+  const activeWalletIndex = useSelector(selectActiveWalletId)
+  const address = useSelector(selectAddress)
+  const history = useHistory()
+
+  useEffect(() => {
+    if (typeof activeWalletIndex !== 'undefined' && address) {
+      history.push(`/account/${address}`)
+    }
+  }, [activeWalletIndex, address, history])
+}
```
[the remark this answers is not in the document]
buberdds: ahh yes for "don't redirect" we will need to compare with prev value. I will move this task to clickup.

src/types/RootState.ts
```diff
@@ -9,24 +9,16 @@ import { TransactionState } from 'app/state/transaction/types'
 import { LedgerState } from 'app/state/ledger/types'
 import { StakingState } from 'app/state/staking/types'
 import { FatalErrorState } from 'app/state/fatalerror/types'
-// [IMPORT NEW CONTAINERSTATE ABOVE] < Needed for generating containers seamlessly
```
lukaw3d, line 12: Let's keep this https://github.com/oasisprotocol/oasis-wallet-web/blob/5cebb40d55c0a51badaf23f46a9303781453e5a3/internals/generators/slice/index.ts#L76

src/types/RootState.ts
```diff
@@ -9,24 +9,16 @@ import { TransactionState } from 'app/state/transaction/types'
...
-  wallet?: WalletState
-  createWallet?: CreateWalletState
-  openWallet?: OpenWalletState
-  account?: AccountState
-  network?: NetworkState
-  transaction?: TransactionState
-  ledger?: LedgerState
-  staking?: StakingState
-  fatalError?: FatalErrorState
-  // [INSERT NEW REDUCER KEY ABOVE] < Needed for generating containers seamlessly
```
lukaw3d, line 31: +

PR #790, opened 2022-04-04
Clean up unit tests warnings

buberdds:
PR removes two types of warnings that are thrown when unit tests are running

- `Warning: React.jsx: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined.`
- `Not implemented: window.scrollTo`

I've copied mock from `react-i18next` repo.  It fixes issue with `undefined` in all places where `Trans` component is used. Instead of adding mock in all places let's add it globally. We can always override it in each test file when needed (`CommissionBounds` is an example where test relies on interpolation/real implementation).

### DIFF
--- a/src/app/pages/StakingPage/Features/CommissionBounds/__tests__/index.test.tsx
+++ b/src/app/pages/StakingPage/Features/CommissionBounds/__tests__/index.test.tsx
@@ -8,6 +8,8 @@ import { configureAppStore } from 'store/configureStore'
 
 import { CommissionBounds } from '..'
 
+jest.unmock('react-i18next')
+
 const renderComponent = (store: any, bounds?: ICommissionBounds[]) =>
   render(
     <Provider store={store}>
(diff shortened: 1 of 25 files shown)


src/app/pages/StakingPage/Features/CommissionBounds/__tests__/index.test.tsx
```diff
@@ -8,6 +8,10 @@ import { configureAppStore } from 'store/configureStore'
 
 import { CommissionBounds } from '..'
 
+jest.mock('react-i18next', () => ({
+  ...jest.requireActual('react-i18next'),
+}))
```
lukaw3d, line 13: This looks like `jest.unmock('react-i18next')`, but I guess it matches:
```ts
jest.mock('react-redux', () => ({
  ...jest.requireActual('react-redux'),
  useDispatch: jest.fn(),
}))
```

lukaw3d (approved):
I don't like the commit name, and prefer if "scrollTo" was a separate commit

PR #810, opened 2022-04-19
Sort all translations and remove English defaults from translations

lukaw3d:
Related to https://github.com/oasisprotocol/oasis-wallet-web/pull/541

### DIFF
--- a/src/locales/de/translation.json
+++ b/src/locales/de/translation.json
@@ -1,16 +1,17 @@
 {
-  "routingFeature": {
-    "title": "Standard Routing",
-    "description": "Routing macht es möglich Seiten (z.B. '/about') Ihrer Anwendung hinzuzufügen."
+  "Hello world": "",
+  "feedbackFeature": {
+    "title": "Sofortiges Feedback",
+    "description": "Genießen Sie die beste Entwicklungserfahrung und programmieren Sie Ihre App so schnell wie noch nie! Ihre Änderungen an dem CSS und JavaScript sind sofort reflektiert, ohne die Seite aktualisieren zu müssen."
   },
   "i18nFeature": {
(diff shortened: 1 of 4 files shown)


matevz:
Ugh, I wouldn't change non-English translations directly outside of Transifex, because they might get out of sync. Let's wait for Transifex to sync the English source (daily) and see how the export goes. The ordering of terms should be according to English source when you export them.

matevz:
> Ugh, I wouldn't change non-English translations directly outside of Transifex, because they might get out of sync. Let's wait for Transifex to sync the English source (daily) and see how the export goes. The ordering of terms should be according to English source when you export them.

I checked Transifex today and the ordering of non-English terms was automatically synced with the new ordering in English source. However, I noticed some regressions in #541 and made a followup PR to address those in #813 (wrong terms at wrong places, some redundant periods and colons, outdated terms). Once #813 is merged, we should simply grab the translations from Transifex and push them to the main branch.

lukaw3d:
Sounds good

matevz:
Some updates:
- Slovenian language is ready to grab from Transifex.
- Turkish doesn't have a maintainer so you can just download it from Transifex as it is (with sorted terms).
- Raphael will translate the missing 20 terms in French in the following days, so you should wait for him and put it in this PR.
- Feel free to remove German leftovers from the codebase since it's just some i18next mock and has nothing to do with the actual translation.
- It seems the "Download for use" behavior in Transifex for JSON key-value format is "Replace untranslated terms with source". See https://docs.transifex.com/formats/json#section-parser-behaviour . I'm not sure why that is default, but it has to do something with i18next, because .po gettext format I use in other projects behaves differently.

matevz:
Slovenian, Turkish and French are now ready to be merged from Transifex.

PR #825, opened 2022-05-06
Stricter types in tests

lukaw3d:


src/app/state/transaction/types.ts
```diff
@@ -1,26 +1,22 @@
 import { ErrorPayload } from 'types/errors'
 
-/**
- * These are manually copied from Oasis-explorer. Later, oasis-explorer should
- * make those an enum so that we maintain strong typing across projects.
```
pro-wh, line 5: does anything rely on these matching up with oasis-explorer?

pro-wh (approved):
these seem like pretty sane changes

src/app/state/transaction/types.ts
lukaw3d: No; we map from oasis-explorer's OperationsRowTypeEnum to our TransactionType here: https://github.com/oasisprotocol/oasis-wallet-web/blob/825bac2d0ad22132032c69cf93615de65ca9fc16/src/vendors/monitor.ts#L123-L140

Oasis-explorer's OperationsRowTypeEnum is generated from https://github.com/oasisprotocol/oasis-wallet-web/blob/825bac2d0ad22132032c69cf93615de65ca9fc16/src/vendors/explorer/swagger.yml#L1125-L1141

and we have this test to try to ensure the enum in swagger.yml is accurate https://github.com/oasisprotocol/oasis-wallet-web/blob/825bac2d0ad22132032c69cf93615de65ca9fc16/cypress/integration/api-transaction-methods.ts#L15-L32

PR #844, opened 2022-06-01
Restyle SearchAddress focus and refactor

lukaw3d:
| Before | After |
| --- | --- |
|  before | after  |

### DIFF
--- a/src/app/components/Toolbar/Features/SearchAddress/SearchBox.tsx
+++ b/src/app/components/Toolbar/Features/SearchAddress/SearchBox.tsx
@@ -1,0 +1,60 @@
+import { Box, Button, TextInput } from 'grommet'
+import { FormClose, Search } from 'grommet-icons/icons'
+import React from 'react'
+import styled from 'styled-components'
+
+interface Props {
+  placeholder: string
+  value: string
+  onChange: (value: string) => void
+  onClear: () => void
+}
+
+const RightIconButton = styled(Button)`
+  margin-left: ${({ theme }) => {
+    const negativeIconButtonWidth = `calc(-${theme.global?.edgeSize?.small} - ${theme.icon?.size?.medium} - ${theme.global?.edgeSize?.small})`
+    return negativeIconButtonWidth
+  }};
+  z-index: 1;
+`
+
+/**
+ * Rounded input with search icon and clear button.
+ */
+export const SearchBox = React.forwardRef<HTMLInputElement, Props>((props, ref) => {
+  const { placeholder, value, onChange, onClear, ...rest } = props
+  const borderRadius = '30px'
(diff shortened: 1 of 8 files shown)


src/app/components/Toolbar/Features/SearchAddress/SearchBox.tsx
```diff
@@ -0,0 +1,51 @@
...
+          <Box margin={{ left: '36px' }} max-height="30px" style={{ textOverflow: 'ellipsis' }} flex="shrink">
+            {placeholder}
+          </Box>
+        }
+        ref={ref}
+        {...rest}
+      />
+      {value && (
+        <Button
+          style={{ marginLeft: '-48px', zIndex: 1, marginTop: '-2px', marginBottom: '-2px' }}
```
buberdds, line 44: Two ideas:
- we can avoid hardcoding all these pixels here , so component will be more 'stable' when updating Grommet or theme in the future
- input height is not aligned with network selection button height, it's more visible on mobile view

something like this may work:

```
const FooButton = styled(Button)`
  margin-left: ${({ theme }) => `-${theme.global.edgeSize.large}`}; // avoid hardcoded pixels
  position: relative;
`
```

```
      <FooButton
        style={{
          visibility: value ? 'visible' : 'hidden',  // avoid jumping
        }}
        icon={<FormClose />}
        onClick={() => onClear()}
      />
```

lukaw3d: Oh yeah, that's better!

I didn't know grommet's theme is in styled-components. Though `theme.global.edgeSize.large` isn't related. 48px comes from "x" button's width: `left-padding + "x" icon width + right-padding`.

Is there any benefit of `position: relative;` vs `z-index: 1;`?

buberdds: `Is there any benefit of position: relative; vs z-index: 1;?` 
I have no idea 😄

buberdds (approved):

src/app/components/Toolbar/Features/SearchAddress/SearchBox.tsx
```diff
@@ -1,6 +1,7 @@
 import { Box, Button, TextInput } from 'grommet'
 import { FormClose, Search } from 'grommet-icons/icons'
 import React from 'react'
+import styled, { useTheme } from 'styled-components'
```
buberdds, line 4: useTheme is not used

lukaw3d: oops :D 

I looked at some linting options to detect these, and now I've got 204 other problems

Issue #828, opened 2022-05-12
Combine `start` and `start:ext`


lukaw3d:
If we want a single `start` command for both web and ext we would need to:
- combine `.parcelrc` and `.parcelrc-ext` configs
- combine `ext-dev` and `ext-prod` targets
- restructure folders so that built folder has `/manifest.json`, `/_locales`, and `/oasis-xu-frame.html` on root

_Originally posted by @lukaw3d in https://github.com/oasisprotocol/oasis-wallet-web/pull/805#discussion_r871595122_

pro-wh:
what's the context of this? do we want to unify them?

lukaw3d:
After https://github.com/oasisprotocol/oasis-wallet-web/pull/805 we will have `yarn start` for serving web and `yarn start:ext` for serving extension. Running both is more annoying, slower, and potentially breaks .parcel-cache.

So I'd like to use parcel-bundler's option to serve multiple targets

PR #846, opened 2022-06-01
Combine `start` and `start:ext` to serve both at the same time

lukaw3d:
Closes https://github.com/oasisprotocol/oasis-wallet-web/issues/828


package.json
```diff
@@ -8,10 +8,9 @@
   },
   "scripts": {
     "clean": "rm -rf build/ .parcel-cache",
-    "start": "REACT_APP_META_CSP= parcel --port 3000 --target web --dist-dir build",
     "parcel:prod": "parcel build --target web --dist-dir build",
     "clean:ext": "rm -rf build-ext/ .parcel-cache",
-    "start:ext": "HMR_PORT=2222 EXTENSION_CSP=`yarn --silent print-extension-csp` parcel --host localhost --target ext --dist-dir build-ext",
+    "start": "REACT_APP_META_CSP= HMR_PORT=2222 EXTENSION_CSP=`yarn --silent print-extension-csp` parcel --host localhost --port 3000 --dist-dir build-dev",
```
ysz, line 13: How about we add `REACT_APP_BYPASS_LOCAL=1` such that `yarn start` works out of the box by default without local node?

ysz (approved):

package.json
lukaw3d: edit: REACT_APP_LOCALNET=

package.json
```diff
@@ -8,18 +8,17 @@
...
     "build:ext": "yarn clean:ext && EXTENSION_CSP=`yarn --silent print-extension-csp` parcel build --target ext --dist-dir build-ext && node ./internals/scripts/validate-ext-manifest.js",
```
lukaw3d, line 14: Hm. Extension is broken if built with `yarn build:ext`, but it works from `yarn start`

lukaw3d: In background page: `Uncaught Error: Cannot find module 'fj44w'`
background.8e475e86.js contains: `"webext-redux": "fj44w",`
popup.9c7b8803.js contains:  `"webext-redux": "fj44w",`
popup.c6ac691b.js contains: `fj44w: [ function(q, H, ee) {`

lukaw3d: Upstream issue https://github.com/parcel-bundler/parcel/issues/8071#issuecomment-1148069991

lukaw3d: Deferring to https://github.com/oasisprotocol/oasis-wallet-web/issues/853

PR #848, opened 2022-06-01
Add a tip about running CI in "Update dumped validators" PRs

lukaw3d:
https://github.com/oasisprotocol/oasis-wallet-web/pull/845 didn't run CI actions until closing and reopening it. Added a link to the explanation in https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs

### DIFF
--- a/.github/workflows/dump-validators.yml
+++ b/.github/workflows/dump-validators.yml
@@ -19,9 +19,15 @@ jobs:
       - run: yarn install --frozen-lockfile
       - run: yarn dump-validators
       - name: Create Pull Request with updated dumped validators
+        # https://github.com/peter-evans/create-pull-request
         uses: peter-evans/create-pull-request@v3
         with:
           token: ${{ secrets.PAT || secrets.GITHUB_TOKEN }}
           commit-message: 'Update dumped validators'
           title: 'Update dumped validators'
+          body: |
+            Created by https://github.com/oasisprotocol/oasis-wallet-web/blob/master/.github/workflows/dump-validators.yml
+
+            If CI actions and checks don't run in this PR: close it and reopen.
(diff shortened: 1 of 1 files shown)


.github/workflows/dump-validators.yml
```diff
@@ -19,9 +19,15 @@ jobs:
       - run: yarn install --frozen-lockfile
       - run: yarn dump-validators
       - name: Create Pull Request with updated dumped validators
+        # https://github.com/peter-evans/create-pull-request
```
aefhm, line 22: Going the extra mile here...

.github/workflows/dump-validators.yml
```diff
@@ -19,9 +19,15 @@ jobs:
...
         uses: peter-evans/create-pull-request@v3
         with:
           token: ${{ secrets.PAT || secrets.GITHUB_TOKEN }}
           commit-message: 'Update dumped validators'
           title: 'Update dumped validators'
+          body: |
+            Created by https://github.com/oasisprotocol/oasis-wallet-web/blob/master/.github/workflows/dump-validators.yml
+
+            If CI actions and checks don't run in this PR: close it and reopen.
```
aefhm, line 31: I'm still not following why the original one didn't run 🤔

aefhm (approved):

.github/workflows/dump-validators.yml
lukaw3d: https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs

Github made it that way :face_exhaling: 
> events triggered by the GITHUB_TOKEN will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs

That link lists 5 workarounds, and i think close + reopen is easiest

Issue #774, opened 2022-03-24
Fix megalinter to agree with eslint


lukaw3d:
My editor adds trailing commas to `./jest.config.js`. And `yarn eslint ./*.js` too. But megalinter disagrees.

why

PR #855, opened 2022-06-10
Stricter eslint config

lukaw3d:
Disabling Standard in MegaLinter fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/774


lukaw3d:
megalinter sucks.

This broke on master... https://github.com/oasisprotocol/oasis-wallet-web/runs/6834514056?check_suite_focus=true
```
...
/github/workspace/internals/generators/component/index.ts
Error:   11:38  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
/github/workspace/internals/generators/slice/index.ts
Error:   12:38  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
```

src/vendors/__tests__/monitor.test.ts
```diff
@@ -14,6 +14,7 @@ describe('monitor', () => {
       parseAccount({
         address: 'oasis1qq3xrq0urs8qcffhvmhfhz4p0mu7ewc8rscnlwxe',
         liquid_balance: 756455428396,
+        /* eslint-disable @typescript-eslint/no-loss-of-precision */
```
pro-wh, line 17: what's this situation here

lukaw3d: Oasismonitor is returning these numbers in JSON, but they lose precision when they are parsed as a number. 336982978187110627 rounds to 336982978187110660.

I had already warned @ennsharma and @aefhm for oasis-indexer: since oasisscan was using ROSE units formatted into string, there wasn't the same worry.

pro-wh: :cry:

PR #862, opened 2022-06-15
Use one dependency to inject env variables during build

buberdds:
We don't need `posthtml` dependency after adding ext build as we can rely on `transformer-env-variables`. This also removes a warning that is thrown during build process.


lukaw3d (approved):
Nice

PR #865, opened 2022-06-17
Only show SendTransaction on your accounts

lukaw3d:
Before: 
- user imports `oasis1qqug..hsu8` to wallet
- search `oasis1qr9t..gxfl`
- see SendTransaction component to transfer from `oasis1qqug..hsu8` to any _Recipient_
  oasis1qquga66n9mzypsayerxnkp8p343g5ufwny4dhsu8_to_oasis1qr9tkn7cg2rg95ps7td0ar3ghqru8y225s4tgxfl

### DIFF
--- a/src/app/pages/AccountPage/Features/SendTransaction/index.tsx
+++ b/src/app/pages/AccountPage/Features/SendTransaction/index.tsx
@@ -9,7 +9,15 @@ import { useEffect, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { useDispatch, useSelector } from 'react-redux'
 
-export function SendTransaction() {
+export interface SendTransactionProps {
+  isAddressInWallet: boolean
+}
+
+export function SendTransaction(props: SendTransactionProps) {
+  if (!props.isAddressInWallet) {
+    throw new Error('SendTransaction component should only appear on your accounts')
+  }
+
   const { t } = useTranslation()
   const dispatch = useDispatch()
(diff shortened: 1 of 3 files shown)


src/app/pages/AccountPage/Features/SendTransaction/index.tsx
```diff
@@ -9,7 +9,15 @@ import { useEffect, useState } from 'react'
...
 import { useDispatch, useSelector } from 'react-redux'
 
-export function SendTransaction() {
+export interface SendTransactionProps {
+  isAddressInWallet: boolean
+}
+
+export function SendTransaction(props: SendTransactionProps) {
+  if (!props.isAddressInWallet) {
+    throw new Error('SendTransaction component should only appear on your accounts')
```
buberdds, line 18: I get idea that entire app will blow up to catch an error, but I think it's better to cover such scenarios in unit tests than in a component via throwing new Error.

buberdds (approved):

src/app/pages/AccountPage/Features/SendTransaction/index.tsx
lukaw3d: I don't think tests can ensure that we devs don't move SendTransaction outside the condition:
```tsx
      {isAddressInWallet && (
        ...
          <SendTransaction
```
Maybe with integration tests of all possible user flows :shrug:

PR #867, opened 2022-06-20
Add latin ext font to render various languages correctly

### LINKED ISSUE #811: Fix rendering non-latin characters
Characters in translations are not necessarily latin, so we'll need more than just Latin in here: https://github.com/oasisprotocol/oasis-wallet-web/pull/772/files#diff-25d6a3ff17a1c8acb9b8e63e990597d94ef65730407b2e171edbf73b7515df3dR6

See google-hosted versions https://fonts.googleapis.com/css2?family=Roboto+Mono:wght@300&family=Rubik:wght@300&display=swap

From slack: Weird "č"
image

buberdds:
Closes #811


pro-wh (approved):
ah yea

didn't check the font files in reviewing this

lukaw3d (approved):
I see we are also missing licenses
- https://github.com/googlefonts/rubik/blob/main/OFL.txt
- https://github.com/google/fonts/blob/main/apache/robotomono/LICENSE.txt (but this one is weird)

When we'll need Cyrillic I'd rather have unicode-ranges instead of bundling everything together

buberdds:
@lukaw3d, what do you mean by "we are also missing licenses"? Should I upload font licenses somewhere or add links to these licenses ? Not sure OFL or Apache need anything like that 🤔

PR #868, opened 2022-06-21
Prevent race conditions when loading account

### LINKED ISSUE #861: Prevent race conditions
Our async logic has race conditions like:
- open some account, e.g. https://wallet.oasisprotocol.org/account/oasis1qramypd6nnuw2utvvjgcunwryut98may7c09mvuk
- search oasis1qqekv2ymgzmd8j2s2u7g0hhc7e77e654kvwqtjwm
- quickly click back
- bug: see wrong balance in oasis1qramypd6nnuw2utvvjgcunwryut98may7c09mvuk

buberdds:
Closes #861

### DIFF
--- a/src/app/state/account/saga.test.ts
+++ b/src/app/state/account/saga.test.ts
@@ -1,0 +1,15 @@
+import { testSaga } from 'redux-saga-test-plan'
+import { accountActions } from '.'
+import { accountSaga, loadAccount, refreshAccountOnTransaction } from './saga'
+
+describe('Account Sagas', () => {
+  test('accountSaga', () => {
+    testSaga(accountSaga)
+      .next()
+      .fork(refreshAccountOnTransaction)
+      .next()
+      .takeLatest(accountActions.fetchAccount, loadAccount)
+      .next()
+      .isDone()
+  })
+})
--- a/src/app/state/network/saga.test.ts
+++ b/src/app/state/network/saga.test.ts
@@ -1,0 +1,23 @@
+import { testSaga } from 'redux-saga-test-plan'
+import { networkActions } from '.'
+import { networkSaga, selectNetwork } from './saga'
+
+describe('Network Sagas', () => {
+  const env = process.env
+
+  afterEach(() => {
+    process.env = { ...env }
+  })
+
+  test('networkSaga', () => {
+    delete process.env.REACT_APP_LOCALNET
+
+    testSaga(networkSaga)
+      .next()
+      .takeLatest(networkActions.selectNetwork, selectNetwork)
+      .next()
+      .put(networkActions.selectNetwork('mainnet'))
+      .next()
+      .isDone()
+  })
+})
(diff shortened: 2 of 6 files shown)


src/app/state/account/saga.test.ts
```diff
@@ -0,0 +1,15 @@
...
+
+describe('Account Sagas', () => {
+  test('accountSaga', () => {
+    testSaga(accountSaga)
+      .next()
+      .fork(refreshAccountOnTransaction)
+      .next()
+      .takeLatest(accountActions.fetchAccount, loadAccount)
+      .next()
+      .isDone()
```
lukaw3d, line 13: I don't think this is a useful test.

I'd rather see approximately
```ts
expectSaga(stakingSaga)
  .provide(getAccount, () => delay(10*Math.random()))
  .put(fetchAccount('a'))
  .put(fetchAccount('b'))
  .hasFinalState({ address: 'b' })
  .run()
```

src/app/state/staking/saga.ts
```diff
@@ -200,6 +200,6 @@ export function* fetchAccount({ payload: address }: PayloadAction<string>) {
 }
 
 export function* stakingSaga() {
-  yield* takeEvery(stakingActions.fetchAccount, fetchAccount)
+  yield* takeLatest(stakingActions.fetchAccount, fetchAccount)
   yield* takeEvery(stakingActions.validatorSelected, getValidatorDetails)
```
lukaw3d, line 204: Race conditions elsewhere
- `stakingActions.validatorSelected` should be `takeLatest` - shows wrong info on race condition
- `networkActions.selectNetwork` would be more correct with `takeLatest` - switches to previous network on race condition 
- `ledgerActions.enumerateAccounts` with `takeLatest` should prevent Ledger crashes, but it doesn't https://github.com/oasisprotocol/oasis-wallet-web/issues/869

src/app/state/account/saga.test.ts
buberdds: I think this unit test if valid unless I am missing something, but I am following one of redux saga test plan pattern. I am adding a unit test not integration test. Unit test is testing a watcher saga only and it doesn't care about integration with other parts of code. The bug exists because of using a wrong effect creator. Unit test is validating if a correct effect creator is used so it covers current bug/potential regression bug.

src/app/state/network/saga.test.ts
```diff
@@ -0,0 +1,15 @@
...
+import { networkActions } from '.'
+import { networkSaga, selectNetwork } from './saga'
+
+describe('Network Sagas', () => {
+  test('networkSaga', () => {
+    testSaga(networkSaga)
+      .next()
+      .takeLatest(networkActions.selectNetwork, selectNetwork)
+      .next()
+      .put(networkActions.selectNetwork(process.env.REACT_APP_LOCALNET ? 'local' : 'mainnet'))
```
buberdds, line 11: @lukaw3d  Do we have some pattern for handling such cases? ('local' for CLI, 'mainnet' for dev)
Other options are: override process env in setupTests or locally in tests, right?

PR #874, opened 2022-06-24
Combine packages into groups in Renovate to reduce PR clutter

lukaw3d:
Whenever we merge a Renovate PR, all others get rebased and spam CI. So waiting for CI and merging one by one is slow. We can group some of these PRs by package name, and be happier until an update breaks things https://docs.renovatebot.com/noise-reduction/#package-grouping


buberdds (approved):

pro-wh (approved):
cool find, let's use that. can add/change groups if we have more ideas later

Issue #853, opened 2022-06-08
Extension built with `yarn build:ext` is broken


lukaw3d:
Hm. Extension is broken if built with `yarn build:ext`, but it works from `yarn start`

In background page: Uncaught Error: Cannot find module 'fj44w'
background.8e475e86.js contains: "webext-redux": "fj44w",
popup.9c7b8803.js contains: "webext-redux": "fj44w",
popup.c6ac691b.js contains: fj44w: [ function(q, H, ee) {

Upstream issue https://github.com/parcel-bundler/parcel/issues/8071#issuecomment-1148069991

_Originally posted by @lukaw3d in https://github.com/oasisprotocol/oasis-wallet-web/pull/846#discussion_r889410360_

PR #904, opened 2022-07-04
Fix extension production build errors and inject missing env variables

buberdds:
Targets #853 
So far I found three workarounds:

1. Init background via background `page` instead of `scripts` array. Compiled html has all chunks and issue with chunks is fixed.

Other workarounds:

2. kill chunks/code split, which is controlled by entry in [package.json](https://parceljs.org/features/code-splitting/#configuration)

```
  "@parcel/bundler-default": {
    "minBundles": 10 // some magic number
  }
```

3. Do not share deps between popup and background (avoid shared chunk) for example by using dynamically imported modules. This can be achieved in different ways probably and would allow us to keep chunks that will work in FF.
Without going too deep in debugging I think background script was executed without runtime (or shared modules), or before runtime was initialised and because of bg popup was failing to init as well. Couldn't make it work with async/await.  

background.ts
```
import('./backgroundImportHelper').then(({ configureAppStore, wrapStore }) => {
   const store = configureAppStore()
   wrapStore(store)
})

export {}
```
backgroundImportHelper.ts

```
export { wrapStore } from 'webext-redux'
export { configureAppStore } from 'store/configureStore'
```

### DIFF
--- a/internals/prod-ext.js
+++ b/internals/prod-ext.js
@@ -1,0 +1,14 @@
+// @ts-check
+const execSync = require('child_process').execSync
+const { getCsp } = require('./getCsp.js')
+const { buildDatetime, buildSha } = require('./getBuildData')
+
+process.env.REACT_APP_BUILD_DATETIME = buildDatetime
+process.env.REACT_APP_BUILD_SHA = buildSha
+process.env.EXTENSION_CSP = getCsp({ isExtension: true })
+
+console.log(`Content-Security-Policy: ${process.env.EXTENSION_CSP}\n`)
+
+execSync('yarn clean:ext', { stdio: 'inherit' })
+execSync('parcel build --target ext --dist-dir build-ext', { stdio: 'inherit' })
+execSync('node ./internals/scripts/validate-ext-manifest.js', { stdio: 'inherit' })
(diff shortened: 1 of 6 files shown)


public/manifest.json
```diff
@@ -30,6 +30,9 @@
   },
   "permissions": ["storage", "notifications", "activeTab"],
   "content_security_policy": "{{{ EXTENSION_CSP }}}",
-  "background": { "scripts": ["../extension/src/background.ts"], "persistent": true },
+  "background": {
+    "page": "../extension/src/background.html",
```
lukaw3d, line 34: Neat

internals/prod-ext.js
```diff
@@ -0,0 +1,17 @@
...
+process.env.EXTENSION_CSP = getCsp({ isExtension: true })
+
+console.log(`Content-Security-Policy: ${process.env.EXTENSION_CSP}\n`)
+
+execSync(
+  'yarn clean:ext && parcel build --target ext --dist-dir build-ext && node ./internals/scripts/validate-ext-manifest.js',
+  {
+    stdio: 'inherit',
+  },
+)
```
lukaw3d, line 17: maybe
```suggestion
execSync('yarn clean:ext', { stdio: 'inherit' })
execSync('parcel build --target ext --dist-dir build-ext', { stdio: 'inherit' })
execSync('node ./internals/scripts/validate-ext-manifest.js', { stdio: 'inherit' })
```

lukaw3d (approved):

PR #914, opened 2022-07-07
Test our type-safe replacements

lukaw3d:
Related to https://github.com/oasisprotocol/oasis-wallet-web/pull/911

### DIFF
--- a/src/utils/@reduxjs/toolkit.test.tsx
+++ b/src/utils/@reduxjs/toolkit.test.tsx
@@ -1,0 +1,38 @@
+// eslint-disable-next-line no-restricted-imports
+import { createSlice as createSliceOriginal } from '@reduxjs/toolkit'
+import { createSlice } from 'utils/@reduxjs/toolkit'
+
+describe('type-only test', () => {
+  describe('utils/@reduxjs/toolkit is strictly typed', () => {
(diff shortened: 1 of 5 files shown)


src/utils/@reduxjs/toolkit.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+      expect(
+        createSlice({
+          name: 'account',
+          initialState: {},
+          reducers: {},
+        }),
+      ).toBeDefined()
+    })
+
+    it('detect mistakes', () => {
```
buberdds, line 17: I don't mind merging this, but what's a real value of these tests ? For example this test is testing  a case which is impossible to implement, because TS will throw an error. How any of these test can fail ?

buberdds (approved):

src/utils/@reduxjs/toolkit.test.tsx
lukaw3d: This ensures that typescript will throw an error. E.g. if we change our type definition or bump typescript, our constraints could become less strict without us noticing otherwise

buberdds: make sense now, thanks!

PR #929, opened 2022-07-13
Fix ValidatorMediaInfo test snapshot

lukaw3d:


src/app/pages/StakingPage/Features/ValidatorMediaInfo/__tests__/index.test.tsx
```diff
@@ -5,8 +5,17 @@ import { ValidatorMediaInfo } from '..'
...
+      <ValidatorMediaInfo
+        mediaInfo={{
+          logotype: 'https://s3.amazonaws.com/s.jpg',
+          website_link: 'https://s.f',
+          twitter_acc: 'https://twitter.com/s',
+          email_address: 's@s.f',
+        }}
+      />,
+    )
+    expect(component.baseElement).toMatchSnapshot()
```
buberdds, line 18: I don't think we should use baseElement everywhere. It's useful when you deal with react portal, but in all other cases there is no need to snapshots of entire dom including body etc.

buberdds (approved):

src/app/pages/StakingPage/Features/ValidatorMediaInfo/__tests__/index.test.tsx
lukaw3d: I think baseElement a good default that can prevent mistakes.
- The difference when there are no portals:
  ```
  <body>
    <div>
    </div>
  </body>
  ```
- The difference when portals are added: saves debugging time

Issue #79, opened 2021-05-24
Display of shares


tjanez:
For example, when a user wants to reclaim (a part of) his delegation, he is also shown the corresponding amount of shares:
![Reclaim delegation](https://user-images.githubusercontent.com/2518544/119335920-4214c480-bc8d-11eb-8989-f59c32558e38.png)

The amount that is presented in actually displayed in gigashares. The actual amount of shares in this case is 9997790262966873 and shares themselves are indivisible.

I agree that displaying the raw amount of shares would be even less clear to the user, so it is better to display it in gigashares. Just make sure to to suffix the amount with the unit. We would prefer using "gigashares".

FWIW, this is a design we have in mind for the next version of the `oasis-node stake account info` CLI command:

```
Balance:
  Total: 38,011,517.084506708 ROSE
  Available: 5,105,018.191459385 ROSE
  Active Delegations from this Account:
    Total: 32,906,498.893047323 ROSE
    Delegations:
      - Address: oasis1qrw82ag2sypeytse9x9k4uxym53l5lc5jyfs2sxv (self)
        Amount:  176,207.030432434 ROSE (169,146.962340923 gigashares, 0.56%)
      - Address: oasis1qrdx0n7lgheek24t24vejdks9uqmfldtmgdv7jzz
        Amount:  32,730,291.862614889 ROSE (31,418,889.047306104 gigashares, 76.34%)
  Debonding Delegations from this Account:
    Total: 0.0 ROSE

Nonce: 5
 
Active Delegations to this Account:
  Total: 31,407,044.571716172 ROSE (30,148,661.90284496 gigashares)
  Delegations:
    - Address: oasis1qrw82ag2sypeytse9x9k4uxym53l5lc5jyfs2sxv (self)
      Amount:  176,207.030432434 ROSE (169,146.962340923 gigashares, 0.56%)
    - Address: oasis1qq2xx4pgk0wa73363l287wy48lmaf3v8tymthyhv
      Amount:  20,814,851.680934195 ROSE (19,980,865.262676913 gigashares, 66.27%)
    - Address: oasis1qrn9c0k92vk5leh62m942nqeq2kl9tmkaqk6lzz4
      Amount:  10,415,485.860349543 ROSE (9,998,169.711273221 gigashares, 33.16%)
    - Address: oasis1qqzt9qamqdl2kau4avryz6pp2nq8rgxmzysz2492
      Amount:  500.0 ROSE (479.966553905 gigashares, <0.01%)
```

[...]

PR #931, opened 2022-07-13
Add "Reclaim all" button

lukaw3d:
Fixes part of https://github.com/oasisprotocol/oasis-wallet-web/issues/492

reclaim all

### DIFF
--- a/src/app/components/ReclaimEscrowForm/index.tsx
+++ b/src/app/components/ReclaimEscrowForm/index.tsx
@@ -49,8 +49,19 @@ export const ReclaimEscrowForm = memo((props: Props) => {
       transactionActions.reclaimEscrow({
         type: 'reclaimEscrow',
         amount: Number(amount),
-        validator: props.address,
         shares: shares,
+        validator: props.address,
+      }),
+    )
+  }
+
+  const reclaimAll = () => {
+    dispatch(
+      transactionActions.reclaimEscrow({
+        type: 'reclaimEscrow',
+        amount: Number(props.maxAmount) / 10 ** 9,
+        shares: Number(props.shares) / 10 ** 9,
+        validator: props.address,
       }),
     )
   }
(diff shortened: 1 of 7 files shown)


pro-wh:
what ranges of numbers does this work on before losing precision? up to ~9M giga shares?

src/app/components/ReclaimEscrowForm/index.tsx
```diff
@@ -49,8 +49,19 @@ export const ReclaimEscrowForm = memo((props: Props) => {
...
+      }),
+    )
+  }
+
+  const reclaimAll = () => {
+    dispatch(
+      transactionActions.reclaimEscrow({
+        type: 'reclaimEscrow',
+        amount: Number(props.maxAmount) / 10 ** 9,
+        shares: Number(props.shares) / 10 ** 9,
```
lukaw3d, line 63: Further refactored in https://github.com/oasisprotocol/oasis-wallet-web/pull/932 to keep precision

lukaw3d:
@pro-wh between `8000000000000001 / 10**9 * 10**9` and `9000000000000001 / 10**9 * 10**9`

pro-wh (approved):
good for us small stakers imo

^^ the above numbers corresponding to 8-9 million gigashares, which is what you'd get for delegating ~8 million ROSE to a new validator

src/app/components/ReclaimEscrowForm/index.tsx
matevz: FYI shares should be treated as integer too, see #79.

matevz:
Can you also please make "Corresponding number of shares" term translatable in this PR.

src/app/components/ReclaimEscrowForm/index.tsx
pro-wh: too many things are using gigashares lol

tjanez (approved):

PR #932, opened 2022-07-13
Refactor dispatching transactions so "Reclaim all" doesn't lose more precision

lukaw3d:


src/app/components/ReclaimEscrowForm/index.tsx
```diff
@@ -59,8 +60,8 @@ export const ReclaimEscrowForm = memo((props: Props) => {
     dispatch(
       transactionActions.reclaimEscrow({
         type: 'reclaimEscrow',
-        amount: Number(props.maxAmount) / 10 ** 9,
-        shares: Number(props.shares) / 10 ** 9,
+        amount: props.maxAmount,
+        shares: props.shares,
```
lukaw3d, line 64: No rounding :tada:

pro-wh:
- could you add a test for 9,999,999,999,999,999,999 shares
- the APIs these come from, they all round, don't they?

lukaw3d:
Oh, we lose precision when parsing from API..

lukaw3d:
Will revisit in another year.

pro-wh (approved):
yea

PR #937, opened 2022-07-19
Change UI to be more similar to current extension: Account Page subnav 

buberdds:
PR updates styles of Transaction List subnav. Once merge I will have to sync translations files via Transifex. In next iteration this menu will be moved from Account to Stake page.

Screenshot 2022-07-19 at 18 33 50
Screenshot 2022-07-19 at 18 33 43

Screenshot 2022-07-19 at 18 33 17
Screenshot 2022-07-19 at 18 31 55

PR #939, opened 2022-07-23
Fix mobile z-index values

buberdds:
Follows #937 
- bring back z-index value 
- set width (when counter was equal to 1, counter bubble did not look good)

### DIFF
--- a/src/app/pages/AccountPage/index.tsx
+++ b/src/app/pages/AccountPage/index.tsx
@@ -33,6 +33,7 @@ import {
 } from '../../state/wallet/selectors'
 import { ActiveDelegationList } from '../StakingPage/Features/DelegationList/ActiveDelegationList'
 import { DebondingDelegationList } from '../StakingPage/Features/DelegationList/DebondingDelegationList'
+import { mobileHeaderZIndex } from '../../components/Sidebar'
 import { ValidatorList } from '../StakingPage/Features/ValidatorList'
 import { AccountDetails } from './Features/AccountDetails'
 import { AccountSummary } from './Features/AccountSummary'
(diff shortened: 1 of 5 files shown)


src/app/pages/AccountPage/index.tsx
```diff
@@ -57,16 +58,19 @@ interface NavItemProps {
...
+const counterZIndex = mobileHeaderZIndex - 1
 const NavItem = ({ counter, label, route }: NavItemProps) => {
   return (
     <StyledNavItem exact to={route}>
       <Text>{label}</Text>
       {!!counter && (
         <Box
-          style={{ position: 'absolute', top: '-3px', right: '-5px' }}
+          align="center"
+          style={{ position: 'absolute', top: '-3px', right: '-5px', zIndex: counterZIndex }}
```
lukaw3d, line 69: Commented on https://github.com/oasisprotocol/oasis-wallet-web/pull/937#discussion_r929082642
I can't reproduce an issue that would need z-index

src/app/pages/AccountPage/index.tsx
```diff
@@ -57,16 +58,19 @@ interface NavItemProps {
...
           responsive={false}
           background="brand"
           pad={{ horizontal: 'xsmall' }}
+          width="20px"
```
lukaw3d, line 73: This looks nicer :)

lukaw3d (approved):

src/app/pages/AccountPage/index.tsx
buberdds: I will keep current version as I want to keep an overlap and in this case there is no benefit of adding extra div/Stack component where I still need a customised `StyledNavItem`

PR #952, opened 2022-08-03
Improve numeric precision with BigNumber.js

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/436

Related https://github.com/oasisprotocol/oasis-wallet-web/pull/932#issuecomment-1183771519


src/app/components/ReclaimEscrowForm/index.tsx
```diff
@@ -64,6 +69,14 @@ export const ReclaimEscrowForm = memo((props: Props) => {
     )
   }
 
+  /**
+   * `<input max="9000000000.111111111">` warns about <=9000000000.11111 so we
+   * round up to `<input max="9000000000.11112">`
+   */
+  const roundedUpStringifiedFloatMaxAmount = new BigNumber(props.maxAmount)
+    .shiftedBy(-9) // / 10 ** 9
+    .toPrecision(15, BigNumber.ROUND_UP)
```
lukaw3d, line 78: Chrome seems to parse `<input max="..">` into a double-float, so we need to be round up when it hits the limit of double-float precision. Is there a good way to describe this operation? And a better way to implement it?

This implementation relies on double-float having at least 15 significant decimal digits.

src/app/components/ReclaimEscrowForm/index.tsx
```diff
@@ -76,8 +89,8 @@ export const ReclaimEscrowForm = memo((props: Props) => {
               placeholder={t('common.amount')}
               type="number"
               step="any"
-              min={0.0001}
-              max={Number(props.maxAmount) / 10 ** 9}
+              min="0"
```
lukaw3d, line 92: Is there some minimum amount for ReclaimEscrow?

src/app/components/ReclaimEscrowForm/index.tsx
pro-wh: I've heard other teams had to stop using `<input type="number"` because of this.

src/app/components/ReclaimEscrowForm/index.tsx
pro-wh: https://github.com/oasisprotocol/oasis-core/blob/master/go/consensus/tendermint/apps/staking/transactions.go#L333 not seeing any

buberdds (approved):

src/app/components/AmountFormatter/index.tsx
```diff
@@ -8,32 +8,37 @@ import * as React from 'react'
 import { memo } from 'react'
 import { useSelector } from 'react-redux'
 import { Text } from 'grommet'
+import { StringifiedBigInt } from 'types/StringifiedBigInt'
+import { formatBaseUnitsAsRose } from 'app/lib/helpers'
 
 export interface AmountFormatterProps {
-  amount: string | number | null
+  amount: string | StringifiedBigInt | number | null
```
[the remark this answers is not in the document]
lukaw3d: ooh, maybe all the bad numbers are gone. Let me check

Issue #951, opened 2022-08-03
Fix amount formatting in transaction preview


lukaw3d:
Regression since https://github.com/oasisprotocol/oasis-wallet-web/pull/935
| Before | Now |
| --- | --- |
| ![localhost_3000_account_oasis1qrtyn2q78jv6plrmexrsrv4dh89wv4n49udtg2km_active-delegations (1)](https://user-images.githubusercontent.com/3758846/182528172-09425d50-1bd0-484f-a1b2-f9b9f806a905.png)  |  ![localhost_3000_account_oasis1qrtyn2q78jv6plrmexrsrv4dh89wv4n49udtg2km_active-delegations](https://user-images.githubusercontent.com/3758846/182528164-0d727763-a28f-42fb-92ba-4e49c0d52255.png) |

PR #953, opened 2022-08-03
Fix transaction preview formatting

buberdds:
https://github.com/oasisprotocol/oasis-wallet-web/issues/951

### DIFF
--- a/src/app/components/ResponsiveGridRow/index.tsx
+++ b/src/app/components/ResponsiveGridRow/index.tsx
@@ -1,23 +1,18 @@
-/**
- *
- * ResponsiveGridRow
- *
- */
 import { Box, Text } from 'grommet'
 import React, { memo } from 'react'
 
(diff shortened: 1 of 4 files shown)


src/app/components/ResponsiveGridRow/index.tsx
```diff
@@ -4,20 +4,22 @@
...
 
-interface Props {
+interface ResponsiveGridRowProps {
+  direction?: DirectionType
   label: string | React.ReactNode
   value: string | React.ReactNode
 }
 
-export const ResponsiveGridRow = memo((props: Props) => {
+export const ResponsiveGridRow = memo(({ direction = 'column', label, value }: ResponsiveGridRowProps) => {
```
lukaw3d, line 16: Do we even need column direction anywhere?

lukaw3d (approved):

PR #958, opened 2022-08-05
Improve descriptions of transactions

lukaw3d:
Partially addresses https://github.com/oasisprotocol/oasis-wallet-web/issues/752

| Before | After |
| --- | --- |
| localhost_3000_account_oasis1qz78ap0456g2rk7j6rmtvasc9v2kjhz2s58qgj90 (1)  | localhost_3000_account_oasis1qz78ap0456g2rk7j6rmtvasc9v2kjhz2s58qgj90  |

### DIFF
--- a/src/app/components/Transaction/__tests__/__snapshots__/index.test.tsx.snap
+++ b/src/app/components/Transaction/__tests__/__snapshots__/index.test.tsx.snap
@@ -470,12 +470,12 @@ exports[`<Transaction  /> should handle unknown transaction types gracefully 1`]
     class="c1"
   >
     <svg
-      aria-label="New"
+      aria-label="Alert"
       class="c2"
       viewBox="0 0 24 24"
     >
       <path
-        d="M12 1v22M2 6l20 12m0-12L2 18"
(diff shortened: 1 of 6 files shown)


buberdds:
- imo we don't have to duplicate amount in title. So it can look like this:
"Type(title) addr, date, amount" and not "Type+amount, addr, date, amount". It's easier to "scan" page when title is more compact
- Title Method root hash.executor might be still not clear for users
- known transaction types are distinguish with unique icon 


https://github.com/oasisprotocol/oasis-wallet-web/issues/752

lukaw3d:
- Removing amounts out of title: I agree for transfers, especially of less pretty amounts. But I like the amount in additional context when setting allowance - maybe the amount shouldn't even be displayed on the right
- I don't know what we'll do with ExecutorCommits. Potentially we could fetch their details and display events that apply to this account. Some of those events will already be displayed as transactions, though some might not be. Anyway I just removed the scary "Unrecognized transaction" prefix
- Maybe ATM icon https://icons.grommet.io/?s=atm for setting allowance :shrug:

buberdds (approved):

Issue #420, opened 2021-11-09
Show more debonding information


matevz:
There is very little information about the debonding period when you want to reclaim your stake. I suggest we provide the user with basic debonding information:
- [ ] model: fetch current epoch, find a way to compute expected epoch duration (genesis file?)
- [ ] in the `reclaimEscrow` preview dialog, there should be a note saying `Current debonding period is x epochs (~y days z hours). You will not be able to access the reclaimed balance until the debonding period is over. Reclaimed balance is not awarded during the debonding period.`
- [ ] in Debonding delegations table add a localized `(~dd-mm-yy)` to values in `Debonding epoch` columns
- [ ] in Debonding delegations table add `Remaining epochs` column showing `x epochs (~y days z hours)`

tjanez:
> model: fetch current epoch, find a way to compute expected epoch duration (genesis file?)

For the expected epoch duration, I think one would need to query Oasis Monitor for the average block length and compute that by multiplying with the expected number of blocks in an epoch.

tjanez:
> Current debonding period is x epochs (~y days z hours). You will not be able to access the reclaimed balance until the debonding period is over. Reclaimed balance is not awarded during the debonding period.

Is there a gRPC query to get the current value of the `staking.params.debonding_interval` parameter?

pro-wh:
yeah, use StakingConsensusParameters https://github.com/oasisprotocol/oasis-sdk/blob/7d7d592b533cc4c43cde39d9c8cc5440958491d8/client-sdk/ts-web/core/src/client.ts#L990

PR #963, opened 2022-08-09
Display estimated debonding time in debonding list

lukaw3d:
Related to https://github.com/oasisprotocol/oasis-wallet-web/issues/420

| Before | After |
| --- | --- |
|   |   |

### DIFF
--- a/src/app/components/TimeToEpoch/index.tsx
+++ b/src/app/components/TimeToEpoch/index.tsx
@@ -1,0 +1,24 @@
+import { selectEpoch } from 'app/state/network/selectors'
+import { useSelector } from 'react-redux'
+import * as React from 'react'
+
+interface Props {
+  epoch: number
+}
+
+const estimatedEpochsPerHour = 1
+
+const relativeFormat = new Intl.RelativeTimeFormat(process?.env?.NODE_ENV === 'test' ? 'en-US' : undefined)
+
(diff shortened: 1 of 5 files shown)


src/app/components/TimeToEpoch/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { useSelector } from 'react-redux'
+import * as React from 'react'
+
+interface Props {
+  epoch: number
+}
+
+const estimatedEpochsPerHour = 1
+
+const relativeFormat = new Intl.RelativeTimeFormat(process?.env?.NODE_ENV === 'test' ? 'en-US' : undefined)
```
buberdds, line 11: The downside is that we loose multi lang support. It will show EN labels no matter what lang user is using.

buberdds (approved):

src/app/components/TimeToEpoch/index.tsx
lukaw3d: in production `new Intl.RelativeTimeFormat(undefined).format(1, 'day')` it should be localized:
image

buberdds: Either way it's not in sync with app lang. Mixing langs is a bit weird, but in this case it is still better than epoch number.

Issue #791, opened 2022-04-05
Support ParaTimes deposit/withdraw (oasisscan only)


lukaw3d:
The initial versions should be hidden behind a flag or dev-only, until all warnings are reimplemented.

lukaw3d:
Some issues:
- no support in oasismonitor
- no support for ledger accounts
- needs to model more oasisscan endpoints in swagger
- needs to show list of paratime transactions
- needs import of ethereum-compatible private keys

lukaw3d:
For reference: querying min gas price https://github.com/oasisprotocol/oasis-wallet-ext/pull/230/files#diff-e9bce3b0c0aee8d5bc193829767cd3b862c8c7431c55435133c9c80c5e1b2ff1R236

lukaw3d:
Related: warnings in extension
- https://github.com/oasisprotocol/oasis-wallet-ext/pull/236
- https://github.com/oasisprotocol/oasis-wallet-ext/pull/245

lukaw3d:
Related PRs so far:
#856
#859
#962
#969
#978
#988
#992
#1114
#1326
#1327
#1346
#1347
#1352
#1353
#1365
#1386

lukaw3d:
Snippet for gas estimation:
https://github.com/oasisprotocol/oasis-wallet-ext/pull/320#discussion_r1358838661

PR #969, opened 2022-08-12
ParaTimes: deposit and withdraw

buberdds:
PR adds
- deposit/withdraw 
- missing UI pieces

Next PRs will cover:
#978 unit tests, lock view for a 'foreign' account

Collapsible advance section
186187500-368d4cee-540a-458b-a269-699e3ffba879

Extended confirmation view
Screenshot 2022-08-25 at 18 32 58

Failed Transaction view 
Screenshot 2022-08-25 at 18 41 55


src/app/state/transaction/saga.ts
```diff
@@ -76,6 +82,40 @@ function* prepareTransfer(signer: Signer, amount: bigint, to: string) {
...
+function* prepareParatimeTransfer(
+  nic: client.NodeInternal,
+  signer: Signer,
+  amount: string,
+  to: string,
+  from: string,
+  runtime: Runtime,
+) {
+  yield* call(assertWalletIsOpen)
+  yield* call(assertSufficientBalance, BigInt(parseRoseStringToBigNumber(amount).toString()))
```
lukaw3d, line 105: toString doesn't always work
```ts
new BigNumber('10000000000000000000').shiftedBy(9).toString()
BigInt('1e+28')
  Cannot convert 1e+28 to a BigInt
```

My workaround was:
https://github.com/oasisprotocol/oasis-wallet-web/blob/b96eb8f88d717d8c3dc28882669709fdee6e917d/src/app/lib/helpers.ts#L59-L62

src/app/state/transaction/saga.ts
```diff
@@ -181,6 +221,44 @@ export function* doTransaction(action: PayloadAction<TransactionPayload>) {
...
+}
+
+export function* submitParaTimeTransaction(runtime: Runtime, amount: string, recipient: string) {
+  const accountAddress = yield* select(selectAccountAddress)
+  const nic = yield* call(getOasisNic)
+  const chainContext = yield* select(selectChainContext)
+  const signer = yield* getSigner()
+  const allowanceDifference = yield* call(getAllowanceDifference, amount, runtime.address)
+
+  if (allowanceDifference.gte(0)) {
```
lukaw3d, line 237: Prefer using base unit bigints where possible, and BigNumbers mainly for converting from inputs and API

src/app/state/transaction/saga.ts
```diff
@@ -181,6 +221,44 @@ export function* doTransaction(action: PayloadAction<TransactionPayload>) {
   }
 }
 
+function* getAllowanceDifference(amount: string, runtimeAddress: string) {
+  const allowances = yield* select(selectAccountAllowances)
+  const allowance = allowances.find((item: Allowance) => item.address === runtimeAddress)?.amount || 0
```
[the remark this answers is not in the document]
buberdds: We need to default to 0 when allowance does not exist for a runtime.

src/app/lib/transaction.ts
```diff
@@ -79,6 +82,73 @@ export class OasisTransaction {
...
+    tw.setFeeGas(gas)
+
+    return tw
+  }
+
+  public static async buildParatimeTransfer(
+    nic: OasisClient,
+    signer: Signer,
+    targetAddress: string,
+    fromAddress: string,
```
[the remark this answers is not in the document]
buberdds: `await publicKeyToAddress(signer.public())`  ?

src/app/state/account/types.ts
```diff
@@ -9,8 +9,14 @@ export interface BalanceDetails {
   total: StringifiedBigInt | null
 }
 
+export interface Allowance {
+  address: string
+  amount: string
```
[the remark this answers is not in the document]
buberdds: bigint or `StringifiedBigInt` ?

src/app/lib/transaction.ts
```diff
@@ -79,6 +82,73 @@ export class OasisTransaction {
     return tw
   }
 
+  public static async buildStakingAllowTransfer(
```
pro-wh, line 86: if not Allow, it should rather be AllowWithdraw

pro-wh: oh I see the rest of the codebase uses a broader meaning of "transfer." carry on then :upside_down_face:

src/app/lib/transaction.ts
```diff
@@ -79,6 +82,73 @@ export class OasisTransaction {
...
+
+    txWrapper
+      .setBody({
+        amount: [
+          oasis.quantity.fromBigInt(
+            BigInt(new BigNumber(amount.toString()).shiftedBy(runtimeDecimals).toFixed()),
+          ),
+          oasisRT.token.NATIVE_DENOMINATION,
+        ],
+        to: await oasis.staking.addressFromBech32(
```
pro-wh, line 143: addressFromBech32 is synchronous. check me on that though

src/app/lib/transaction.ts
```diff
@@ -79,6 +82,73 @@ export class OasisTransaction {
...
+
+  public static async buildParatimeTransfer(
+    nic: OasisClient,
+    signer: Signer,
+    targetAddress: string,
+    fromAddress: string,
+    amount: string,
+    runtimeId: string,
+    runtimeDecimals: number,
+  ): Promise<TW<oasisRT.types.ConsensusDeposit>> {
```
pro-wh, line 117: paratime transactions have a different wrapper class

src/app/lib/transaction.ts
```diff
@@ -94,6 +164,10 @@ export class OasisTransaction {
     return tw.sign(new oasis.signature.BlindContextSigner(signer), chainContext)
   }
 
+  public static async signParaTime<T>(chainContext: string, signer: Signer, tw: TW<T>): Promise<void> {
```
pro-wh, line 170: here too, need to use the paratime wrapper type

src/app/state/transaction/saga.ts
```diff
@@ -257,6 +257,13 @@ export function* submitParaTimeTransaction(runtime: Runtime, amount: string, rec
 
   yield* call(OasisTransaction.signParaTime, chainContext, signer as Signer, tw)
   yield* call(OasisTransaction.submit, nic, tw)
+  yield* put(
+    transactionActions.transactionSent({
```
pro-wh, line 257: what does this part do? it looks like it creates a transaction record of a "transfer." but this is after doing a paratime deposit. and the commit message is "refresh account balance." and the balance won't update immediately

src/config.ts
```diff
@@ -79,12 +80,15 @@ export enum RuntimeTypes {
 
 const emeraldConfig: ParaTimeConfig = {
   mainnet: {
+    address: 'oasis1qzvlg0grjxwgjj58tx2xvmv26era6t2csqn22pte',
```
pro-wh, line 83: what was the discussion on hard coding these? gonna scroll around in this PR to see if there's any

src/app/lib/transaction.ts
```diff
@@ -109,20 +110,22 @@ export class OasisTransaction {
...
+    const txWrapper = new oasisRT.consensusAccounts.Wrapper(consensusRuntimeId)[
+      type === TransactionTypes.Deposit ? 'callDeposit' : 'callWithdraw'
+    ]()
     const accountsWrapper = new oasisRT.accounts.Wrapper(consensusRuntimeId)
     const nonce = await accountsWrapper
       .queryNonce()
       .setArgs({ address: await oasis.staking.addressFromBech32(fromAddress) })
       .query(nic)
-    const feeAmount = 0n
+    const feeAmount = 1500000n
```
pro-wh, line 128: we need to use zero for deposits to support users doing their first deposit

src/config.ts
buberdds: There was a comment that these values can be hardcoded.
https://github.com/oasisprotocol/oasis-wallet-web/pull/969#discussion_r944747406

related commit: https://github.com/oasisprotocol/oasis-wallet-web/pull/969/commits/12034bdad2067f81e689896efa740d9319788678

src/app/state/transaction/saga.ts
buberdds: Added  a dedicated action for paraTime transfer to trigger refresh in account and wallet sagas.

Yes, balance won't update immediately that's why there is a delay effect. It's not ideal, but it's acceptable. We don't support WebSocket or polling to handle such case better.

buberdds:
I will push more UI changes here

src/config.ts
pro-wh: merely that we can. it's not like they'll change, after all. and it saves the extra awaiting and surely there's some CPU time saved. and it's definitely helpful for that to be grep-able

maybe add a comment with the code for doing the derivation so that we have it handy when adding new paratimes

src/app/state/account/types.ts
```diff
@@ -9,8 +9,14 @@ export interface BalanceDetails {
...
+}
+
 export interface Account extends BalanceDetails {
   address: string
+  allowances?: Allowance[]
```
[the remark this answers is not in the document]
buberdds: We do not support monitor and ledger https://github.com/oasisprotocol/oasis-wallet-web/issues/791#issuecomment-1090487222

src/config.ts
lukaw3d: I like that it's searchable; and I don't know how the derivation would best be described and named. But hardcoding does add a possibility of mistakes :/ Add a test?
```ts
describe('paraTimesConfig', () => {
  it('addresses should be derived from runtimeId', async () => {
    for (const paraTimeConfig of Object.values(paraTimesConfig)) {
      for (const paraTimeNetwork of [paraTimeConfig.mainnet, paraTimeConfig.testnet, paraTimeConfig.local]) {
        if (paraTimeNetwork.runtimeId) {
          const address = await staking.addressFromRuntimeID(misc.fromHex(paraTimeNetwork.runtimeId))
          expect(staking.addressToBech32(address).toLowerCase()).toEqual(paraTimeNetwork.address)
        }
      }
    }
  })
})
```

src/app/lib/transaction.ts
```diff
@@ -79,6 +91,82 @@ export class OasisTransaction {
...
+      amount_change: oasis.quantity.fromBigInt(amount),
+    })
+
+    const gas = await tw.estimateGas(nic, signer.public())
+    tw.setFeeGas(gas)
+
+    return tw
+  }
+
+  public static async buildParatimeTransfer(
```
lukaw3d, line 118: (I like this paratime capitalization `buildParatimeTransfer` better than `submitParaTimeTransaction`)

src/app/lib/transaction.ts
```diff
@@ -79,6 +82,73 @@ export class OasisTransaction {
...
+    const signerInfo = {
+      address_spec: { signature: { ed25519: signer.public() } },
+      nonce,
+    }
+
+    txWrapper
+      .setBody({
+        amount: [
+          oasis.quantity.fromBigInt(
+            BigInt(new BigNumber(amount.toString()).shiftedBy(runtimeDecimals).toFixed()),
```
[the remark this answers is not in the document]
lukaw3d: This looks unresolved

src/app/lib/transaction.ts
```diff
@@ -79,6 +82,73 @@ export class OasisTransaction {
...
+
+    return tw
+  }
+
+  public static async buildParatimeTransfer(
+    nic: OasisClient,
+    signer: Signer,
+    targetAddress: string,
+    fromAddress: string,
+    amount: string,
```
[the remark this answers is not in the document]
lukaw3d: This was updated to 
```ts
  public static async buildParatimeTransfer(
    nic: OasisClient,
    signer: Signer,
    transaction: ParaTimeTransaction, // for amount other methods use bigint
    fromAddress: string,
    runtimeId: string,
    runtimeDecimals: number,
  )
```

I think it should be more detailed:
```ts
  /**
   * Note: uses raw user input amounts in ROSE as parameter instead of 
   * normalized bigint base units, because it needs to convert to base units 
   * dependent on paratime decimals.
   */
  public static async buildParatimeTransfer(
    nic: OasisClient,
    signer: Signer,
    transaction: ParaTimeTransaction,
    fromAddress: string,
    runtimeId: string,
    runtimeDecimals: number,
  )
```

src/app/lib/transaction.ts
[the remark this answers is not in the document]
lukaw3d: Probably

If that's correct, it can probably be simplified `await oasis.staking.addressFromBech32(await publicKeyToAddress(signer.public()))` to `await shortPublicKey(signer.public())`

src/app/lib/transaction.ts
```diff
@@ -79,6 +91,82 @@ export class OasisTransaction {
...
+    const nonce = await accountsWrapper
+      .queryNonce()
+      .setArgs({ address: await oasis.staking.addressFromBech32(fromAddress) })
+      .query(nic)
+    const feeAmountDecimal = new BigNumber(10).pow(consensusDecimals)
+    const feeAmount = BigInt(
+      new BigNumber(transaction.feeAmount ? transaction.feeAmount : isDepositing ? '0' : '1500000')
+        .shiftedBy(runtimeDecimals)
+        .dividedBy(feeAmountDecimal)
+        .toFixed(),
```
lukaw3d, line 142: See https://github.com/oasisprotocol/oasis-wallet-web/pull/969/files#r944765518
should be toFixed(0)

src/app/lib/transaction.ts
```diff
@@ -79,6 +91,82 @@ export class OasisTransaction {
...
+      isDepositing ? 'callDeposit' : 'callWithdraw'
+    ]()
+    const accountsWrapper = new oasisRT.accounts.Wrapper(consensusRuntimeId)
+    const nonce = await accountsWrapper
+      .queryNonce()
+      .setArgs({ address: await oasis.staking.addressFromBech32(fromAddress) })
+      .query(nic)
+    const feeAmountDecimal = new BigNumber(10).pow(consensusDecimals)
+    const feeAmount = BigInt(
+      new BigNumber(transaction.feeAmount ? transaction.feeAmount : isDepositing ? '0' : '1500000')
```
lukaw3d, line 139: I'd put default fee into its own variable, and keep the comment from https://github.com/oasisprotocol/oasis-wallet-ext/blob/38d82b5/src/popup/pages/Send/index.js#L187-L188

lukaw3d: Also, shouldn't that default "1500000" be "0.0015"?

src/app/lib/transaction.ts
```diff
@@ -79,6 +91,82 @@ export class OasisTransaction {
...
+        .shiftedBy(runtimeDecimals)
+        .dividedBy(feeAmountDecimal)
```
lukaw3d, line 141: `.shiftedBy(-consensusDecimals)`

src/app/pages/ParaTimesPage/index.tsx
```diff
@@ -22,12 +25,15 @@ const getActiveFormStepComponent = (step: number) => {
       return <TransactionConfirmation />
     case TransactionFormSteps.TransactionSummary:
       return <TransactionSummary />
+    case TransactionFormSteps.TransactionError:
+      return <TransactionError />
     default:
       return <ParaTimeTransferType />
```
lukaw3d, line 31: This could use ExhaustedTypeError

src/app/state/account/types.ts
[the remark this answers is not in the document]
lukaw3d: StringifiedBigInt

src/app/state/account/types.ts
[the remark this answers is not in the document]
lukaw3d: Ah, it isn't used in oasismonitor.

Still:
- selectAccountAllowances shouldn't fall back to `[ ]` and lose the information that it's missing. It can do that in validateAllowance, since always setting new allowance is acceptable behaviour
- allowances is part of normal account information and not clearly reserved for paratimes, so it would be less of a pitfall if we can get it from oasismonitor or gRPC
- in which case it never needs to be undefined, and we don't need ` || [ ]`

src/app/state/transaction/saga.ts
```diff
@@ -181,6 +225,62 @@ export function* doTransaction(action: PayloadAction<TransactionPayload>) {
   }
 }
 
+export function* getAllowanceDifference(amount: string, runtimeAddress: string) {
+  const allowances = yield* select(selectAccountAllowances)
+  const allowance = allowances.find((item: Allowance) => item.address === runtimeAddress)?.amount || 0
+  return new BigNumber(amount).minus(allowance)
+}
+
+function* validateAllowance(
```
lukaw3d, line 234: This looks like it is setting sufficient allowance. Rename?

src/app/lib/transaction.ts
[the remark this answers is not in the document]
lukaw3d: Does that work correctly for emerald withdrawals?

src/app/state/transaction/saga.ts
```diff
@@ -76,6 +83,43 @@ function* prepareTransfer(signer: Signer, amount: bigint, to: string) {
...
+  signer: Signer,
+  transaction: ParaTimeTransaction,
+  from: string,
+  runtime: Runtime,
+) {
+  yield* call(assertWalletIsOpen)
+  if (transaction.type === TransactionTypes.Deposit) {
+    yield* call(
+      assertSufficientBalance,
+      BigInt(parseRoseStringToBigNumber(transaction.amount).toFixed(0).toString()),
```
lukaw3d, line 108:
```suggestion
      BigInt(parseRoseStringToBaseUnitString(transaction.amount)),
```

src/app/state/transaction/saga.ts
```diff
@@ -181,6 +225,62 @@ export function* doTransaction(action: PayloadAction<TransactionPayload>) {
...
+    yield* call(OasisTransaction.submit, nic, tw)
+  }
+}
+
+const transactionSentDelay = 1000 // to increase a chance to get updated account data from BE
+
+export function* submitParaTimeTransaction(runtime: Runtime, transaction: ParaTimeTransaction) {
+  const fromAddress = transaction.privateKey
+    ? yield* call(getEvmBech32Address, privateToEthAddress(transaction.privateKey))
+    : yield* select(selectAccountAddress)
```
lukaw3d, line 260: Seeing how privateKey is used, it should be renamed to ethPrivateKey

src/app/pages/ParaTimesPage/TransactionConfirmation/index.tsx
```diff
@@ -32,39 +83,71 @@ export const TransactionConfirmation = () => {
...
+            checked={transactionForm.confirmTransferToValidator}
+            description={t(
+              'paraTimes.confirmation.confirmTransferToValidatorDescription',
+              'This is a validator wallet address. Transfers to this address do not stake your funds with the validator.',
+            )}
+            label={t(
+              'paraTimes.confirmation.confirmTransferToValidatorLabel',
+              'I confirm I want to transfer tokens to a validator address',
+            )}
+            name="confirmTransferToValidator"
```
lukaw3d, line 105: Can this be type-checked to be a key of transactionForm?

src/app/pages/ParaTimesPage/TransactionAmount/FeesSection.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const [open, setOpen] = useState(false)
+
+  return (
+    <Box pad={{ vertical: 'medium' }}>
+      <Box align="start">
+        <Button
+          margin={{ bottom: 'medium' }}
+          secondary
+          size="small"
+          onClick={() => setOpen(!open)}
```
lukaw3d: What happens if user opens advanced, chooses fees, and closes advanced

src/app/state/account/selectors.ts
```diff
@@ -11,3 +11,4 @@ export const selectTransactionsError = createSelector([selectAccount], account =
 export const selectAccountAddress = createSelector([selectAccount], account => account.address)
 export const selectAccountAvailableBalance = createSelector([selectAccount], account => account.available)
 export const selectAccountIsLoading = createSelector([selectAccount], account => account.loading)
+export const selectAccountAllowances = createSelector([selectAccount], account => account.allowances || [])
```
lukaw3d, line 14: Looks like paratimes are using wrong account information

Reproduce steps
- create a wallet and open paratimes: http://localhost:3000/account/oasis1qztn43zuxtjyk49q6ee3fmy3cjyn6lk0fscnrsnh/paratimes
- click home and create a new wallet: http://localhost:3000/account/oasis1qq0xldx5xu2a05gdhn7l3h9w4u44ucx53qw7cex4
- click back until http://localhost:3000/account/oasis1qztn43zuxtjyk49q6ee3fmy3cjyn6lk0fscnrsnh/paratimes
- now the opened account is oasis1qq0xldx5xu2a05gdhn7l3h9w4u44ucx53qw7cex4 ("wallet" state)
- but viewed account is oasis1qztn43zuxtjyk49q6ee3fmy3cjyn6lk0fscnrsnh ("account" state)

(or open multiple accounts, open paratimes on one, select another account, and click back)

Issues:
- paratimes flow uses balance and other info from viewed account
- paratimes form should be hidden on foreign accounts, see
  https://github.com/oasisprotocol/oasis-wallet-web/pull/865 and https://github.com/oasisprotocol/oasis-wallet-web/pull/918

src/app/pages/ParaTimesPage/TransactionAmount/FeesSection.tsx
```diff
@@ -0,0 +1,60 @@
...
+          secondary
+          size="small"
+        />
+      </Box>
+      <Collapsible open={open}>
+        <FormField
+          name="feeAmount"
+          validate={{
+            regexp: /^(\s*|[0-9][0-9]*)$/,
+            message: t('paraTimes.validation.invalidFee', 'Value must be integer greater than or equal to 0'),
```
lukaw3d, line 31: Pretty sure feeAmount is in ROSE units, and input should allow decimal places.
Also add `step="any" min="0"` to number input

feeGas seems to use (paratime's?) base units. I would vote for consistency with feeAmount.

@pro-wh did extension have fee inputs in base units for good reason?
And setFeeAmount uses paratime's denomination, but what does setFeeGas use?

src/app/pages/ParaTimesPage/TransactionAmount/FeesSection.tsx
```diff
@@ -0,0 +1,60 @@
...
+          }}
+        >
+          <TextInput
+            name="feeAmount"
+            placeholder={t('paraTimes.amount.feeAmountPlaceholder', 'Fee Amount (nano {{ticker}})', {
```
lukaw3d, line 36: @pro-wh do we have any explanation of "Fee Amount" vs "Fee Gas" anywhere? Do we name them more-understandably anywhere?

src/app/pages/ParaTimesPage/TransactionAmount/FeesSection.tsx
buberdds: It's working the same way as in the current ext. Collapse/Expand is just a visual effect with no effect on form input data. 

1) keep current 
2) disable collapsing when any input has a value and start with an open state set to true when nav between steps if there's value
3) when fee section is closed but has a value show some labels instead

src/app/lib/transaction.ts
buberdds: For now I will use `ParaTime` here as well to unify pattern.

src/app/lib/transaction.ts
buberdds: Comment added. It is nano Rose (at least for now)

src/app/lib/transaction.ts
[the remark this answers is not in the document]
buberdds: It doesn't

src/app/state/account/types.ts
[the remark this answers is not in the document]
buberdds: check moved to validateAllowance

src/app/pages/ParaTimesPage/TransactionAmount/FeesSection.tsx
pro-wh: dunno, but it seems nice not to have to type 0.00000000 all the time

lukaw3d: oh i was wrong, feeAmount really is in nano ROSE, and feeGas is in "gas units" (according to docs https://docs.oasis.dev/oasis-core/consensus/transactions/#gas)

Can add `step="1" min="0"` to number input

src/app/pages/ParaTimesPage/TransactionAmount/FeesSection.tsx
lukaw3d: Alright, found this in docs:
- https://docs.oasis.dev/oasis-core/consensus/transactions#gas
- https://docs.oasis.dev/general/manage-tokens/advanced/oasis-cli-tools/setup#common-transaction-flags

and they don't mean what thought they did :no_mouth:

src/app/state/account/types.ts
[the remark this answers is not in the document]
lukaw3d: Test can be simply changing oasisscan.test.ts
```ts
    expect(
      // https://api.oasisscan.com/mainnet/chain/account/info/oasis1qp53ud2pcmm73mlf4qywnrr245222mvlz5a2e5ty
      parseAccount({
        rank: 0,
        address: 'oasis1qp53ud2pcmm73mlf4qywnrr245222mvlz5a2e5ty',
        available: '7791.547645364',
        escrow: '312240.929087243',
        debonding: '0',
        total: '320032.476732607',
        nonce: 89,
        allowances: [
          {
            address: 'oasis1qrnu9yhwzap7rqh6tdcdcpz0zf86hwhycchkhvt8',
            amount: '1',
          },
        ],
      }),
    ).toMatchSnapshot()
```
and oasismonitor.test.ts
```ts
    expect(
      // https://monitor.oasis.dev/data/accounts/oasis1qp53ud2pcmm73mlf4qywnrr245222mvlz5a2e5ty
      parseAccount({
        address: 'oasis1qp53ud2pcmm73mlf4qywnrr245222mvlz5a2e5ty',
        liquid_balance: 7791547645377,
        /* eslint-disable @typescript-eslint/no-loss-of-precision */
        escrow_balance: 55597275304680782,
        escrow_debonding_balance: 2613007343417,
        delegations_balance: 312240929087242,
        debonding_delegations_balance: 0,
        self_delegation_balance: 312240929087242,
        total_balance: 55605066852326159,
        created_at: '2021-08-15T22:20:44Z',
        last_active: '2022-08-31T22:35:29Z',
        nonce: 89,
        type: 'validator',
        entity_address: 'oasis1qp53ud2pcmm73mlf4qywnrr245222mvlz5a2e5ty',
        validator: {
          rate_change_interval: 1,
          rate_bound_lead: 14,
          max_rate_steps: 21,
          max_bound_steps: 21,
          status: 'inactive',
          node_address: 'oasis1qr4200x0hpza2cgfydlrckr6hwllwg7ssyh96mdu',
          consensus_address: 'EC98C32FAAB7718052E09187D16FD06048EDC8D7',
          depositors_count: 14,
          blocks_count: 18831,
          signatures_count: 2072250,
        },
      }),
    ).toMatchSnapshot()
```

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
```diff
@@ -101,14 +102,26 @@ export const TransactionAmount = () => {
...
+                  (amount: string) =>
+                    balanceValidator(amount, balance!)
+                      ? {
+                          message: t('errors.insufficientBalance', 'Insufficient balance'),
+                          status: 'error',
+                        }
+                      : undefined,
+                ]}
               >
                 <TextInput disabled={disabled} name="amount" placeholder="0" value={transactionForm.amount} />
```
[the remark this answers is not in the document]
buberdds: yes, but I will add `inputMode="decimal"` to improve mobile exp

src/app/state/account/selectors.ts
buberdds: as long as I understand the problem correctly https://github.com/oasisprotocol/oasis-wallet-web/pull/978 covers this with a redirect. Basically, user won't be even able to open  "foreign" /paratime route, because there is no value in such action.

src/app/lib/transaction.ts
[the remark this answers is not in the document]
buberdds: I don't see failing tests, so what you mean by `it is now incorrectly split across functions` ? Other than that user cannot submit `'0.1111111111111111111111111111111111111111111111111111'`

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
```diff
@@ -46,11 +48,10 @@ export const TransactionAmount = () => {
     dispatch(
       isEvmcParaTime
         ? paraTimesActions.fetchBalanceUsingEthPrivateKey({
-            privateKey: transactionForm.privateKey,
+            ethPrivateKey: transactionForm.ethPrivateKey,
```
lukaw3d, line 51: I think it would be better to only dispatch the address (slightly lower chance of private key leaking)

src/app/pages/ParaTimesPage/TransactionAmount/FeesSection.tsx
lukaw3d: I vote disable

Another bug flow:
- withdraw from emerald: fill private key, address, amount, click advanced, fill fees
- click previous
- withdraw from cipher: advanced fees are prefilled, but not opened

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
```diff
@@ -91,6 +92,7 @@ export const TransactionAmount = () => {
       <Box margin={{ bottom: 'medium' }}>
         <Form
           messages={{ required: t('paraTimes.validation.required', 'Field is required') }}
+          noValidate
```
lukaw3d, line 95: how come?

src/app/lib/transaction.ts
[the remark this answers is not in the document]
lukaw3d: - Throw on too many decimals:
  Ah, the validation now prevents too many decimals. I don't know, I'd still use `toFixed(0)` :shrug: 

- Incorrectly split across functions:
  `parseRoseStringToBigNumber` doesn't lose any precision. `toFixed(0)` in `parseRoseStringToBaseUnitString` loses precision, so the warning check is in the wrong function.
  But I think parseRoseStringToBigNumber should be the place where higher-than-decimals precision is lost by calling `baseUnitBN.decimalPlaces(0)`, and then warning should stay there.

- Other:
  `amount.toString()`: amount is already typed as a string.
  
  `new BigNumber(amount).shiftedBy(runtimeDecimals)` could be `parseRoseStringToBigNumber(amount, runtimeDecimals)` to have the extra checks

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
buberdds: I don't want to show a default html validation triggered by min and step HTML attrs

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
buberdds: In this case I will dispatch an action without payload if we want to use a selector in saga to get transactionForm data

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
```diff
@@ -124,7 +138,9 @@ export const TransactionAmount = () => {
                   }}
                   plain
                   label="MAX"
```
lukaw3d, line 140: Translations

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
```diff
@@ -124,7 +138,9 @@ export const TransactionAmount = () => {
...
-                  onClick={() => setTransactionForm({ ...transactionForm, amount: formatter(balance) })}
+                  onClick={() =>
```
lukaw3d, line 141: MAX doesn't work with fees.

E.g. withdrawing MAX from cipher (after removing setAllowance) fails with `Unknown gRPC error: withdraw: insufficient runtime balance`

src/app/pages/ParaTimesPage/index.tsx
```diff
@@ -37,6 +43,10 @@ export const ParaTimes = () => {
     }
   }, [resetTransactionForm])
 
+  useEffect(() => {
+    resetTransactionForm()
+  }, [resetTransactionForm, selectedNetwork])
```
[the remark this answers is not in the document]
lukaw3d: Filed https://github.com/oasisprotocol/oasis-wallet-web/issues/987

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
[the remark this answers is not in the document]
lukaw3d: Because it's just a text input it allows user to enter "1,1" and get confused by `Maximum of 9 decimal places is allowed` error

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
```diff
@@ -101,16 +97,34 @@ export const TransactionAmount = () => {
...
-                      }
-                    : undefined
-                }
+                validate={[
+                  (amount: string) =>
+                    !new RegExp(`^\\d*(?:[.][0-9]{0,${decimals}})?$`).test(amount)
+                      ? {
+                          message: t(
+                            'paraTimes.validation.invalidDecimalValue',
+                            'Maximum of {{token}} decimal places is allowed',
```
lukaw3d, line 106: Minor inaccuracy that we probably don't care about, and don't want an uglier regex for:
0.400000000000000000000000000 shows "Maximum of 9 decimal places is allowed"

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
buberdds: I will address this one in a new PR.

lukaw3d (approved):

src/app/pages/ParaTimesPage/TransactionRecipient/__tests__/__snapshots__/index.test.tsx.snap
```diff
@@ -52,21 +35,16 @@ exports[`<TransactionRecipient /> should render EVMc withdraw variant component
...
-      Emerald
-    </strong>
-     (EVMc) ParaTime and then click "Next"
-  </span>
-</div>
+  Please enter the address of the withdrawing wallet on the 
+  <strong>
+    Emerald
+  </strong>
+   (EVMc) ParaTime and then click "Next"
```
lukaw3d, line 46: Text is misleading

- When withdrawing from Emerald user should enter **withdrawing private key** and **receiving address**
image

- When withdrawing from Cipher user enters **receiving address**
image

Issue #792, opened 2022-04-05
Persist state


lukaw3d:
Add an option to keep imported accounts in password-encrypted localStorage

Currently user loses all state on reload.

Make sure stored information is also migrate-able from existing extension https://github.com/oasisprotocol/oasis-wallet-ext/blob/e6998cff910449fd0f33fc998d9c5e306d59aa99/src/background/service/APIService.js#L177-L189

lukaw3d:
Proof of concept https://github.com/oasisprotocol/wallet-poc/pull/1

lukaw3d:
Main issues with proof of concept - storing and syncing whole redux state:
- it stores fatal error
- syncing multiple tabs conflicts with cleanup, e.g. clearAccount
    open account in two tabs http://localhost:3000/account/oasis1qzwl8nn8lt8cv4gxh85skv9rw3h0raudws66q4y5
    navigate to home on one tab (dispatches clearAccount)
    other tab loses data and shows loading

lukaw3d:
Instead of persisting and syncing everything, we should be more selective.

Current state, by longterm-ness
- theme, network, wallet, account, dapp (soon)
- create tx, create wallet, import ledger, stake
- fatalError

We definitely need to store: theme, network (partially), wallet, selected account, dapp (soon)
and we definitely shouldn't store and sync: fatalError.

More considerations:
- Remove fatalError from redux?
- To avoid conflicts in "create tx": only allow one popup? or sync exactly?
- Remove account ID from URL to have simpler state? But then having two tabs is useless, unless we don't sync selected account ID. Maybe support old URLs still, but only show readonly state?
- If user changes password: different selections on tabs might get lost. Don't care?.

### So

![nomnoml](https://user-images.githubusercontent.com/3758846/188751112-c2c1f7cc-676a-486a-a813-670456949bbc.svg)

[...]

PR #975, opened 2022-08-18
Persist state and sync between tabs

lukaw3d:
Part of https://github.com/oasisprotocol/oasis-wallet-web/issues/792

~Temporarily deployed to https://lukaw3d.github.io/oasis-wallet-web/ce527a5/#/~
Updated deploy https://lukaw3d.github.io/oasis-wallet-web/32fbe2b/#/

<details><summary>Old preview of <a href="https://github.com/oasisprotocol/oasis-wallet-web/compare/22369e06ec03a96cc88cd835f6a03ce98375e91e">22369e0</a></summary>

profile

</details>

### Preview of [ce527a5](https://github.com/oasisprotocol/oasis-wallet-web/compare/ce527a521a1e5bdd9c2eb9ceb6c778b2aa9d27be)

preview-ce527a5

TODO
- [x] don't store fatal errors
- [x] syncing multiple tabs and cleanup, e.g. clearAccount
    open account in two tabs http://localhost:3000/account/oasis1qzwl8nn8lt8cv4gxh85skv9rw3h0raudws66q4y5
    navigate to home on one tab (dispatches clearAccount)
    other tab loses data and shows loading
- [x] skip login:
  - [x] refactor to make networkSaga maintainable:
      always show unlock modal and
      if profile: on unlock emit setUnlockedRootState
      if profile: on skip emit event like setUnlockedRootState to use in networkSaga
      if no profile: auto click skip login emit event like setUnlockedRootState to use in networkSaga
- [x] walletId global var
- [x] close wallet = lock
- [x] change password
- [x] choose password and create profile when importing
- [x] test ui
- [x] test persistence

Deferred and moved to issue:
- auto lock timeout
- multiple profiles
- migrate old storage @metamask/browser-passworder
- test syncing tabs


pro-wh (commented):
reading about how this works

src/app/state/persist/encryption.ts
```diff
@@ -0,0 +1,67 @@
...
+  const encryptedObj: EncryptedObject = {
+    secretbox: nacl.secretbox(dataBytes, nonce, keyWithSalt.key),
+    nonce: nonce,
+    salt: keyWithSalt.salt,
+  }
+  const encryptedString = base64andStringify(encryptedObj)
+  return encryptedString
+}
+
+export async function decryptWithPassword<T>(password: string, encryptedString: EncryptedString): Promise<T> {
```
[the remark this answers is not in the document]
lukaw3d: I don't understand :D 

A subset of the comment: true, we could avoid `deriveKeyFromPassword` (with a new random salt) in `decryptState` (on unlock) by returning `derivedKeyWithSalt` (with old salt) from `decryptWithPassword`. Seems safe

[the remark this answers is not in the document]
pro-wh: ya I guess, if you prefer to couple the derivation and decryption into one call

lukaw3d:
I temporarily deployed it to https://lukaw3d.github.io/oasis-wallet-web/ce527a5/#/

nikolaglumac:
Great work @lukaw3d 🎉 
I will do a proper review later today or tomorrow - one thing that I want to mention now (and perhaps that is something we can address in a separate PR) is that we should think about password rules. Having users set a single-character password is definitely not what we should allow.

nikolaglumac:
Hi @lukaw3d - the flow works pretty well! I think it is a good first version - we can later iterate on it and improve it over time (e.g. by adding "remove-account" options etc).

There is this minor thing I found:
close-wallet

It should be an easy fix.

Thanks!

lukaw3d:
Minimal edge-cases to write tests for:
- lock while viewing an account causes error with clearTransaction (fixed in https://github.com/oasisprotocol/oasis-wallet-web/pull/975/commits/966b15cfbd6da72192098c21b8cb132d5abd34cd)
- quickly unlock + lock: caused error with network fetching action (fixed in https://github.com/oasisprotocol/oasis-wallet-web/pull/975/commits/966b15cfbd6da72192098c21b8cb132d5abd34cd)
- create wallet, enable "Store private keys", type password, disable "Store private keys", next. Expect not to be stored.
- create profile, erase profile, open a new account, causes an error (fixed in https://github.com/oasisprotocol/oasis-wallet-web/pull/975/commits/787037b34b146485eb2760971c8b23c74d2bf79d)

Write test harness for interactions in multiple tabs (probably in playwright?) (defer to separate PR)

lukaw3d:
> @nikolaglumac
> There is this minor thing I found: Once you create a profile, regardless of if you opened it later or not, "Close wallet" button is always visible

That was intentional so users can get back to login screen after clicking "Continue without the profile" (without it: user would need to create a wallet to get that same button)

buberdds:
Loading state issue (web build only)
- open an account 
- open a new tab with the same url
- close wallet
- in the previous tab loading state "hangs"

lukaw3d:
Fixing 1. is annoying
- I could refactor `resetRootState` to keep some slices, instead of resetting everything but: it doesn't feel correct to keep state after user clicks Close Wallet; and it probably introduces state combinations that previously weren't reachable and handled.
- Another option: I could redirect home on `resetRootState`, but only in webapp `takeLatest(resetRootState, () => { if (runtimeIs === 'webapp' && selectUnlockedStatus === 'emptyUnpersisted') location.href = '/' })`. Luckily extension with saga in background page won't support multiple tabs / popups yet. And that still needs `navigate('/'); dispatch(persistActions.lockAsync())` on click.

csillag:
I see that it's possible to use two different accounts in two different tabs, but I can't use Mainnet on one tab and Testnet on the other. Is that an intentional choice?

src/config.ts
```diff
@@ -65,3 +65,5 @@ export const config: BackendConfig = {
     },
   },
 }
+
+export const runtimeIs = window.chrome?.runtime?.id ? 'extension' : 'webapp'
```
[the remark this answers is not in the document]
lukaw3d: I only checked in Firefox, but I see webextension-polyfill is using the same mechanism 
https://github.com/mozilla/webextension-polyfill/blob/6e3e26c488fe6096e49e05dd7bd1e834552be869/src/browser-polyfill.js#L9-L10

src/app/components/Persist/UnlockForm.tsx
```diff
@@ -0,0 +1,82 @@
...
+              type="submit"
+              label={t('persist.loginToProfile.unlock', 'Unlock')}
+              primary
+              onClick={() => dispatch(persistActions.unlockAsync({ password: password }))}
+            />
+
+            <Button
+              label={t('persist.loginToProfile.skipUnlocking', 'Continue without the profile')}
+              onClick={() => {
+                navigate('/')
```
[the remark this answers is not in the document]
lukaw3d: Good UX question. I've been considering the skipped profile like incognito mode, not read-only profile.
- Against:
  - We've removed search for foreign accounts, so read-only view is barely supported.
  - I was considering removing remove account ID from URL. To have fewer sources of state. That would also increase privacy - after locking wallet, someone couldn't check browser history (or click skip) to get your address.
- For:
  - Would allow wallet owner to quickly get **their account's address**. Pre-redirect URL might not be locked account (open a foreign account URL from bookmark and click skip). More complex proposal is to have a wallet pin to unlock read-only, and only ask for password before signing transactions.
  - Would allow quickly viewing **foreign account's balance** from URL/bookmark. But this is just worse search.

src/app/state/persist/syncTabs.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+/**
+ * When interacting with a second tab it syncs these actions.
+ */
+export const whitelistTabSyncActions = [
+  themeActions.changeTheme.type,
+  walletActions.walletOpened.type,
+  walletActions.updateBalance.type,
+  networkActions.networkSelected.type,
```
lukaw3d, line 45: > @csillag: I see that it's possible to use two different accounts in two different tabs, but I can't use Mainnet on one tab and Testnet on the other. Is that an intentional choice?

Yeah. I'm syncing `networkSelected` here. It would take quite some refactoring to keep it separate. I'm syncing wallet slice and it includes balances, so after switching networks it showed testnet balance on mainnet tab.

Related note in https://github.com/oasisprotocol/oasis-wallet-web/issues/792#issuecomment-1238715243
> selected network needs to be synced as long as network-specific balance is in state.wallets[*].balance

playwright/tests/syncTabs.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  test.describe('switch network and open second tab', () => {
+    test('unpersisted', async ({ page, context }) => {
+      await page.goto('/open-wallet/private-key')
+      await page.getByPlaceholder('Enter your private key here').fill(privateKey)
+      await page.keyboard.press('Enter')
+      await expect(page).toHaveURL(new RegExp(`/account/${privateKeyAddress}`))
+      const tab2 = await context.newPage()
+      await testSyncingNetwork(page, tab2)
+    })
```
lukaw3d, line 95: This is failing :/ and it's a mess

- first tab initializes
- attempts to sync state from other tabs
- networkSaga initializes, dispatches 'skipUnlocking', and fetches mainnet network info
- **user switches to testnet**, networkSaga fetches testnet network info
- second tab initializes
- second tab: attempts to sync state from other tabs
- second tab: networkSaga initializes, dispatches 'skipUnlocking', and fetches mainnet network info
- first tab: sends sync state, including testnet selection
- first tab: gets synced skipUnlocking, and fetches mainnet network info
- second tab: gets synced state, including testnet selection
- first tab: dispatches selectNetwork with fetched mainnet network info
- second tab: gets synced selectNetwork with fetched mainnet network info
- both tabs are **showing mainnet** instead of testnet

lukaw3d: https://github.com/oasisprotocol/oasis-wallet-web/pull/975/commits/c125984a415bd1a3fd9dd61c9ce6f2634e8ef8eb added a timeout to wait for tabs to sync, and only skipUnlocking if tabs don't sync

lukaw3d:
Fixed everything, maybe. Deployed to https://lukaw3d.github.io/oasis-wallet-web/32fbe2b/#/

buberdds:
Routing bug still occurs and it's not related to foreign accounts. Repro steps: import two accounts using mnemonic, try to open the second opened account in the second tab / fail. However, I can open the second account using Account Switcher in the second tab, but selectedWallet value will be now incorrect in the first tab, right?

lukaw3d:
@buberdds I'm not able to reproduce this
Can you make more specific steps (when to open another tab, what URL, choosing persistence?)

I tried:
- tab1: http://localhost:3000/create-wallet, open acc1 + acc2 without persistence
- tab2: http://localhost:3000/, switch to acc2 without persistence, send transaction from acc2
- tab1: send transaction from acc1
.
- tab1: http://localhost:3000/create-wallet, open acc1 + acc2 with persistence
- tab2: http://localhost:3000/, switch to acc2, send transaction from acc2
- tab1: send transaction from acc1
.
- tab1: http://localhost:3000/create-wallet, open acc1 without persistence
- tab2: http://localhost:3000/create-wallet, open acc2 without persistence, send transaction from acc2
- tab1: send transaction from acc1
.
- tab1: http://localhost:3000/create-wallet
- tab2: http://localhost:3000/
- tab1: open acc1 without persistence
- tab2: home, create wallet, open acc2 without persistence, send transaction from acc2
- tab1: send transaction from acc1

buberdds:
- tab1 open acc1 + acc2 via mnemonic without persistence - acc1 is open
- open tab2 - enter acc 2 url - FAIL 
As a workaround in tab2 I can open acc2 using Account Switcher, but if wallet slice is synced is value of `selectedWallet` in tab1 correct?

lukaw3d:
Ah yeah, that is currently treated as a foreign account.

Workaround should work: wallet slice (including `selectedWallet`) is synced from tab1 to tab2 when it opens. But after initializing, `selectWallet` action isn't synced: https://github.com/oasisprotocol/oasis-wallet-web/blob/32fbe2b08ffa05a8893e03fd05f2527ba7967c03/src/app/state/persist/syncTabs.ts#L41-L49

So tab1 will continue to have selectedWallet=acc1 after tab2 switched accounts

buberdds (approved):
I don't have any other comments and I am fine with merging with "redirect" issue just to move forward, but would wait for @pro-wh approval related to sec.

pro-wh:
oh uh

src/app/state/persist/saga.ts
```diff
@@ -0,0 +1,165 @@
...
+  if (action.type.startsWith('@')) return // Ignore @@INIT from redux-devtools-instrument
+  if (action.type.startsWith('&')) return // Ignore e.g. &_GET_INIT_STATE from redux-state-sync
+  if (isActionSynced(action)) return // Ignore actions synced across tabs from redux-state-sync
+
+  // Handle actions while state is locked
+  // TODO: If we used private/public key-pair we could encrypt and persist
+  // state while state is locked. But not if we support multiple profiles.
+  if (!latestState.persist.stringifiedEncryptionKey) {
+    if (walletActions.walletOpened.match(action)) {
+      throw new Error('Could not add account while state is locked.')
```
pro-wh, line 152: security review: what's this for? coverage indicates we don't test this. do we catch this somewhere?

src/app/state/persist/saga.ts
```diff
@@ -0,0 +1,165 @@
...
+import { walletActions } from 'app/state/wallet'
+import { selectUnlockedStatus } from 'app/state/selectUnlockedStatus'
+import { runtimeIs } from 'config'
+
+function* watchPersistAsync() {
+  yield* fork(function* () {
+    const channelQueue = yield* actionChannel<AnyAction>('*')
+    while (true) {
+      // TODO: could flush(channelQueue) to speedup multiple consecutive writes
+      const action: AnyAction = yield* take(channelQueue)
```
pro-wh, line 24: non-security: does this result in every open tab trying to write on each change?

src/app/state/persist/index.ts
```diff
@@ -0,0 +1,112 @@
...
+
+/** Syncing tabs is only needed in web app, not in extension. */
+export const needsSyncingTabs = runtimeIs === 'webapp'
+
+// Simulate with `delete window.BroadcastChannel`
+export const isSyncingTabsSupported = typeof BroadcastChannel === 'function'
+
+export function getInitialState(): PersistState {
+  return {
+    hasPersistedProfiles: !!localStorage.getItem(STORAGE_FIELD),
```
pro-wh, line 25: security review: only uses of `localStorage` in master branch are related to theme selection

non-security: tests use `window.localStorage`, app code uses `localStorage`, previous app code uses `window.localStorage && localStorage`

src/app/state/persist/saga.ts
```diff
@@ -0,0 +1,165 @@
...
+// Handlers
+function* setPasswordAsync(action: ReturnType<typeof persistActions.setPasswordAsync>) {
+  /**
+   * Latest state does not match state immediately after action was dispatched,
+   * but when it is queued by {@link watchPersistAsync}.
+   */
+  const latestState: RootState = yield* select()
+  const keyWithSalt = yield* call(deriveKeyFromPassword, action.payload.password)
+  const encryptedState = yield* call(encryptState, latestState, keyWithSalt)
+  localStorage.setItem(STORAGE_FIELD, encryptedState)
```
pro-wh, line 67: security review: callsite 1/2, stores only encrypted data :+1:

src/app/state/persist/saga.ts
```diff
@@ -0,0 +1,165 @@
...
+    if (walletActions.walletOpened.match(action)) {
+      throw new Error('Could not add account while state is locked.')
+    } else {
+      return // Ignore non-essential actions emitted while state is locked.
+    }
+  }
+
+  const keyWithSalt: KeyWithSalt = fromBase64andParse(latestState.persist.stringifiedEncryptionKey)
+  const encryptedState = yield* call(encryptState, latestState, keyWithSalt)
+  localStorage.setItem(STORAGE_FIELD, encryptedState)
```
pro-wh, line 160: security review: callsite 2/2, stores only encrypted data :+1:

src/app/state/persist/encryption.ts
```diff
@@ -0,0 +1,67 @@
...
+  const encryptedObj = fromBase64andParse<EncryptedObject>(encryptedString)
+  const derivedKeyWithSalt = await deriveKeyFromPassword(password, encryptedObj.salt)
+  const dataBytes = nacl.secretbox.open(encryptedObj.secretbox, encryptedObj.nonce, derivedKeyWithSalt.key)
+  if (!dataBytes) throw new PasswordWrongError()
```
pro-wh, line 48: security review: :information_source: https://github.com/dchest/tweetnacl-js#naclsecretboxopenbox-nonce-key

pro-wh (approved):
:shrug: no security blockers from me

src/app/state/persist/saga.ts
lukaw3d: Context: Next line skips writing on each change if wallet is locked.
For example:
- user might switch network
- it starts loading
- switch to another tab, click lock
- network info loads and dispatches networkSelected
- skip writing because we don't have the key to write
- after unlocking old network is still selected

I think it's acceptable to ignore almost all actions like this, except when silently losing an account:
- add an account from private key
- switch to another tab, click lock
- account info loads and dispatches addWallet
- throw error because we don't have the key to write

pro-wh: couldn't understand that

is the network selection somehow related with unlocking?

in the second example, why wouldn't we have the key to write? is there a way to add an account from private key before setting up a password?

lukaw3d: > is the network selection somehow related with unlocking?

If you mean in relation to "after unlocking old network is still selected": selected network is persisted after network info loads. So normally it would keep last selected network after unlocking.

> in the second example, why wouldn't we have the key to write? is there a way to add an account from private key before setting up a password?

when user locks one tab that syncs to other tabs and it removes stringifiedEncryptionKey from state. So we can't encrypt the new account to storage

src/app/components/Persist/ChoosePasswordFields.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const isChoiceDisabled =
+    isPersistenceUnsupported ||
+    unlockedStatus === 'unlockedProfile' ||
+    unlockedStatus === 'skippedUnlockingProfile'
+
+  const isChoosingPassword = !isChoiceDisabled && startPersisting
+
+  if (isPersistenceUnsupported) {
+    return <p>{t('persist.createProfile.unsupported', 'Persistence is not supported in this browser.')}</p>
```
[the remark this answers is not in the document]
lukaw3d: yep, I should also throw inside setPasswordAsync, if user somehow gets past UI

src/app/state/persist/saga.ts
pro-wh: is there a more general way to stop sagas (?) when locking?

thanks for explaining

lukaw3d: There might be, with `sagaMiddleware.run(rootSagas).cancel()`, but I haven't looked at actually implementing it and what the possible consequences could be.

None-general way would be to add `race({ task, cancel: cancel: take(persistActions.lockAsync) })` to most tasks, except the infinitely running ones. Or alternatively wait for most tasks to finish before locking (except infinitely running).

I rate it: too much effort

pro-wh: k

playwright/tests/syncTabs.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    test('incognito', async ({ page, context }) => {
+      await addPersistedStorage(page)
+      await page.goto('/')
+      await page.getByRole('button', { name: 'Continue without the profile' }).click()
+      await expect(page.getByTestId('account-selector')).toBeHidden()
+
+      const tab2 = await context.newPage()
+      await tab2.goto('/')
+      await expect(tab2.getByRole('button', { name: 'Unlock' })).toBeHidden()
```
lukaw3d, line 59: This is failing on CI, but not locally

lukaw3d: I haven't found a reason why it matches differently on CI, only fixed it with https://github.com/oasisprotocol/oasis-wallet-web/commit/77be85730f689ec2e9586c0a4c399c2b6fbe5ce7.
`await expect(tab2.getByRole('button', { name: 'Unlock' })).toBeHidden()` on CI matched
```tsx
<a aria-label="Open wallet">
  <button tabindex="-1">
    <div>
      <svg aria-label="Unlock">..</svg>
      Open wallet
</
```

iSQL:
Hello,

Is there any info when will this be published?

lukaw3d:
@iSQL I'm not sure. We'll want to deploy it after we move wallet to a new domain (so there's no confusion about saved data on current domain)

and perhaps finish more of https://github.com/oasisprotocol/oasis-wallet-web/issues/792#issuecomment-1269273318

lukaw3d:
@iSQL it is finally published

Issue #82, opened 2021-05-24
Add ability to open additional accounts for the same wallet / mnemonic


tjanez:
Besides the _primary account_ (described in [ADR 0008](https://github.com/oasisprotocol/oasis-core/blob/master/docs/adr/0008-standard-account-key-generation.md#key-derivation-paths)), also allow opening additional accounts for non-zero key numbers, i.e. `m/44'/474'/1'`, `m/44'/474'/2'`, ...

Esya:
Makes sense for it to work with non-ledger wallets too - this implies a good bit of refactoring though - tagging this as a post-release feature https://github.com/Esya/oasis-wallet/milestone/2

PR #983, opened 2022-09-06
Derive multiple accounts from mnemonic

buberdds:
Closes https://github.com/oasisprotocol/oasis-wallet-web/issues/82

Most of this stuff was extracted or moved from "ledger". FromLedgerModal is now called MultiAccountsSelectionModal, redux part was renamed as well as it's not longer bound to ledger.

Screenshot 2022-09-06 at 14 55 46

Screenshot 2022-09-06 at 14 56 35


src/types/RootState.ts
```diff
@@ -17,7 +17,7 @@ export interface RootState {
   account: AccountState
   network: NetworkState
   transaction: TransactionState
-  ledger: LedgerState
+  multiAccounts: MultiAccountsState
```
lukaw3d, line 20: I'd name it importAccounts or chooseAccountsToImport

and I'd name PR "Derive multiple accounts from mnemonic"

src/app/state/multiaccounts/index.ts
```diff
@@ -1,39 +1,49 @@
...
       state.accounts[index].selected = !state.accounts[index].selected
     },
-    accountsListed(state, action: PayloadAction<LedgerAccount[]>) {
+    accountsListed(state, action: PayloadAction<MultiAccountsListAccount[]>) {
       state.accounts = action.payload
     },
-    setStep(state, action: PayloadAction<LedgerStep>) {
+    setMnemonic(state, action: PayloadAction<string>) {
+      state.mnemonic = action.payload
+    },
```
lukaw3d, line 35: I don't like this. I think better solution would be for MultiAccountsSelector to take something like AddWalletPayload[], so there isn't a need to call `getAccountSigner(mnemonic, index)` twice (when enumerating and when opening) and risk a discrepancy in indexes

src/app/pages/OpenWalletPage/Features/MultiAccountsSelectionModal/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  return (
+    <Box gap="small">
+      {accounts.map((a, index) => (
+        <Account
+          index={index}
+          address={a.address}
+          balance={a.balance.available} // TODO: get total balance
+          type={WalletType.Ledger}
```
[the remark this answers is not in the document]
buberdds: Mnemonic label is fixed. Account Switcher is not showing derivation path and that's how it has to work based on Tadej's feedback.

src/app/state/multiaccounts/index.ts
buberdds: To render a selection modal we don't need public and private keys. We may not even need those keys as user can cancel an action/close modal. Question is do we want to generate private keys upfront or later when they actually needed.

src/app/pages/OpenWalletPage/Features/MultiAccountsSelectionModal/index.tsx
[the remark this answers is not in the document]
lukaw3d: Oh, that's odd

| Importing accounts | Switching accounts |
| --- | --- |
| image image  |  image |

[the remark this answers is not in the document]
buberdds: One of a next step is to show a custom account name instead.

lukaw3d (approved):
Great! Just split the first commit and combine some others

src/app/state/transaction/saga.ts
```diff
@@ -7,7 +7,7 @@ import { call, put, race, select, take, takeEvery } from 'typed-redux-saga'
 import { ErrorPayload, ExhaustedTypeError, WalletError, WalletErrors } from 'types/errors'
 
 import { transactionActions } from '.'
-import { sign } from '../ledger/saga'
+import { sign } from '../importaccounts/saga'
```
lukaw3d, line 10: I didn't notice before, this ledger-specific sign function is now oddly named / oddly placed

buberdds: indeed, should sign and getUSBTransport generator functions be moved to `app/lib/ledger.ts` ?

lukaw3d: yep

maybe also have fewer ledger sign functions?
- sign from ../importaccounts/saga calls
- OasisTransaction.signUsingLedger calls
- LedgerSigner.sign calls
- OasisApp.sign

PR #980, opened 2022-08-31
Fix importing multiple accounts from ledger

buberdds:
Select the first wallet from a selection after all wallets are loaded

Closes https://github.com/oasisprotocol/oasis-wallet-web/issues/934

PR #984, opened 2022-09-07
Fix importing and reselecting the same account and cleanup

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/pull/980#issuecomment-1238192565

### DIFF
--- a/src/app/state/wallet/saga.ts
+++ b/src/app/state/wallet/saga.ts
@@ -125,7 +125,8 @@ export function* openWalletFromMnemonic() {
  * If the wallet exists already, do nothingg
  * If it has "selectImmediately", we select it immediately
  */
-export function* addWallet({ payload: newWallet }: PayloadAction<AddWalletPayload>) {
+export function* addWallet({ payload }: PayloadAction<AddWalletPayload>) {
+  const { selectImmediately, ...newWallet } = payload
   const existingWallet = yield* call(getWalletByAddress, newWallet.address)
   if (!existingWallet) {
     yield* put(walletActions.walletOpened(newWallet))
@@ -133,7 +134,7 @@ export function* addWallet({ payload: newWallet }: PayloadAction<AddWalletPayloa
 
   const walletId = existingWallet ? existingWallet.id : newWallet.id
 
-  if (newWallet.selectImmediately) {
+  if (selectImmediately) {
     yield* put(walletActions.selectWallet(walletId))
   }
 }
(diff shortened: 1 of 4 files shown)


buberdds:
re-importing of a wallet does not work without `walletSelected`. if we want to get rid of it we should handle that in `selectWallet`

lukaw3d:
I'm not sure what you mean. Yes, re-importing of a wallet does not work, either way. This is not an attempt to fix it tho.

And an ugly way to fix it remains possible:
```ts
  if (selectImmediately) {
    yield* put(walletActions.selectWallet(undefined))
    yield* put(walletActions.selectWallet(walletId))
  }
```

buberdds:
Yes, please include this action if you want to remove `walletSelected`.

lukaw3d:
Added back selecting `undefined` as a quickfix.

Larger solutions:
- when importing: disable checkboxes for existing accounts
- remove useRouteRedirects and redirect in import components

src/app/state/wallet/saga.ts
```diff
@@ -129,15 +127,17 @@ export function* openWalletFromMnemonic({ payload: mnemonic }: PayloadAction<str
...
   const existingWallet = yield* call(getWalletByAddress, newWallet.address)
   if (!existingWallet) {
     yield* put(walletActions.walletOpened(newWallet))
   }
 
   const walletId = existingWallet ? existingWallet.id : newWallet.id
 
-  if (newWallet.selectImmediately) {
+  if (selectImmediately) {
+    yield* put(walletActions.selectWallet(undefined)) // Workaround so useRouteRedirects detects selecting the same account
```
buberdds, line 140: Why you have restored this line in a different place?

lukaw3d: To avoid bugs with importing multiple accounts

buberdds:
not sure it's wip or ready for review, but reselecting is still failing for me for a mnemonic and ledger, but works for a priv key.

lukaw3d:
oh, it's supposed to work :/

lukaw3d:
Alright, it's ready now.
I'll simplify into these commits after review:
```
Remove unused walletLoaded action
Deduplicate walletSelected and selectWallet
Don't keep selectImmediately in wallet state
Deduplicate selecting wallet by using selectImmediately
Test importing multiple accounts
Test reimporting and reselecting the same accounts
Fix importing and reselecting the same account
```

src/app/state/wallet/saga.ts
```diff
@@ -114,29 +107,30 @@ export function* openWalletFromMnemonic() {
...
+    yield* delay(1)
```
buberdds, line 133: ouch, can you add a comment that this is used to avoid React batch updates?

buberdds (approved):

src/app/state/wallet/saga.ts
lukaw3d: yep yep

Tho I'm not sure why this would be batched in react 17 :shrug:

PR #985, opened 2022-09-07
Use named components in translations

lukaw3d:


src/locales/en/translation.json
```diff
@@ -114,7 +114,7 @@
   "createWallet": {
     "confirmSaved": "I saved my keyphrase",
     "doNotShare": "Never share your keyphrase, anyone with your keyphrase can access your wallet and your tokens.",
-    "instruction": "Save your keyphrase <1>in the right order</1> in a secure location, you will need it to open your wallet.",
+    "instruction": "Save your keyphrase <strong>in the right order</strong> in a secure location, you will need it\n                to open your wallet.",
```
buberdds, line 117: Why you need those weird whitespaces and why only in EN?

src/app/pages/CreateWalletPage/__tests__/__snapshots__/index.test.tsx.snap
```diff
@@ -1466,13 +1466,7 @@ exports[`<CreateWalletPage  /> should match snapshot 1`] = `
       >
         <span
           class="c17"
-        >
-          Save your keyphrase
```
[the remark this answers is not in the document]
lukaw3d: Mocked i18next doesn't include default string

https://github.com/oasisprotocol/oasis-wallet-web/blob/9bc96779f9136a5f83b2eb66066a8952be0e9f2c/src/__mocks__/react-i18next.js#L42

(not sure if we should deviate from https://github.com/i18next/react-i18next/blob/master/example/test-jest/src/__mocks__/react-i18next.js just to add it back)

src/locales/en/translation.json
lukaw3d: Oh, good catch. Must have come from `yarn lint` and `yarn extract-messages`
```jsx
                defaults="Save your keyphrase <strong>in the right order</strong> in a secure location, you will need it
                to open your wallet."
```

buberdds (approved):

PR #988, opened 2022-09-09
ParaTimes: handle MAX value in transaction amount step

buberdds:
- set max amount based on transaction type and/or custom fee
- extend form validation to include a case where user uses max button and later changes fee amount

### DIFF
--- a/src/app/pages/ParaTimesPage/TransactionAmount/__tests__/index.test.tsx
+++ b/src/app/pages/ParaTimesPage/TransactionAmount/__tests__/index.test.tsx
@@ -20,6 +20,7 @@ describe('<TransactionAmount />', () => {
   const mockUseParaTimesResult = {
     balance: '1000000000',
     balanceInBaseUnit: true,
+    consensusDecimals,
     decimals: consensusDecimals,
     isDepositing: true,
     isEvmcParaTime: false,
--- a/src/app/pages/ParaTimesPage/TransactionAmount/__tests__/index.test.tsx
+++ b/src/app/pages/ParaTimesPage/TransactionAmount/__tests__/index.test.tsx
@@ -5,7 +5,7 @@ import { paraTimesActions } from 'app/state/paratimes'
 import { useParaTimes, ParaTimesHook } from '../../useParaTimes'
 import { useParaTimesNavigation, ParaTimesNavigationHook } from '../../useParaTimesNavigation'
 import { TransactionAmount } from '..'
-import { ParaTime } from 'config'
+import { consensusDecimals, ParaTime } from 'config'
 
 const mockDispatch = jest.fn()
 jest.mock('react-redux', () => ({
@@ -20,6 +20,7 @@ describe('<TransactionAmount />', () => {
   const mockUseParaTimesResult = {
     balance: '1000000000',
     balanceInBaseUnit: true,
+    decimals: consensusDecimals,
     isDepositing: true,
     isEvmcParaTime: false,
     isWalletEmpty: false,
@@ -103,6 +104,37 @@ describe('<TransactionAmount />', () => {
     )
   })
 
+  it('should set input value to max and include custom fee', async () => {
+    const setTransactionForm = jest.fn()
+    jest.mocked(useParaTimes).mockReturnValue({
+      ...mockUseParaTimesResult,
+      setTransactionForm,
+      balance: '91004888889',
+      transactionForm: {
+        ...mockUseParaTimesResult.transactionForm,
+        feeAmount: '4888889',
+      },
+    })
+    render(<TransactionAmount />)
+
+    await userEvent.click(screen.getByRole('button', { name: 'MAX' }))
+    expect(setTransactionForm).toHaveBeenNthCalledWith(1, expect.objectContaining({ amount: '91' }))
+  })
+
+  it('should set input value to max and include default fee value for withdraws', async () => {
+    const setTransactionForm = jest.fn()
+    jest.mocked(useParaTimes).mockReturnValue({
+      ...mockUseParaTimesResult,
+      setTransactionForm,
+      balance: '2439997766000',
+      isDepositing: false,
+    })
+    render(<TransactionAmount />)
+
+    await userEvent.click(screen.getByRole('button', { name: 'MAX' }))
+    expect(setTransactionForm).toHaveBeenNthCalledWith(1, expect.objectContaining({ amount: '2439.996266' }))
+  })
+
   it('should require amount field on form submit', async () => {
     const navigateToConfirmation = jest.fn()
     jest.mocked(useParaTimesNavigation).mockReturnValue({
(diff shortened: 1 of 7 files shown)


src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
```diff
@@ -11,13 +11,27 @@ import {
...
 
+function getMaxAmountValue(
+  isDepositing: boolean,
+  accountBalance: StringifiedBigInt,
+  feeAmount: string,
+): StringifiedBigInt {
+  const maxValue =
+    isDepositing && !feeAmount
+      ? accountBalance
+      : BigInt(accountBalance!) - (feeAmount ? BigInt(feeAmount) : BigInt(defaultWithdrawFeeAmount))
```
lukaw3d, line 31: This doesn't seem to match how feeAmount is calculated in buildParaTimeTransfer; and it duplicates its logic

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
```diff
@@ -127,21 +152,32 @@ export const TransactionAmount = () => {
...
-                  onClick={() =>
-                    setTransactionForm({ ...transactionForm, amount: formatter(balance).replaceAll(',', '') })
-                  }
-                />
+                >
+                  <Tip
+                    content={t('paraTimes.amount.tooltip', 'Max value may be decreased by the fee')}
+                    dropProps={{ align: { bottom: 'top' } }}
+                  >
+                    <Button
```
lukaw3d, line 168: Doesn't work:

When withdrawing MAX `0.0055` from emerald it calculates `0.0054999999985` but fails with `Unknown gRPC error: client: transaction check failed: runtime error: module: consensus code: 5 message: consensus: amount not representable`

err

lukaw3d (changes requested):

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+): StringifiedBigInt {
+  const balance = balanceInBaseUnit
+    ? accountBalance
+    : new BigNumber(accountBalance).shiftedBy(consensusDecimals - runtimeDecimals).toFixed(0)
+  const maxValue =
+    isDepositing && !feeAmount
+      ? accountBalance
+      : BigInt(balance) - BigInt(feeAmount || defaultWithdrawFeeAmount)
+  return maxValue.toString()
+}
```
lukaw3d, line 34: I don't understand any of this code.

useParaTimes starts with `balance`, `accountBalance`, and `walletBalance` (depends on isDepositing). Then it renames `walletBalance` to `balance`.

TransactionAmount takes useParaTimes's `balance` and getMaxAmountValue renames it to `accountBalance`. Then getMaxAmountValue computes a `balance` and `maxValue` (depends on isDepositing).

And `balanceInBaseUnit`, `decimals`, and `isWalletEmpty` refer to only one of those balances.. probably. I also no longer like `isWalletEmpty` name, since it sounds like there are no accounts in a wallet.

And getMaxAmountValue's `runtimeDecimals` comes from useParaTimes's `decimals` - which can be consensusDecimals or runtimeDecimals

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+  balanceInBaseUnit: boolean,
+  consensusDecimals: number,
+  runtimeDecimals: number,
+): StringifiedBigInt {
+  const balance = balanceInBaseUnit
+    ? accountBalance
+    : new BigNumber(accountBalance).shiftedBy(consensusDecimals - runtimeDecimals).toFixed(0)
+  const maxValue =
+    isDepositing && !feeAmount
+      ? accountBalance
```
lukaw3d, line 31: Add comment: is this relying on `defaultDepositFeeAmount = '0'`, or are fees paid from different account's balance?

I assume the second, but then `&& !feeAmount` doesn't make sense

src/app/pages/ParaTimesPage/TransactionAmount/__tests__/index.test.tsx
```diff
@@ -0,0 +1,255 @@
...
+    expect(screen.getByRole('button', { name: 'Next' })).toBeDisabled()
+  })
+
+  it('should set input value to max', () => {
+    const setTransactionForm = jest.fn()
+    jest.mocked(useParaTimes).mockReturnValue({ ...mockUseParaTimesResult, setTransactionForm })
+    const { rerender } = render(<TransactionAmount />)
+
+    userEvent.click(screen.getByRole('button', { name: 'MAX' }))
+    expect(setTransactionForm).toHaveBeenNthCalledWith(1, expect.objectContaining({ amount: '1' }))
```
lukaw3d, line 83: `toHaveBeenLastCalledWith` would be simpler than `toHaveBeenNthCalledWith`

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+                            amount: formatBaseUnitsAsRose(
+                              getMaxAmountValue(
+                                isDepositing,
+                                balance,
+                                transactionForm.feeAmount,
+                                balanceInBaseUnit,
+                                consensusDecimals,
+                                decimals,
+                              ),
+                            ).replaceAll(',', ''),
```
lukaw3d, line 195: Now that there's no conditional `formatter`, can this be refactored without `replaceAll(',', '')`?

src/app/pages/ParaTimesPage/useParaTimes.ts
```diff
@@ -94,7 +94,7 @@ export const useParaTimes = (): ParaTimesHook => {
     accountAddress,
     accountIsLoading,
     availableParaTimesForSelectedNetwork,
-    balance: walletBalance,
+    balance: walletBalance || '0',
```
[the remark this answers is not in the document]
lukaw3d: Related: ParaTimesState.balance type should be `| null` instead of
```ts
export interface ParaTimesState {
  balance: StringifiedBigInt

export const initialState: ParaTimesState = {
  balance: '',
```

src/app/pages/ParaTimesPage/TransactionAmount/index.tsx
lukaw3d: Did deposit and withdrawal amounts to any paratime need to have at most 9 decimal places? Even though EVM has 18?

pro-wh: deposit and withdrawal amounts in the transaction need to be expressed in 10^-18 rose, but they have to be a multiple of 10^9 of those

buberdds:
will restore ParaTime work in new PRs

PR #990, opened 2022-09-09
Display hardening in derivation path

lukaw3d:
| Before | After |
| --- | --- |
| localhost_3000_create-wallet (1)  |   localhost_3000_create-wallet |


tjanez (approved):

src/app/state/importaccounts/saga.ts
```diff
@@ -51,11 +51,12 @@ function* enumerateAccountsFromMnemonic(action: PayloadAction<string>) {
         address,
         balance,
         path: [44, 474, i],
+        pathDisplay: `m/44'/474'/${i}'`, // Hardening should match oasis.hdkey.HDKey.getAccountSigner implementation
```
buberdds, line 54: Now when we have two properties to handle paths, what is a use case for `path` property for wallets opened via mnemonic?

lukaw3d: There's currently no use for it

I think the ideal would be to remove `path` and allow validated string path as input to ledger and HDKey
- https://github.com/oasisprotocol/ledger-js/blob/9551915abda2701fb037c472810338cf845cb5fd/src/index.js#L243-L244
- https://github.com/oasisprotocol/oasis-sdk/blob/09f227c7b29a7efb927834be445a0c8d4ba8d7a8/client-sdk/ts-web/core/src/hdkey.ts#L24-L28

buberdds (approved):

src/app/state/importaccounts/saga.ts
tjanez: > I think the ideal would be to remove `path` and allow validated string path as input to ledger and HDKey

Huh... I think we should do the opposite. The underlying ts-web SDK should handle the low level details and have a method/function to return the path string (if required) but I don't think the wallet should be manipulating with derivation paths and formatting them on its own. I find this too error prone and less reusable.

lukaw3d: This PR is approximately:
  ```ts
  storedWallet = [
    {
      type: 'mnemonic',
      path: [44, 474, i], // unused, but `i` could be used to re-derive
      pathDisplay: `m/44'/474'/${i}'`,
      privateKey: oasis.hdkey.HDKey.getAccountSigner(mnemonic, i).secretKey,
      // signs tx with `privateKey`
    },
    {
      type: 'ledger',
      path: [44, 474, 0, 0, i],
      pathDisplay: `m/44'/474'/0'/0'/${i}'`,
      // signs tx with `LedgerSigner(path)`
    }
  ]
  ```

My ideal:
  ```ts
  storedWallet = [
    {
      type: 'mnemonic',
      path: `m/44'/474'/${i}'`, // displayed and could be used to re-derive

      // strictly validated string - e.g. ensures every level is hardened
      privateKey: oasis.hdkey.HDKey.getAccountSigner(mnemonic, path).secretKey,
      // signs tx with `privateKey`
    },
    {
      type: 'ledger',
      path: `m/44'/474'/0'/0'/${i}'`,
      // signs tx with `LedgerSigner(path)`
      // strictly validated string - e.g. ensures hardened and has 3 levels (ADR8) or 5 (legacy)
    }
  ]
  ```

If I understand correctly, your ideal:
  ```ts
  storedWallet = [
    {
      type: 'mnemonic',
      path: i, // unused, but could be used to re-derive
      pathDisplay: oasis.hdkey.HDKey.getAccountSigner(mnemonic, i).pathDisplay,
      privateKey: oasis.hdkey.HDKey.getAccountSigner(mnemonic, i).secretKey,
      // signs tx with `privateKey`
    },
    {
      type: 'ledger',
      path: { type: 'legacy' or 'adr8', index: i },
      pathDisplay: LedgerSigner(path.type, path.index).pathDisplay,
      // signs tx with `LedgerSigner(path.type, path.index)`
    }
  ]
  ```

I think only PR is actually error prone

PR #1001, opened 2022-09-19
Update react-router to v6

buberdds:
I did not find a way to use `react-transition-group` with router v6. There were issues with nested routes animations, app re-renders and console warnings. In order to use new router I am switching to `framer-motion`, but my preference would be not to use any animation lib for routes to keep things simple and wait for official v6 API to support routes animations.


lukaw3d:
react-router always breaking everything :facepalm:
Are nested animations problematic because it's missing something like "useNestedLocation"?

I agree with removing animations. If we do keep them though, I'd rather see
```tsx
            <Route
              element={<FadeIn><HomePage /></FadeIn>}
              path="/"
            />
```
than `<FadeIn>` on the root of every page component.

tjanez:
> I agree with removing animations

I would also rather (temporarily) remove animations. I would just file a GitHub issue to consider re-enabling/re-adding animations once react-router supports them natively.

buberdds:
Commit with `framer-motion` was removed.

lukaw3d (approved):

PR #1004, opened 2022-09-21
Handle invalid address in url

### LINKED ISSUE #940: Prevent app crash when invalid address is provided
In one of a CR comment there was an incorrect link to a wallet (missing last char)
https://wallet.oasisprotocol.org/account/oasis1qz78ap0456g2rk7j6rmtvasc9v2kjhz2s58qgj9

Instead of app crash we can treat that case as 'route not found' and redirect user to main page.

csillag:
Fixes #940.

It doesn't redirect (as suggested in the issue), just indicate the problem and avoids the crash.

### DIFF
--- a/src/app/pages/AccountPage/index.tsx
+++ b/src/app/pages/AccountPage/index.tsx
@@ -35,6 +35,7 @@ import { mobileHeaderZIndex } from '../../components/Sidebar'
 import { ValidatorList } from '../StakingPage/Features/ValidatorList'
 import { AccountDetails } from './Features/AccountDetails'
 import { AccountSummary } from './Features/AccountSummary'
+import { isValidAddress } from '../../lib/helpers'
 
 const StyledNavItem = styled(NavLink)`
   display: flex;
@@ -84,13 +85,13 @@ const NavItem = ({ counter, label, route }: NavItemProps) => {
   )
 }
 
-interface Props {}
+interface AccountPageProps {}
 
 interface AccountPageParams {
   address: string
 }
 
-export function AccountPage(props: Props) {
+function AccountPageInternal(props: AccountPageProps) {
   const { t } = useTranslation()
   const isMobile = React.useContext(ResponsiveContext) === 'small'
   const { address } = useParams<keyof AccountPageParams>()
@@ -216,4 +217,17 @@ export function AccountPage(props: Props) {
       )}
     </Box>
   )
+}
+
+export function AccountPage(props: AccountPageProps) {
+  const { t } = useTranslation()
+  const { address } = useParams<keyof AccountPageParams>()
+  if (!isValidAddress(address!)) {
+    return (
+      <Box pad={'small'}>
+        <AlertBox color={'status-error'}>{t('errors.invalidAddress', 'Invalid address')}</AlertBox>
+      </Box>
+    )
+  }
+  return <AccountPageInternal {...props} />
 }


csillag:
On thing to consider here is the up to now,  `PrettyAddress` was throwing an exception for illegal addresses, but this PR changes the behavior so that now it only indicated the error, but lets the render continue. This might change the application's behavior in other situations in presence of illegal addresses... so some testing might be needed.

csillag:
Sorry, wait, something is not right here...

src/app/components/PrettyAddress/__tests__/index.test.tsx
```diff
@@ -16,4 +16,9 @@ describe('<PrettyAddress  />', () => {
       /oasis1 qq2v zcvx n0js 5uns ch5m e2xz 4kr4 3vca sv0d 5eq4/,
     )
   })
+
+  it('should indicate invalid addresses', () => {
+    const page = render(<PrettyAddress address="oasis1qq2vzcvxn0js5unsch5me2xz4kr43vcasv0d5eq" />)
```
pro-wh, line 21: :information_source: the above with last digit removed

src/app/state/account/saga.ts
```diff
@@ -19,7 +19,20 @@ export function* fetchAccount(action: PayloadAction<string>) {
 
   yield* put(actions.setLoading(true))
   const nic = yield* call(getOasisNic)
-  const publicKey = yield* call(addressToPublicKey, address)
+  let publicKey: Uint8Array
+  try {
+    publicKey = yield* call(addressToPublicKey, address)
+  } catch (error) {
```
pro-wh, line 25: swallows the error, I'm opposed to this

pro-wh (changes requested):
swallows error from addressToPublicKey

src/app/state/account/saga.ts
csillag: It dispatches an `accountError` action, so I think it transforms the errors, not swallows it... or am I missing something here?  (I probably am.)

csillag: This is what the user will see, when the address in the URL is wrong:

image

buberdds (approved):

src/app/state/account/saga.ts
buberdds: The downside of this approach is that we will hit APIs to fetch delegations, transactions, debonding info when we already know URL is wrong. UI is in a bit weird state too. I would ask PO/PM if this layout is Ok.

src/app/components/PrettyAddress/index.tsx
```diff
@@ -28,6 +29,7 @@ export const PrettyAddress = memo((props: Props) => {
 
     return <>{prettyAddress}</>
   } else {
-    throw new Error('Invalid bech32 address')
+    console.error('Invalid beck32 address', props.address)
+    return <AlertBox color={'status-error'}>Invalid bech32 address {props.address}</AlertBox>
```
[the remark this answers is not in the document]
csillag: Yes, I will re-implement this in a way that doesn't involve PrettyAddress at all.

src/app/state/account/saga.ts
csillag: Yes, I'm going to change that.

csillag:
@buberdds @pro-wh  this has been revamped. No more changes in PrettyAddress. No more unnecessary API request to load staking data. Ready for review again.

src/app/state/account/saga.ts
```diff
@@ -17,6 +17,17 @@ import { selectAccountAddress } from './selectors'
...
   const address = action.payload
 
+  if (!isValidAddress(address)) {
+    yield* put(
+      actions.accountError({
+        code: WalletErrors.InvalidAddress,
+        message: 'Invalid address.',
+      }),
+    )
+    yield* put(actions.setLoading(false))
```
buberdds, line 27: loading was not initiated at this point so we don't need to dispatch this action

src/app/pages/AccountPage/index.tsx
```diff
@@ -135,6 +136,14 @@ export function AccountPage(props: Props) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [dispatch, walletsPublicKeys.join(','), selectedNetwork])
 
+  if (address && !isValidAddress(address)) {
```
buberdds, line 139: shorten version !isValidAddress(address!)

src/app/state/staking/saga.ts
```diff
@@ -190,6 +190,11 @@ export function* getValidatorDetails({ payload: address }: PayloadAction<string>
 }
 
 export function* fetchAccount({ payload: address }: PayloadAction<string>) {
+  if (!isValidAddress(address)) {
+    yield* put(stakingActions.updateDelegations({ delegations: [], debondingDelegations: [] }))
```
buberdds, line 194: Do we need to dispatch this action?

src/app/state/account/saga.ts
```diff
@@ -17,6 +17,17 @@ import { selectAccountAddress } from './selectors'
 export function* fetchAccount(action: PayloadAction<string>) {
   const address = action.payload
 
+  if (!isValidAddress(address)) {
+    yield* put(
+      actions.accountError({
```
buberdds, line 22: Account error is set in app via this action, but AccountPage is not selecting accountError from a store. It has its own check.

src/app/state/account/saga.ts
csillag: That's what I thought, too, but in fact the stats _is_ initialized with loading set to true, see here: https://github.com/oasisprotocol/oasis-wallet-web/blob/94ab4f055cbd43a57f241ccc9b7ca82d69a8188c/src/app/state/account/index.ts#L17

So if loading fails, we do need to set it to false somehow. (I have tested it, and the loading dialog stays without this line.)

Now it's good question why is it initialized like that. I don't know, maybe it's to avoid some flickering when initially loading the code? It has been changed from false to true by @Esya here: https://github.com/oasisprotocol/oasis-wallet-web/commit/dc226b682889bdf9798378cb9c038ede573b6c78#diff-dfe080912ab14112544da12f1bc8e10274d9392417a4edc981d0cc504cb7c98e

Anyway I didn't want to mess with the state, so I just cleared it here.

src/app/pages/AccountPage/index.tsx
csillag: OK

src/app/state/staking/saga.ts
csillag: We don't strictly need it, but I don't like to idea of keeping data belonging to some previous address.
One could argue that the data won't be displayed anyway, since the account details page is hidden, and if/when a new, correct address will be provided, the new data will be reloaded anyway, and that's all true, but still, there is no need to keep the old, invalid data in memory any longer at this point.

Since the unnecessary line is bothering you, I will get rid of it.

src/app/state/account/saga.ts
csillag: That's true, but my intention was to keep the app's internal data model (which is independent of the UI) consistent.
If we don't register an error here, the state stored in redux will look like as if we have a valid account loaded, which is definitely not the case. I wanted to avoid that, even if the UI is not visible because of the check on AccountPage.

But if you say that we don't have to care about the consistency of the internal app state  (since the UI is hidden anyway), I'm happy to remove it.

csillag:
@buberdds I have replied to all your comments, and pushed my latest changes.

src/app/state/account/saga.ts
buberdds: I have nothing against keeping an error in state. I was more curious why we are not using it in AccountPage to render AlertBox.

csillag: Ah, I see. There is actually a reason for that: the `AccountPage` is rendered _before_ the account loading saga is triggered. So if we used the presence of the error as the condition, then when initializing the app, there would be a short while when the error is not yet marked in the state, and we try to render the UI, and get an exception at `PrerttyAddress`, which then crashed the whole UI, and it doesn't recover any more.

To avoid this race condition, it's easier and safer to simply check the validity of the address in the UI, too.

kostko:
Task linked: [CU-2yrztvw Prevent app crash when invalid address is provided](https://app.clickup.com/t/2yrztvw)

pro-wh:
thanks, dismissing the changes-requested status.

src/app/pages/AccountPage/index.tsx
```diff
@@ -135,6 +136,14 @@ export function AccountPage(props: Props) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [dispatch, walletsPublicKeys.join(','), selectedNetwork])
 
+  if (!isValidAddress(address!)) {
```
lukaw3d, line 139: Can this be checked at the router level?

csillag: I'm not aware of any method for doing conditional route matching with React Router _based on the current value of dynamic segments_. (Although I'm not too familiar with React Router; I usually prefer to use my own router implementation.) As far as I can tell, it's not supported.

lukaw3d: awful :/

I think we should add a wrapper around AccountPage to validate then, so AccountPage doesn't dispatch a bunch of actions we need to discard

csillag: Yes, I was thinking about the same thing today.

csillag: >  we should add a wrapper around AccountPage to validate then, so AccountPage doesn't dispatch a bunch of actions we need to discard

OK I have added the wrapper.

src/app/state/account/saga.ts
csillag: At this point, `store.getState().account.loading` is still true, because it has been initialized that way.
The loading dialog is invisible because it's normally rendered by the account page, which is now hidden, but still, the internal state of the app is erroneously marked as loading. So I think it makes sense to set loading to false.

src/app/pages/AccountPage/index.tsx
```diff
@@ -213,3 +214,15 @@ export function AccountPage(props: Props) {
...
   )
 }
+
+export function AccountPage(props: AccountPageProps) {
+  const { t } = useTranslation()
+  const { address } = useParams<keyof AccountPageParams>()
+  return isValidAddress(address!) ? (
+    <AccountPageInternal {...props} />
+  ) : (
+    <Box pad={'small'}>
```
lukaw3d, line 224: (I have slight style preference to have general case as the last return, and special cases as early returns)
```ts
  if (!isValidAddress(address!)) {
    return <Box pad={'small'}> ..
  }

  return <AccountPageInternal {...props} />
```

src/app/state/account/saga.ts
```diff
@@ -17,6 +17,17 @@ import { selectAccountAddress } from './selectors'
 export function* fetchAccount(action: PayloadAction<string>) {
   const address = action.payload
 
+  if (!isValidAddress(address)) {
+    yield* put(
```
lukaw3d, line 21: I'd remove if we don't need

src/app/state/staking/saga.ts
```diff
@@ -190,6 +190,10 @@ export function* getValidatorDetails({ payload: address }: PayloadAction<string>
 }
 
 export function* fetchAccount({ payload: address }: PayloadAction<string>) {
+  if (!isValidAddress(address)) {
+    return
```
lukaw3d, line 194: I'd remove if we don't need

lukaw3d (approved):

src/app/pages/AccountPage/index.tsx
csillag: I agree. Changed accordingly.

src/app/state/account/saga.ts
csillag: We don't strictly need it at this point, so I removed

src/app/state/staking/saga.ts
csillag: We don't strictly need it at this point, so I removed

PR #1007, opened 2022-09-21
Detect invalid private keys when importing them

### LINKED ISSUE #960: Can we detect typos in private key?
If user imports `oPNmZybIOymrQCpCGGICczsaopANP02kwOhCyxETXljLLmRChL1QJGzJq3Pf3i+dFBN+peIK2vQ3Ew0wSQbp3g==` it opens `oasis1qpd0sz9s0s5ewmmmgyn7h44hg6yygv2reyr9mmre` and they can sign transactions.

If user imports `aaamZybIOymrQCpCGGICczsaopANP02kwOhCyxETXljLLmRChL1QJGzJq3Pf3i+dFBN+peIK2vQ3Ew0wSQbp3g==` it opens `oasis1qpd0sz9s0s5ewmmmgyn7h44hg6yygv2reyr9mmre` and they will only get an error when trying to sign a transaction: `signature verification failed`

csillag:
Fixes #960.

### DIFF
--- a/src/app/lib/key.ts
+++ b/src/app/lib/key.ts
@@ -1,5 +1,16 @@
 import nacl from 'tweetnacl'
 
+/**
+ * Verify if a key pair is actually usable for signatures
+ */
+function canSignWith(keyPair: nacl.SignKeyPair) {
+  const { secretKey, publicKey } = keyPair
+  const testMessage = Uint8Array.from([1, 2, 3])
+  const signedTestMessage = nacl.sign(testMessage, secretKey)
+  const testDecoded = nacl.sign.open(signedTestMessage, publicKey)
+  return `${testDecoded}` === `${testMessage}`
+}
+
 export class OasisKey {
   /**
    * @param key 32-bytes seed or 64-bytes secret (seed + public)
@@ -9,7 +20,11 @@ export class OasisKey {
     if (key.length === 32) {
       return nacl.sign.keyPair.fromSeed(key).secretKey
     } else if (key.length === 64) {
-      return nacl.sign.keyPair.fromSecretKey(key).secretKey
+      const keyPair = nacl.sign.keyPair.fromSecretKey(key)
+      if (!canSignWith(keyPair)) {
+        throw new Error('Invalid private key. There must a typo somewhere.')
+      }
+      return keyPair.secretKey
     } else {
       throw new Error('Invalid private key shape')
     }
--- a/cypress/integration/open-wallet.ts
+++ b/cypress/integration/open-wallet.ts
@@ -53,6 +53,12 @@ describe('Open wallet', () => {
       cy.findByText(/Invalid private key/).should('be.visible')
     })
 
+    it('Should reject invalid keys, even if base64 seems to be OK', () => {
+      cy.findByTestId('privatekey').type('aaamZybIOymrQCpCGGICczsaopANP02kwOhCyxETXljLLmRChL1QJGzJq3Pf3i+dFBN+peIK2vQ3Ew0wSQbp3g==')
+      cy.findByRole('button', { name: /Import my wallet/ }).click()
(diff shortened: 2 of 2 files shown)


src/app/lib/key.ts
```diff
@@ -9,7 +20,11 @@ export class OasisKey {
     if (key.length === 32) {
       return nacl.sign.keyPair.fromSeed(key).secretKey
     } else if (key.length === 64) {
-      return nacl.sign.keyPair.fromSecretKey(key).secretKey
+      const keyPair = nacl.sign.keyPair.fromSecretKey(key)
+      if (!canSignWith(keyPair)) {
+        throw new Error('Invalid private key. There must a typo somewhere..')
```
pro-wh, line 25: extra dot at the end?

src/app/lib/key.ts
```diff
@@ -1,5 +1,16 @@
...
 
+/**
+ * Verify if a key pair is actually usable for signatures
+ */
+function canSignWith(keyPair: nacl.SignKeyPair) {
+  const { secretKey, publicKey } = keyPair
+  const testMessage = Uint8Array.from([1, 2, 3])
+  const signedTestMessage = nacl.sign(testMessage, secretKey)
+  const testDecoded = nacl.sign.open(signedTestMessage, publicKey)
+  return testDecoded && testDecoded.toString() === testMessage.toString()
```
pro-wh, line 11: :information_source: reason for the `testDecoded &&`

> nacl.sign.open(signedMessage, publicKey)
>
> Verifies the signed message and returns the message without signature.
>
> Returns null if verification failed.

the `.toString()` gets us an easily comparable `1,2,3`

pro-wh (approved):
reasonable way to check

src/app/lib/key.ts
csillag: Removed.

pro-wh: thanks

src/app/lib/key.ts
csillag: SInce the returned value can also be `null`, I can't simply call `.toString()`, but `?.toString()` works, so I got rid of `testDecoded &&`.

pro-wh: oops I mean that bit about the toString as a separate note.

could also do `'' + testDecoded` or `String(testDecoded)`

csillag: So instead of `testDecoded?.toString() === testMessage.toString()`, would you prefer `testDecoded + '' === testMessage + ''` ?

pro-wh: really the `testDecoded &&` was fine imo

csillag: The linter actually came up with this:
```
return `${testDecoded}` === `${testMessage}`
```
This version is probably marginally more efficient (with the string building), and can also handle the null without an extra check.

pro-wh: that sounds fine too

csillag: OK, so let's go with that, and I'm marking this conversation as resolved.

lukaw3d (commented):
Nice

src/app/lib/key.ts
```diff
@@ -1,5 +1,16 @@
...
 
+/**
+ * Verify if a key pair is actually usable for signatures
+ */
+function canSignWith(keyPair: nacl.SignKeyPair) {
+  const { secretKey, publicKey } = keyPair
+  const testMessage = Uint8Array.from([1, 2, 3])
+  const signedTestMessage = nacl.sign(testMessage, secretKey)
+  const testDecoded = nacl.sign.open(signedTestMessage, publicKey)
+  return `${testDecoded}` === `${testMessage}`
```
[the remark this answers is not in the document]
csillag: OK, this has been improved in https://github.com/oasisprotocol/oasis-wallet-web/pull/1009

PR #1009, opened 2022-09-22
 Make private key testing code more explicit

csillag:
... when doing UInt8Array to string conversion.

This improves upon #1007.

### DIFF
--- a/src/app/lib/key.ts
+++ b/src/app/lib/key.ts
@@ -1,4 +1,5 @@
 import nacl from 'tweetnacl'
+import { uint2hex } from './helpers'
 
 /**
  * Verify if a key pair is actually usable for signatures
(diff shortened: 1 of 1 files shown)


lukaw3d (approved):

src/app/lib/key.ts
```diff
@@ -8,7 +9,7 @@ function canSignWith(keyPair: nacl.SignKeyPair) {
   const testMessage = Uint8Array.from([1, 2, 3])
   const signedTestMessage = nacl.sign(testMessage, secretKey)
   const testDecoded = nacl.sign.open(signedTestMessage, publicKey)
-  return `${testDecoded}` === `${testMessage}`
+  return testDecoded && uint2hex(testDecoded) === uint2hex(testMessage)
```
lukaw3d, line 12: (we probably only need to check that testDecoded isn't null, but this is fine)

kostko:
Task linked: [CU-3gb22b9 Implement early detection of invalid private key on import](https://app.clickup.com/t/3gb22b9)

PR #1018, opened 2022-09-22
Extract PasswordField from FromPrivateKey

lukaw3d:
Just making it reusable for persisted state

### DIFF
--- a/src/app/components/PasswordField/index.tsx
+++ b/src/app/components/PasswordField/index.tsx
@@ -1,33 +1,47 @@
-export function PasswordField() {
+import { Box, FormField, Button, TextInput, Tip } from 'grommet'
+import { View, Hide } from 'grommet-icons/icons'
+import * as React from 'react'
+
+interface Props {
+  placeholder: string
+  inputElementId: string
+
+  autoComplete: 'on' | 'off' | 'new-password' | 'current-password'
+  value: string
+  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void
+  error: string | false
+  showTip: string
+  hideTip: string
+
+  width: string
+}
+
+export function PasswordField(props: Props) {
+  const [passwordIsVisible, setPasswordIsVisible] = React.useState(false)
+
   return (
     <FormField
-      htmlFor="privateKey"
-      error={privateKeyIsValid === false ? t('openWallet.privateKey.error', 'Invalid private key') : ''}
+      htmlFor={props.inputElementId}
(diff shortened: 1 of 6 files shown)


src/app/components/PasswordField/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+          placeholder={props.placeholder}
+          value={props.value}
+          onChange={props.onChange}
+          type={passwordIsVisible ? 'text' : 'password'}
+          plain
+        />
+        <Tip content={passwordIsVisible ? props.hideTip : props.showTip}>
+          <Button
+            onClick={() => setPasswordIsVisible(!passwordIsVisible)}
+            icon={passwordIsVisible ? <View /> : <Hide />}
```
csillag, line 41: I think the icons should be the other way around.
 * When the text is hidden, we want to offer showing in, hence the "view" icon (the eye) should be shown
 * When the text is visible, we want to offer hiding it, hence the "hide" icon should be shown.
 
I tried to look at other implementations, and most are doing it that way.

image

Twitter:

image

image

BitStamp:

image

image

It's not fully consistent; for example FaceBook has it backwards. 

But I still think that the majority is on this side, and it makes more sense.

csillag (approved):
Generally, LGTM.

As I have explained, I would switch the icons around, but it's just a small things.

src/app/pages/OpenWalletPage/Features/FromPrivateKey/index.tsx
```diff
@@ -54,38 +53,18 @@ export function FromPrivateKey(props: Props) {
...
-              }
-            >
-              <Button
-                onClick={() => setPrivateKeyIsVisible(!privateKeyIsVisible)}
-                icon={privateKeyIsVisible ? <View /> : <Hide />}
-              />
-            </Tip>
-          </Box>
-        </FormField>
+        ></PasswordField>
```
csillag, line 66: This could be a self closing tab.

src/app/components/PasswordField/index.tsx
lukaw3d: Moved into https://github.com/oasisprotocol/oasis-wallet-web/issues/1022

src/app/pages/OpenWalletPage/Features/FromPrivateKey/index.tsx
lukaw3d: Changed. I don't really care tho

If it is good practice then we should add react/self-closing-comp

PR #1019, opened 2022-09-22
Remove global walletId and index accounts by address

lukaw3d:
Motivation: `let walletId = 0` is annoying to add to persisted state

### DIFF
--- a/src/app/state/wallet/index.ts
+++ b/src/app/state/wallet/index.ts
@@ -16,18 +16,18 @@ const slice = createSlice({
     openWalletFromMnemonic(state, action: PayloadAction<void>) {},
     openWalletFromPrivateKey(state, action: PayloadAction<string>) {},
     openWalletsFromLedger(state, action: PayloadAction<void>) {},
-    selectWallet(state, action: PayloadAction<number | undefined>) {
+    selectWallet(state, action: PayloadAction<string | undefined>) {
       state.selectedWallet = action.payload
     },
     closeWallet(state, action: PayloadAction<void>) {},
     fetchWallet(state, action: PayloadAction<Wallet>) {},
     updateBalance(state, action: PayloadAction<BalanceUpdatePayload>) {
-      Object.assign(state.wallets[action.payload.walletId].balance, action.payload.balance)
+      Object.assign(state.wallets[action.payload.address].balance, action.payload.balance)
     },
     addWallet(state, action: PayloadAction<AddWalletPayload>) {},
     walletOpened(state, action: PayloadAction<Wallet>) {
       const newWallet = action.payload
-      state.wallets[newWallet.id] = Object.assign({}, newWallet)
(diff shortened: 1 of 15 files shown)


src/app/state/wallet/types.ts
```diff
@@ -43,6 +42,6 @@ export interface AddWalletPayload extends Wallet {
 /* --- STATE --- */
 export interface WalletState {
   isOpen: boolean
-  selectedWallet?: number
-  wallets: { [id: number]: Wallet }
+  selectedWallet?: string
+  wallets: { [address: string]: Wallet }
```
csillag, line 46: Instead of using  `{ [address: string]: Wallet }` I find the `Record<string, Wallet>` form easier to read.

csillag (approved):
Reviewed. LGTM.

(Almost feels like it should have been written like this from the start.)

src/app/state/wallet/types.ts
lukaw3d: I think the key name `address` is helpful here. Too many things are a string

csillag: That's true. In fact it's so true that I would consider adding `export type Address = string` somewhere, and then consistently using `Address` instead of `string` wherever we are talking about addresses. (But that's a story of a different PR.)

lukaw3d: Yeah :D would we consistently use it tho? maybe :D 

If we add it, we should alias it like StringifiedBigInt, or make it branded
https://github.com/oasisprotocol/oasis-wallet-web/blob/8be6fcb13735fa5dee04c8be75918078e5b6ef56/src/types/StringifiedBigInt.ts#L1-L3

src/app/state/wallet/index.ts
```diff
@@ -16,18 +16,18 @@ const slice = createSlice({
...
     fetchWallet(state, action: PayloadAction<Wallet>) {},
     updateBalance(state, action: PayloadAction<BalanceUpdatePayload>) {
-      Object.assign(state.wallets[action.payload.walletId].balance, action.payload.balance)
+      Object.assign(state.wallets[action.payload.address].balance, action.payload.balance)
     },
     addWallet(state, action: PayloadAction<AddWalletPayload>) {},
     walletOpened(state, action: PayloadAction<Wallet>) {
       const newWallet = action.payload
-      state.wallets[newWallet.id] = Object.assign({}, newWallet)
+      state.wallets[newWallet.address] = { ...newWallet }
```
buberdds, line 30: hmm I thought `immer` takes care of immutability and there is no need to use spread operator or Object.assign when assigning an object to state.

buberdds (approved):

src/app/state/wallet/index.ts
lukaw3d: Alright I'll risk it :D

PR #1020, opened 2022-09-22
Remove executable flag from source files

lukaw3d:


csillag (approved):
Reviewed. LGTM.

It would be nice to check for these in the `lint` or `lint:fix` scripts  (defined in package.json), using something like  `find . -executable -type f`), but I guess we should avoid adding unix system commands to `package.json`, so that windows developers can also use it... right?

lukaw3d:
:shrug: I don't think it matters enough to maintain an automated system for it, with all the ignored folders.

Windows developers: we haven't been trying to make it compatible, but I think things so far would work in git-bash

lukaw3d:
but cmd.exe would crash at `"start": "REACT_APP_META_CSP= HMR_PORT=2222 `

csillag:
If we don't care about windows compatibility, then we could potentially add these very simple scripts: https://github.com/oasisprotocol/oasis-wallet-web/commit/ef7a2027654186e804acaf51d562ea3142f0a167

... and then `yarn fix-executables` would do the trick.

lukaw3d:
More clutter that no-one uses in package.json :cry: 

https://github.com/oasisprotocol/oasis-wallet-web/commit/ef7a2027654186e804acaf51d562ea3142f0a167 doesn't seem accurate:
- we have .sh file, that is reasonably executable
- I think it would have trouble with files with spaces
- I don't know about the extra folders like .parcel-cache and build

csillag:
> * we have .sh file, that is reasonably executable

If you mean ` .github/release.sh `, then it's excluded by the `| grep -v \\.git` part.

> *  think it would have trouble with files with spaces

True. Good thing we have none of those. But it's a valid concern, since we could have.

> * I don't know about the extra folders like .parcel-cache and build

I checked, and there is nothing executable in there.

   * * *
But we can drop it, it's not worth the time talking about it. (I'm removing the branch.)

PR #1027, opened 2022-09-24
Add `yarn clean:deep` which purges `node_modules`, too

csillag:
... since it's recreated as part of the installation anyway.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "build": "node ./internals/prod-web.js",
     "build:ext": "node ./internals/prod-ext.js",
     "clean": "rm -rf build/ .parcel-cache",
+    "clean:deep": "yarn clean; rm -rf node_modules",
     "clean:ext": "rm -rf build-ext/ .parcel-cache",
     "start": "REACT_APP_META_CSP= HMR_PORT=2222 EXTENSION_CSP=`yarn --silent print-extension-csp` parcel --host localhost --port 3000 --dist-dir build-dev",
     "cypress:open": "cypress open",


buberdds (commented):
build task is broken

csillag:
> build task is broken

Build is now fixed, but re-installing the npm packages takes an extra 41 secs, so it will slow down the CI a little bit. I don't know if it's acceptable. If it's not, then instead of purging the `node_modules` in `yarn clean`, I will just add another target called 'clean:deep", which does all that the original `clean` did, plus the `node_modules`.

buberdds (approved):

package.json
```diff
@@ -11,6 +11,7 @@
     "build": "node ./internals/prod-web.js",
     "build:ext": "node ./internals/prod-ext.js",
     "clean": "rm -rf build/ .parcel-cache",
+    "cleen:deep": "yarn clean; rm -rf node_modules",
```
buberdds, line 14: typo

package.json
```diff
@@ -11,6 +11,7 @@
     "build": "node ./internals/prod-web.js",
     "build:ext": "node ./internals/prod-ext.js",
     "clean": "rm -rf build/ .parcel-cache",
+    "clean:deep": "yarn clean; rm -rf node_modules",
```
lukaw3d, line 14: When does someone need to delete whole node_modules?? Did you mean node_modules/.cache?

csillag: Whenever I want to remove all build artifacts. It can be restored by a simple yarn command, and part of .gitignore, so its not really part of the codebase..

lukaw3d: Build artifacts are in build/, build-ext/, and build-dev/

Sounds like you're trying to do `git clean -fdx`, but I still don't know why

csillag: > Sounds like you're trying to do `git clean -fdx`, but I still don't know why

Almost, but not _exactly_. I'm trying to reduce the directory to the smallest possible size, without any chance of losing any real (ie. non-derived) information.
 - The current `yarn clean` doesn't do that, because it leaves the `node_modules` directory lying around, which is currently 780MB in my case.
 - The proposed `git clean -fdx` _also_ doesn't do that, because it will remove untracked files, which may very well be new code.

So I still think there is valid use-case for this.

csillag:
As discussed, this is no longer necessary. (I will just use my custom Makefile)

PR #1037, opened 2022-09-27
When amount is an empty string, interpret it as 0

### LINKED ISSUE #1031: Clicking "Send" on an empty transfer throws an error
To reproduce:
- open a wallet
- click Send

crash

csillag:
Fixes #1031


buberdds:
I am not exactly sure what are expectations here. Do we really don't want to catch an error when a function input is invalid? Why we even trigger a submit event when a form is empty?

csillag:
> I am not exactly sure what are expectations here. Do we really don't want to catch an error when a function input is invalid? Why we even trigger a submit event when a form is empty?

The form has been implemented in such a way that when nothing is entered for the amount, it displays a "0". So it's not an invalid value, but an implicit 0. 

I asked about it, and a 0 seems to be a valid value.

So I came to the conclusion that 
 - an empty value should mean 0
 - 0 is a valid value

That's why I modified the conversion function so that we interpret "" as 0.
The other option would have been to implement this conversion at the form level, but I wanted to catch the biggest possible group of errors (ie. we might be using the same conversion function in other forms in the future), hence the implementation.

src/app/lib/helpers.ts
```diff
@@ -51,8 +51,13 @@ export function concat(...parts: Uint8Array[]) {
...
 }
 
+/**
+ * Parse a ROSE (or TEST) amount from a string.
+ *
+ * An empty string will be interpreted as 0.
+ */
 export function parseRoseStringToBaseUnitString(value: string): StringifiedBigInt {
-  const baseUnitBN = new BigNumber(value).shiftedBy(9) // * 10 ** 9
+  const baseUnitBN = new BigNumber(value || '0').shiftedBy(9) // * 10 ** 9
```
lukaw3d, line 60: This function shouldn't be responsible for handling this

lukaw3d (commented):
Looks like `onSubmit` needs to be moved from `<Button type="submit" label={t('account.sendTransaction.send', 'Send')} onClick={onSubmit}`  to `<Form onSubmit={onSubmit}`

csillag:
> Looks like `onSubmit` needs to be moved from `<Button type="submit" label={t('account.sendTransaction.send', 'Send')} onClick={onSubmit}` to `<Form onSubmit={onSubmit}`

OK, but then we have to adjust the fields a little bit, in order to keep the ability to submit the form without touching the amount field. (I guess we want that, since we had a "0" placeholder on the amount field. ) Done that.

src/app/lib/helpers.ts
csillag: > I am not exactly sure what are expectations here. Do we really don't want to catch an error when a function input is invalid? Why we even trigger a submit event when a form is empty?

@buberdds your concerns have been addressed in the new implementation. (We no longer submit the form when the field is empty.)

src/app/pages/AccountPage/Features/SendTransaction/__tests__/index.test.tsx
```diff
@@ -20,12 +20,23 @@ describe('<SendTransaction />', () => {
...
 
+  it('should NOT dispatch sendTransaction action when amount is missing', async () => {
+    const spy = jest.spyOn(store, 'dispatch')
+    renderComponent(store)
+
+    await userEvent.type(screen.getByPlaceholderText('account.sendTransaction.enterAddress'), 'walletAddress')
+    await userEvent.clear(screen.getByPlaceholderText('account.sendTransaction.enterAmount'))
+    spy.mockClear()
+    await userEvent.click(screen.getByRole('button'))
+    expect(spy).toHaveBeenCalledTimes(0)
```
lukaw3d, line 31: Jest also has slightly more readable `.not.toHaveBeenCalled()`

lukaw3d (approved):

src/app/pages/AccountPage/Features/SendTransaction/__tests__/index.test.tsx
```diff
@@ -20,12 +20,23 @@ describe('<SendTransaction />', () => {
     store = configureAppStore()
   })
```
lukaw3d, line 22: Could you extend an e2e test too

```diff
  it('Should send a transaction', () => {
+   // Expect not to crash
+   cy.findByRole('button', { name: /Send/ }).click()
+
    cy.get('#recipient-id').type('oasis1qz0k5q8vjqvu4s4nwxyj406ylnflkc4vrcjghuwk')
    cy.get('#amount-id').type('10')
    cy.findByRole('button', { name: /Send/ }).click()
    cy.findByRole('button', { name: /Confirm/ }).click()
    cy.contains('Transaction successfully sent').should('be.visible')
  })
```

csillag: Extended the cypress test.

csillag:
Meanwhile, part of this has already been covered by https://github.com/oasisprotocol/oasis-wallet-web/pull/1070.

What remains here:
 - Jest test for the error condition
 - Cypress test for the error condition
 - Better placeholder value for the send form (to make it obvious that an amount must be entered.)

src/app/pages/AccountPage/Features/SendTransaction/index.tsx
```diff
@@ -25,7 +25,7 @@ export function SendTransaction(props: SendTransactionProps) {
   const { error, success } = useSelector(selectTransaction)
   const validators = useSelector(selectValidators)
   const [recipient, setRecipient] = useState('')
-  const [amount, setAmount] = useState('')
+  const [amount, setAmount] = useState('0')
```
buberdds, line 28: We really don't want to show a placeholder to user?

buberdds (approved):

src/app/pages/AccountPage/Features/SendTransaction/index.tsx
csillag: The idea was that the amount field should be initialized in a valid state, which can be submitted right away.
If we show the "0" as a placeholder (as before), it gives the false impression that the field is valid, when in fact it is not.

With this change, we actually put 0 into the field as the initial value, so it's valid.

And if it's removed, then we show the real placeholder, see below.

> placeholder={t('account.sendTransaction.enterAmount', 'Enter an amount')}

csillag: > I would validate this with Don.

Discussed with Don; we will keep the empty starting value (but replace the placeholder.)

csillag:
Got approval from Don about the UI/UX perspective. Technical review concluded earlier. Merging...

PR #1044, opened 2022-09-28
Migration to React 18

buberdds:
- cannot update all React related deps due to issues with `webext-redux`
- entry in `resolutions` is needed to avoid `act` warnings in tests when using userEvents
- there are still 3 warnings in loadable tests, but I have no idea how to avoid this, because warnings will be thrown when suspended data is not resolved and these tests cover such case
- `importAccountsActions.clear()` action had to be moved to saga. This UI part was not working with a new React 18 strict mode


src/utils/__tests__/loadable.test.tsx
```diff
@@ -42,10 +42,9 @@ describe('loadable', () => {
   })
```
lukaw3d, line 42: Oddly this works:
```tsx
  it('should render fallback if given one', async () => {
    const component = render(<LazyComponentWithFallback />)
    expect(await screen.findByText(/lazy-loaded/)).toBeInTheDocument()
    expect(component.container.firstChild).toMatchSnapshot()
  })
```
but this complains `Warning: A suspended resource finished loading inside a test, but the event was not wrapped in act(...).`:
```tsx
  it('should render fallback if given one', async () => {
    const component = render(<LazyComponentWithFallback />)
    expect(await screen.findByText(/Loading/)).toBeInTheDocument()
    expect(component.container.firstChild).toMatchSnapshot()
    expect(await screen.findByText(/lazy-loaded/)).toBeInTheDocument()
    expect(component.container.firstChild).toMatchSnapshot()
  })
```

src/app/state/wallet/saga.ts
```diff
@@ -77,7 +78,7 @@ export function* openWalletFromPrivateKey({ payload: privateKey }: PayloadAction
   const walletAddress = yield* call(publicKeyToAddress, publicKeyBytes)
   const publicKey = uint2hex(publicKeyBytes)
   const balance = yield* call(getBalance, publicKeyBytes)
-
+  yield* put(importAccountsActions.clear())
```
lukaw3d, line 81: Please split into a separate commit

buberdds: done

lukaw3d (approved):

PR #1051, opened 2022-09-30
Restrict Node.js version

csillag:
... to exclude at least Node.js v18, since currently our cypress tests behave differently with that version

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,6 +177,9 @@
     "store": "./src/store",
     "vendors": "./src/vendors"
   },
+  "engines": {
+    "node": ">=14 <18"
+  },
   "targets": {
     "// see also .parcelrc and .terserrc": {},
     "web": {


buberdds (approved):

csillag:
Something is not quite right here, since now `yarn start` gets me:

```
yarn run v1.22.19
$ REACT_APP_META_CSP= HMR_PORT=2222 EXTENSION_CSP=`yarn --silent print-extension-csp` parcel --host localhost --port 3000 --dist-dir build-dev
Server running at http://localhost:3000
🚨 Build failed.

@parcel/core: Only browser targets are supported in serve mode
```

This is definitely triggered by the node version option added to package.json. Happens with both node.js v14 and v16.

csillag:
Turns out the the issue was _not_ node v18, but something else, so there is no need for this at all. Closing.

Issue #1064, opened 2022-10-05
Replace click listeners on submit buttons with submit listeners on forms


lukaw3d:
Related to https://github.com/oasisprotocol/oasis-wallet-web/pull/1037#pullrequestreview-1124333939

Although accessibility is good in both cases (both trigger submit on <kbd>Enter</kbd> and on click), only `<Form onSubmit` handles validation. Fix it upstream or add types+lint to enforce correct usage.

Before:
```tsx
<Form>
  <Button type="submit" onClick={onSubmit}>
```

After:
```tsx
<Form onSubmit={onSubmit}>
  <Button type="submit">
```

lukaw3d:
Sure!
I'll keep the issue open for the "Fix it upstream or add types+lint to enforce correct usage" part

PR #1070, opened 2022-10-07
fix 1064: submit forms not just on submit button click

alexander-elgin:
Part of oasisprotocol/oasis-wallet-web#1064

### DIFF
--- a/src/app/components/MnemonicValidation/index.tsx
+++ b/src/app/components/MnemonicValidation/index.tsx
@@ -76,11 +76,7 @@ export function MnemonicValidation(props: Props) {
             </FormField>
           </Box>
           <Box direction="row" gap="small" margin={{ top: 'medium' }}>
-            <Button
-              type="submit"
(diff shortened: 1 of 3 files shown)


pro-wh (approved):
Implemented as requested, thanks!

src/app/components/MnemonicValidation/index.tsx
```diff
@@ -80,7 +80,6 @@ export function MnemonicValidation(props: Props) {
               type="submit"
               label={t('openWallet.mnemonic.import', 'Import my wallet')}
               primary
-              onClick={onSubmit}
```
pro-wh, line 83: the linter wants us to collapse this into one line after the removal. @alexander-elgin would you be able to make that change for us too?

alexander-elgin: @pro-wh Done. Sorry, I did not check it at the first time

pro-wh:
thanks, merging

PR #1079, opened 2022-10-12
Fix extension WebUSB permission issue

### LINKED ISSUE #1045: Extension throws LedgerNoDeviceSelected
Trying to use ledger in extension always throws LedgerNoDeviceSelected. The reason seems to be that background page can't ask user to give permission to selected USB device; and we run saga in background page.

One potential solution is to ask user for USB permission in the popup, before dispatching to saga:
```tsx
onClick={async () => {
  if (await TransportWebUSB.isSupported()) await TransportWebUSB.create()
  dispatch(importAccountsActions.enumerateAccountsFromLedger())
}}
```

But that actually doesn't work, because default popup ALSO can't ask for permission. It does work in the other kind of popup tho:
```ts
chrome.windows.create({
  url: location.href,
  type: 'popup',
  width: 500,
  height: 600,
})
```
https://github.com/oasisprotocol/oasis-wallet-ext/blob/990a8c78999a2c594d2f0a11991a73a5ca214f9f/src/utils/ledger.js#L28 does this

Note: I only checked this in Chrome, with manifest v2

Debugging note: after granting USB permission you can remove it from extension in chrome://settings/content/usbDevices?search=usb

buberdds:
Fixes #1045
During sync meeting we decided to go with a popup like in a current ext.
Screenshot 2022-10-12 at 18 27 51
Screenshot 2022-10-12 at 18 28 06
Screenshot 2022-10-12 at 18 28 56
Screenshot 2022-10-12 at 18 29 04
Screenshot 2022-10-12 at 18 29 13

error:
Screenshot 2022-11-08 at 13 14 15

### DIFF
--- a/src/app/pages/ConnectDevicePage/index.tsx
+++ b/src/app/pages/ConnectDevicePage/index.tsx
@@ -1,0 +1,134 @@
+import React, { useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { useDispatch } from 'react-redux'
+import { Box, Button, Spinner, Text } from 'grommet'
+import { StatusCritical, StatusGood } from 'grommet-icons'
+import { Header } from 'app/components/Header'
+import { ErrorFormatter } from 'app/components/ErrorFormatter'
+import { AlertBox } from 'app/components/AlertBox'
+import { WalletErrors } from 'types/errors'
+import { importAccountsActions } from 'app/state/importaccounts'
+import { requestDevice } from 'app/lib/ledger'
+import logotype from '../../../../public/logo192.png'
+
+type ConnectionStatus = 'connected' | 'disconnected' | 'connecting' | 'error'
+type ConnectionStatusIconPros = {
+  success?: boolean
+  label: string
+  withMargin?: boolean
+}
+
+function ConnectionStatusIcon({ success = true, label, withMargin = false }: ConnectionStatusIconPros) {
+  return (
+    <Box
+      align="center"
+      direction="row"
+      gap="small"
+      justify="center"
+      margin={{ bottom: withMargin ? 'large' : 'none' }}
+    >
+      {success ? (
+        <StatusGood color="successful-label" size="30px" />
+      ) : (
+        <StatusCritical color="status-error" size="30px" />
+      )}
+      <Text weight="bold" size="large" textAlign="center">
+        {label}
+      </Text>
+    </Box>
+  )
+}
+
+export function ConnectDevicePage() {
+  const { t } = useTranslation()
+  const dispatch = useDispatch()
+  const [connection, setConnection] = useState<ConnectionStatus>('disconnected')
+  const handleConnect = async () => {
+    setConnection('connecting')
+    try {
+      const device = await requestDevice()
+      if (device) {
+        setConnection('connected')
+        dispatch(importAccountsActions.enumerateAccountsFromLedger())
+      }
+    } catch {
+      setConnection('error')
+    }
+  }
+
+  return (
+    <Box
+      style={{ minHeight: '100vh' }}
+      justify="center"
+      align="center"
+      pad="medium"
+      background="background-back"
+    >
+      <Box
+        elevation="small"
+        round="5px"
+        background="background-front"
+        pad="large"
+        style={{ position: 'relative' }}
+      >
+        <Box margin={{ bottom: 'medium' }} align="center">
+          <img src={logotype} alt="Oasis" width="75" height="75" />
+        </Box>
+        <Header textAlign="center">{t('ledger.extension.grantAccess', 'Grant access to your Ledger')}</Header>
+        <Box gap="medium">
+          <ol>
+            <li>
+              {t('ledger.instructionSteps.connectLedger', 'Connect your Ledger device to the computer')}
+            </li>
+            <li>{t('ledger.instructionSteps.closeLedgerLive', 'Close Ledger Live app on the computer')}</li>
+            <li>{t('ledger.instructionSteps.openOasisApp', 'Open the Oasis App on your Ledger device')}</li>
+            <li>
+              {t(
+                'ledger.extension.instructionStep',
+                'Once device is connected continue the operation in the wallet app',
+              )}
+            </li>
+          </ol>
+
+          {connection === 'connecting' && (
+            <Box
+              style={{ position: 'absolute', top: 0, right: 0, bottom: 0, left: 0 }}
+              background={{
+                color: 'background-front',
+                opacity: 'medium',
+              }}
+              align="center"
+              justify="center"
+            >
+              <Spinner size="medium" />
+            </Box>
+          )}
+          {connection === 'connected' && (
+            <ConnectionStatusIcon label={t('ledger.extension.succeed', 'Device connected')} />
+          )}
+          {connection === 'error' && (
+            <Box margin={{ bottom: 'medium' }}>
+              <ConnectionStatusIcon
+                success={false}
+                label={t('ledger.extension.failed', 'Connection failed')}
+                withMargin
+              />
+              <AlertBox color="status-error">
+                <ErrorFormatter code={WalletErrors.LedgerNoDeviceSelected} />
+              </AlertBox>
+            </Box>
+          )}
+          {connection !== 'connected' && (
+            <Box justify="center">
+              <Button
+                onClick={handleConnect}
+                label={t('ledger.extension.connect', 'Connect Ledger device')}
+                primary
+              />
+            </Box>
+          )}
+        </Box>
+      </Box>
+    </Box>
+  )
+}
--- a/src/app/pages/OpenWalletPage/__tests__/index.test.tsx
+++ b/src/app/pages/OpenWalletPage/__tests__/index.test.tsx
@@ -1,0 +1,30 @@
+import React from 'react'
+import { render, screen } from '@testing-library/react'
+import { MemoryRouter } from 'react-router-dom'
+import { SelectOpenMethod } from '..'
+
+jest.mock('react-redux', () => ({
+  useSelector: jest.fn(),
+}))
+
+const renderComponent = (webExtensionLedgerAccess?: () => void) =>
+  render(
+    <MemoryRouter>
+      <SelectOpenMethod webExtensionLedgerAccess={webExtensionLedgerAccess} />
+    </MemoryRouter>,
+  )
+
+describe('<SelectOpenMethod />', () => {
+  it('should render component', () => {
+    const { container } = renderComponent()
+
+    expect(container).toMatchSnapshot()
+  })
+
+  it('should render component with an access button', () => {
+    renderComponent(() => {})
+
+    expect(screen.queryByText('openWallet.method.ledger')).not.toBeInTheDocument()
+    expect(screen.getByText('ledger.extension.grantAccess')).toBeInTheDocument()
+  })
+})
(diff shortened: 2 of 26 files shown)


src/app/pages/OpenWalletPage/webextension.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+import { openTab } from 'utils/webextension'
+import { SelectOpenMethod } from './'
+
+export function OpenWalletPageWebExtension() {
+  return (
+    <SelectOpenMethod webExtensionLedgerAccess={() => openTab(`${window.location.href}/connect-device`)} />
```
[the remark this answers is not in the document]
buberdds: None of `useLocation, useNavigation, useHref, useResolvedPath` hooks return a value of `window.location.href` which I need here `chrome-extension://<extension_id>/popup.<hash>.html#/open-wallet/connect-device`. Didn't find alternatives for setting base href in popup.html as well. Any other idea?

src/index.tsx
```diff
@@ -29,20 +27,20 @@ import './locales/i18n'
...
   <Provider store={store}>
     <ThemeProvider>
       <HelmetProvider>
-        <BrowserRouter>
-          <React.StrictMode>
-            <App />
-          </React.StrictMode>
-        </BrowserRouter>
+        <React.StrictMode>
+          <RouterProvider router={router} />
```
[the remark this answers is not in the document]
lukaw3d: Hm. Previous extension code generated links like `<a href="#/create-wallet">` that navigated to `#/create-wallet`. Now it generates `<a href="/create-wallet">` and navigates to `#/create-wallet`. Middle clicking links with hash-router now opens a broken path.

Not sure if this is an upstream issue or misconfiguration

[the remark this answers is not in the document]
lukaw3d: They don't have a single test for https://github.com/remix-run/react-router/search?q=createHashRouter :angry: and this poop has 8 million weekly downloads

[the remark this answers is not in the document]
lukaw3d: Fixed by https://github.com/remix-run/react-router/pull/9409 in `react-router-dom@pre` release

src/app/pages/OpenWalletPage/webextension.tsx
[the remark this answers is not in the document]
lukaw3d: If react-router wasn't broken `useHref('connect-device')` and `useHref('/open-wallet/connect-device')` would return `#/open-wallet/connect-device` in a HashRouter. And then reliably combined with `window.location.pathname`

src/app/pages/ConnectDevicePage/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+              justify="center"
+            >
+              <Spinner size="medium" />
+            </Box>
+          )}
+          {connection === 'connected' && (
+            <ConnectionStatusIcon label={t('ledger.extension.succeed', 'Device connected')} />
+          )}
+          {connection === 'error' && (
+            <ConnectionStatusIcon success={false} label={t('ledger.extension.failed', 'Connection failed')} />
```
lukaw3d, line 104: Errors aren't as good as before

| Before - Firefox extension | After |
| --- | --- |
| image | image |

| Before - Chrome with blocked webusb | After |
| --- | --- |
| Peek 2022-10-26 02-02 | Peek 2022-10-26 02-00 |

src/app/pages/ConnectDevicePage/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+                opacity: 'medium',
+              }}
+              align="center"
+              justify="center"
+            >
+              <Spinner size="medium" />
+            </Box>
+          )}
+          {connection === 'connected' && (
+            <ConnectionStatusIcon label={t('ledger.extension.succeed', 'Device connected')} />
```
lukaw3d, line 101: Extra popup should probably automatically close after this

src/app/pages/OpenWalletPage/index.tsx
```diff
@@ -33,9 +44,18 @@ export function SelectOpenMethod() {
...
-        <span>
-          <ButtonLink to="ledger" label={t('openWallet.method.ledger', 'Ledger')} primary />
-        </span>
+        {webExtensionLedgerAccess ? (
+          <Button
+            style={{ width: 'fit-content' }}
+            onClick={webExtensionLedgerAccess}
+            label={t('ledger.extension.grantAccess', 'Grant access to your Ledger')}
+            primary
+          />
```
lukaw3d, line 53: This flow should only be needed the first time for each unpaired ledger device, not every time. Background page already has the needed permissions.

lukaw3d: How about on click: attempt to get a device and if it fails openLedgerAccessPopup

buberdds: Can you share more details? Do you want to dispatch saga action first and handle "LedgerNoDeviceSelected" differently or there is some way to get a device and checked if it's paired?

lukaw3d: I think just `try { await TransportWebUSB.create(); to="ledger" } catch (e) { openLedgerAccessPopup() }`

buberdds: If we want to create usb transport  here we will have to close it in React component as well to avoid `USB Transport error: Failed to execute 'claimInterface' on 'USBDevice': Unable to claim interface.` At this point everything looks fine, but when I will revoke usb access I go through process again and it will end up with `USB Transport error: Failed to execute 'open' on 'USBDevice': The device was disconnected..`.

src/app/pages/OpenWalletPage/__tests__/index.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+  })
+
+  it('should render component', async () => {
+    const { container } = await waitFor(() => renderComponent())
+
+    expect(container).toMatchSnapshot()
+  })
+
+  it('should render component with an access button', async () => {
+    await waitFor(() => renderComponent(() => {}))
```
lukaw3d, line 40: ??

waitFor runs the function multiple times until expect() inside is successful. So this renders infinite number of components, or I guess it does nothing because there are no expects inside

src/app/pages/OpenWalletPage/__tests__/index.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    expect(screen.queryByText('openWallet.method.ledger')).not.toBeInTheDocument()
+    expect(screen.getByText('ledger.extension.grantAccess')).toBeInTheDocument()
```
lukaw3d, line 43: I'd merge this into `should render component` test

src/app/pages/OpenWalletPage/__tests__/index.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+  it('should render variant with web usb support', async () => {
+    jest.mocked(canAccessNavigatorUsb).mockResolvedValue(true)
+    const { rerender } = await waitFor(() => renderComponent())
+    rerender(
+      <MemoryRouter>
+        <SelectOpenMethod />
+      </MemoryRouter>,
+    )
+
+    expect(screen.queryByText('errors.usbTransportNotSupported')).not.toBeInTheDocument()
```
lukaw3d, line 65: Add a test that establishes `screen.queryByText('errors.usbTransportNotSupported')` can sometimes be in the document. Otherwise `not.toBeInTheDocument` never detects if we renamed it, and error actually was shown

src/utils/webextension.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const getExtensionUrl = (path: string) =>
+  browser.runtime.getURL(`${browser.runtime.getManifest()?.browser_action?.default_popup}${path}`)
+
+const openPopup = (path: string, dimensions: Dimensions) => {
+  const existingPopupWindow = browser.extension.getViews().find(window => window.location.hash === path)
+
+  if (existingPopupWindow) {
+    existingPopupWindow.close()
+  }
```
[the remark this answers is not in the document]
buberdds: - yes, extension is using HashRouter
- added a guard
- it does not have subroutes

src/app/pages/OpenWalletPage/__tests__/index.test.tsx
buberdds: Why when it's an opposite case to what default snapshot is rendering?

src/app/pages/OpenWalletPage/__tests__/index.test.tsx
buberdds: That is handled by snapshot already.

src/utils/webextension.ts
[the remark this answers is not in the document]
lukaw3d: More robust than implicitly encoding HashRouter and adding a guard:
`.find(window => window.location.href === getPopupUrl(path))`

src/utils/webextension.ts
```diff
@@ -0,0 +1,25 @@
...
+  browser.windows.create({
```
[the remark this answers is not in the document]
lukaw3d: No, it just might be annoying if we later need to make it async to use the value `browser.windows.create` resolves to, or catch the promise rejection

src/app/pages/OpenWalletPage/__tests__/index.test.tsx
lukaw3d: Oh, oops :O

src/app/pages/OpenWalletPage/__tests__/index.test.tsx
lukaw3d: Yep yep, I'd simply add `expect(screen.getByText('errors.usbTransportNotSupported')).toBeInTheDocument()` after snapshot

src/app/pages/ConnectDevicePage/index.tsx
lukaw3d: Not sure if you changed this in an odd way: popup1 closes after returning to it, after success
popup1-disappears

src/app/pages/OpenWalletPage/index.tsx
lukaw3d: Alright I am able to reproduce this
```tsx
  try {
    await (await TransportWebUSB.create()).close()
    navigate('/open-wallet/ledger')
  } catch (e) {
    webExtensionLedgerAccess()
  }
```

what's very odd is that it works after reloading background page, as if `close()` doesn't fully reset device communication
- start with extension having USB permission
- enumerate accounts **works**, cancel
- remove USB permission
- connect ledger flow, enumerate accounts **does not work**, cancel
- reload background page
- enumerate accounts works, cancel
- remove USB permission
- reload background page
- connect ledger flow, enumerate accounts **works**, cancel

src/app/pages/ConnectDevicePage/index.tsx
lukaw3d: | Updated error in Firefox (reproduced by temporarily installing ext in `about:debugging#/runtime/this-firefox`) | Chrome with blocked webusb is still stuck (reproduced with `Don't allow sites to connect to USB devices` in chrome://settings/content/usbDevices) |
| --- | --- |
| Screenshot from 2022-11-08 03-40-27  | Peek 2022-11-08 03-47  |

buberdds: No specific error is thrown in this case just generic "No device selected". I will show ErrorFormatter in this case

src/app/pages/ConnectDevicePage/index.tsx
buberdds: checked both dev and prod builds and I cannot repro this

src/utils/webextension.ts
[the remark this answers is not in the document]
lukaw3d: (`if (path` guard isn't needed anymore)

lukaw3d (approved):

PR #1081, opened 2022-10-12
Refactor password field validation and restyle

lukaw3d:
I tried to make validation part of the input field, so that onSubmit doesn't need to re-check validity of every input field. But then error-specific-styling doesn't work `contentProps={{ border: props.error ? 'bottom' : false }}`.

| Before | After |
| --- | --- |
|  localhost_3000_open-wallet_private-key (2) | localhost_3000_open-wallet_private-key  |

### DIFF
--- a/src/app/components/PasswordField/index.tsx
+++ b/src/app/components/PasswordField/index.tsx
@@ -2,36 +2,45 @@ import { Box, FormField, Button, TextInput, Tip } from 'grommet'
 import { View, Hide } from 'grommet-icons'
 import * as React from 'react'
 
-interface Props {
-  placeholder: string
+interface Props<TFormValue> {
+  name: Extract<keyof TFormValue, string>
+  label?: string
+  placeholder?: string
   inputElementId: string
 
   autoComplete: 'on' | 'off' | 'new-password' | 'current-password'
-  value: string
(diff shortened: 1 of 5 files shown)


src/app/components/PasswordField/index.tsx
```diff
@@ -2,39 +2,45 @@ import { Box, FormField, Button, TextInput, Tip } from 'grommet'
...
+  label?: string
+  placeholder?: string
   inputElementId: string
 
   autoComplete: 'on' | 'off' | 'new-password' | 'current-password'
-  value: string
-  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void
-  error: string | false
+  value?: string
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void
```
buberdds, line 13: It looks like these type changes does not belong to this PR. I don't see any use cases for optional `value` and `onChange` or 'TFormValue' here.

lukaw3d: The new recommended usage is with `name` and `validate`, but `value`, `onChange`, `error` still come useful. And keeping them is consistent with grommet's FormField.

I added some tests to validate both uses

buberdds (approved):

PR #1097, opened 2022-10-21
Refactor HomePage buttons into ButtonLinks to improve accessibility

lukaw3d:
Fixes issues:
- Keyboard navigating through these buttons would first focus on the link, and then on the button
- Multiple buttons were marked as type="submit" without being in multiple forms
- On a small screen "Mnemonic / Private key / Ledger" buttons would collapse into rows of left-aligned buttons, but links would fill up the whole row

### DIFF
--- a/src/app/components/ButtonLink/index.tsx
+++ b/src/app/components/ButtonLink/index.tsx
@@ -1,0 +1,20 @@
+import { Button, ButtonExtendedProps } from 'grommet'
+import { Link, To } from 'react-router-dom'
+import React from 'react'
+
+type ButtonLinkProps = Omit<ButtonExtendedProps, 'label'> & { label: string; to: To }
+
+/**
+ * Behaves like {@link Link} and styled like {@link Button}. Needed because:
+ * - Link isn't styled but routes correctly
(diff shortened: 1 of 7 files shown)


src/app/components/ButtonLink/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+type ButtonLinkProps = LinkProps & ButtonExtendedProps
+
+/**
+ * Behaves like {@link Link} and styled like {@link Button}. Needed because:
+ * - Link isn't styled but routes correctly
+ * - Button is styled but can't navigate
+ * - Wrapping Link around Button is bad for accessibility
+ */
+export const ButtonLink = (props: ButtonLinkProps) => {
+  return <Button as={Link} {...props} />
```
buberdds, line 14: This throws tons of console warnings, because props are assigned as HTML attributes. HTML markup is also wrong `theme="[object Object]"`

buberdds:
`findByRole('link')` in cypress tests

csillag (approved):
Reviewed. LGTM.

buberdds (approved):

PR #1098, opened 2022-10-21
Restyle default tooltips and toast notifications

lukaw3d:
| Before | After |
| --- | --- |
|    | b c |
|   |   |

Adds a solid background and removes the shadow to be consistent with SidebarTooltip 
localhost_3000_account_oasis1qqetax7q9gh5jxasjqj2gkpvntwamudc75vnt0rl

### DIFF
--- a/src/styles/theme/ThemeProvider.tsx
+++ b/src/styles/theme/ThemeProvider.tsx
@@ -137,6 +137,10 @@ const grommetCustomTheme: ThemeType = {
   notification: {
     toast: {
       time: 2000,
+      container: {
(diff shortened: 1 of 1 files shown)


csillag:
Shouldn't we do something similar about the notifications,  too? The shadow / elevation seems to be an overkill.

image

csillag (approved):
The change is OK.
(It would be nice adjust Notification, too, but from a cursory look of the available settings, I'm not sure it's trivially possible.)

lukaw3d:
Good reminder!

PR #1111, opened 2022-10-26
Fix ethereumjs imports

buberdds:


lukaw3d (approved):

buberdds:
This package is not working in new extension. If I won't find workaround I will revert this change.

PR #1117, opened 2022-10-27
Update @oasisprotocol dependencies

buberdds:


buberdds:
Error: Webpack Compilation Error ./node_modules/@oasisprotocol/client/dist/signature.js 146:42

pro-wh:
what was the error?

pro-wh:
ah

```
Module parse failed: Unexpected token (146:42)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|         if (parts.length === 2) {
|             const [context2, chainContext] = parts;
>             if (handlers.withChainContext?.[context2]) {
|                 handlers.withChainContext[context2](chainContext, misc.fromCBOR(message));
|                 return true;
 @ ./node_modules/@oasisprotocol/client/dist/index.js 21:20-42
```

pro-wh:
so what all is going on here? this project uses parcel to build, but tests with cypress and webpack 4?

buberdds:
It has to transpile a cypress test file which is "unique". It imports our vendor files
https://github.com/oasisprotocol/oasis-wallet-web/blob/master/cypress/e2e/api-transaction-methods.cy.ts#L1

The only ts target that changes something is es2020, but there's a bug with nullish coalescing so libs will be transpiled, but error will occur while transpiling our vendor file. I didn't look at Cypress preprocessor option. We can also extract parsers like `parseTransactionsList` into external file as a workaround. Not sure it's worth doing as we are switching to Playwright? Is that correct @lukaw3d ?

lukaw3d:
I wasn't able to migrate encryption-unit-tests and api-transaction-methods from Cypress to Playwright :/

api-transaction-methods could easily be migrated to jest though. And encryption-unit-tests shouldn't hit these syntax issues.

Another option is to serve a page with `import * as monitor from 'vendors/monitor'; window.monitor = monitor;` and inject JS `cy.window().then((win) => win.monitor.parseTransactionsList([transactionFromApi]))`

PR #1120, opened 2022-10-28
Remove IE11 and Android 4.4.x from the list of supported browsers

buberdds:
Workaround for a transpiler issue with bigints/bigintegers after switching to `@ethereumjs/util` package. After removing these two browsers from `browserlist` extension works correctly.

Related to https://github.com/oasisprotocol/oasis-wallet-web/pull/1111#issuecomment-1293603751

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,9 @@
       ">0.2%",
       "not dead",
       "not op_mini all",
-      "not ios_saf < 13"
+      "not ios_saf < 13",
+      "not android <= 4.4.4",
+      "not ie <= 11"
     ],
     "development": [
       "last 1 chrome version",
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -5,7 +5,6 @@
  * code.
  */
 
-import 'react-app-polyfill/ie11'
 import 'react-app-polyfill/stable'
 
 import * as React from 'react'
--- a/src/setupTests.ts
+++ b/src/setupTests.ts
@@ -3,7 +3,6 @@
 // this adds jest-dom's custom assertions
 import '@testing-library/jest-dom/extend-expect'
 
-import 'react-app-polyfill/ie11'
 import 'react-app-polyfill/stable'
 import 'portable-fetch'
 


package.json
```diff
@@ -35,7 +35,9 @@
       ">0.2%",
       "not dead",
       "not op_mini all",
-      "not ios_saf < 13"
+      "not ios_saf < 13",
+      "not android 4.4.3-4.4.4",
+      "not ie 11"
```
pro-wh, line 40: could we say not ie altogether :laughing:

package.json
```diff
@@ -35,7 +35,9 @@
       ">0.2%",
       "not dead",
       "not op_mini all",
-      "not ios_saf < 13"
+      "not ios_saf < 13",
+      "not android 4.4.3-4.4.4",
```
pro-wh, line 39: similarly here, would a < be suitable? or are we supporting 4.4.2?

pro-wh (commented):
implemented as described

what was the issue though?

package.json
buberdds: all other versions are covered by "not dead" entry, but sure we can have something like 
```
      "not android <= 4.4.4",
      "not ie <= 11"
```

pro-wh: o lol dead

pro-wh:
what was the transpiler issue with bigints/bigintegers after switching to @ethereumjs/util package?

buberdds:
`Uncaught TypeError: Cannot convert a BigInt value to a number at Math.pow` https://github.com/paulmillr/noble-ed25519/issues/23#issuecomment-864388973

pro-wh:
what will removing these browsers do? allow us to use native bigint and `**`?

pro-wh:
what's the context of that link btw? I thought we weren't using noble-ed25519

pro-wh (approved):
overall I'm okay with removing support for these browsers. but I feel like there are things that we brought up within this PR that we should talk about

pro-wh:
and not on the critical path, is there a transpiler issue tracker where we can report this issue?

buberdds:
- noble-ed25519 is @ethereumjs/util dependency
- issue was reported in parcel bundler a while ago  https://github.com/parcel-bundler/parcel/issues/7310

pro-wh:
> * noble-ed25519 is @ethereumjs/util dependency
> * issue was reported in parcel bundler a while ago  [replacing `**` with `Math.pow` breaks the code when operands are `BigInt`s parcel-bundler/parcel#7310](https://github.com/parcel-bundler/parcel/issues/7310)

thanks for the answers! one last thing:

what does removing IE11 and Android 4.4.x do in connection to that transpiling issue? were these the last browsers that didn't have native `**` support?

pro-wh:
oh shoot, and that parcel issue only links to a babel issue where they refused to fix it :skull:

Issue #869, opened 2022-06-21
Ledger crashes after enumerating twice without waiting


lukaw3d:
## Crash

- After https://github.com/oasisprotocol/oasis-wallet-web/pull/864/commits/4eb371ed17652bb7ef617a1765989abca54b40a5
- open /open-wallet/ledger
- click "Select accounts to open"
- quickly press <kbd>Esc</kbd>
- click "Select accounts to open"
- Enumerating never finishes, and Ledger device needs to be restarted

This probably happens because `takeEvery` calls `enumerateAccounts` again, calling `getUSBTransport` before old `transport.close` https://github.com/oasisprotocol/oasis-wallet-web/blob/ad9591363d53216551537d37603ff67fb01fa2e4/src/app/state/ledger/saga.ts#L92
https://github.com/oasisprotocol/oasis-wallet-web/blob/ad9591363d53216551537d37603ff67fb01fa2e4/src/app/state/ledger/saga.ts#L72-L74

## Solution

If we used `takeLatest`, it would cancel old `enumerateAccounts` and call `transport.close` first, before calling `enumerateAccounts` again. But cancellation doesn't wait for `finally` to finish https://redux-saga.js.org/docs/advanced/TaskCancellation/#note, so it still doesn't work.

Maybe reimplement `takeLatest` with additional timeout?
```ts
export function* ledgerSaga() {
  let enumerateTask = null
  while (yield* take(ledgerActions.enumerateAccounts)) {
    if (enumerateTask?.isRunning()) {
      yield* cancel(enumerateTask)
      // Workaround: cancellation doesn't wait for `finally` to finish
      yield new Promise(r => setTimeout(r, 2000))
    }
    enumerateTask = yield* fork(enumerateAccounts)
  }
```

Maybe just disable <kbd>Esc</kbd> again, and write a long comment?

PR #1124, opened 2022-11-01
Implement pagiation on account import

csillag:


buberdds:
UI issues
- horizontal scroll bar in extension
- Screenshot 2022-11-03 at 11 56 51
- possible to generate fatal error while fetching Ledger accounts (fetch and click next)
- Screenshot 2022-11-03 at 11 58 48
- UI is flickering while loading balances

src/app/state/importaccounts/saga.ts
```diff
@@ -127,6 +158,8 @@ export function* sign<T>(signer: LedgerSigner, tw: oasis.consensus.TransactionWr
 }
 
 export function* importAccountsSaga() {
-  yield* takeEvery(importAccountsActions.enumerateAccountsFromLedger, enumerateAccountsFromLedger)
+  yield* takeEvery(importAccountsActions.enumerateFirstAccountsFromLedger, enumerateSomeAccountsFromLedger)
+  yield* takeEvery(importAccountsActions.enumerateMoreAccountsFromLedger, enumerateSomeAccountsFromLedger)
```
buberdds, line 162: Do we need two separate actions for Ledger? Maybe we don't need `accounts = []`  in `enumerateAccountsFromLedger`  if that was a case.

src/app/state/importaccounts/index.ts
```diff
@@ -15,23 +23,37 @@ const slice = createSlice({
...
+    },
+    accountGenerated(state, action: PayloadAction<ImportAccountsListAccount>) {
+      state.accounts.push(action.payload)
     },
     accountsListed(state, action: PayloadAction<ImportAccountsListAccount[]>) {
       state.accounts = action.payload
     },
+    updateAccount(state, action: PayloadAction<ImportAccountChange>) {
+      const change = action.payload
+      Object.assign(state.accounts[change.index], change)
```
buberdds, line 52: we care only about balance 
`state.accounts[action.payload.index].balance = action.payload.balance`

src/app/state/importaccounts/saga.ts
```diff
@@ -71,33 +77,58 @@ function* enumerateAccountsFromMnemonic(action: PayloadAction<string>) {
...
+function* ensureBalancesArePresent(action: PayloadAction<number>) {
+  const existingAccounts = yield* select(selectImportAccountsDisplayList)
+  yield* setStep(ImportAccountsStep.LoadingBalances)
+  for (let i = 0; i < existingAccounts.length; i++) {
+    const account = existingAccounts[i]
+    if (!account.balance) {
+      const balance = yield* call(getBalance, hex2uint(account.publicKey))
+      yield* put(
+        importAccountsActions.updateAccount({
+          index: account.index,
```
buberdds, line 89: do we really need account.index ? Data structure looks like 
0: { index: 0 ...}
1: { index: 1 ...} 
in UI you can extract index from derivation path if you need it in selection modal

src/app/components/Toolbar/Features/AccountSelector/index.tsx
```diff
@@ -70,6 +71,11 @@ export const Account = memo((props: AccountProps) => {
           <CheckBox checked={props.isActive} />
         </Box>
       )}
+      <Box alignSelf="center" pad={{ left: 'small', right: 'small' }} style={{ minWidth: '2.5em' }}>
+        <Text weight="bold" style={{ width: '2em' }}>
+          {props.index}
```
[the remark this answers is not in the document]
csillag: I don't know, the ClickUp task had this image:
image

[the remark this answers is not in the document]
buberdds: I was referring to "Switch to another account" view. Maybe while importing accounts it makes sense, but we have derivation paths.

csillag:
> possible to generate fatal error while fetching Ledger accounts (fetch and click next)

should be fixed

> fatal error when opening account while balance is loading

should be fixed

> UI is flickering while loading balances

Can you show me what do you mean by flickering?

src/app/pages/OpenWalletPage/Features/ImportAccountsSelectionModal/index.tsx
```diff
@@ -8,28 +8,32 @@ import { ImportAccountsStepFormatter } from 'app/components/ImportAccountsStepFo
...
+  selectImportAccounts,
+  selectImportAccountsDisplayList,
+  selectSelectedAccounts,
+  selectImportAccountsPageNumber,
+} from 'app/state/importaccounts/selectors'
 import { ImportAccountsListAccount, ImportAccountsStep } from 'app/state/importaccounts/types'
 import { walletActions } from 'app/state/wallet'
-import { Box, Button, Spinner, Text } from 'grommet'
+import { Anchor, Box, Button, ResponsiveContext, Spinner, Text } from 'grommet'
+import { numberOfPages } from '../../../../state/importaccounts/saga'
```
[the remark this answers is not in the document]
csillag: OK, I fixed that. (btw the `../../state` was also present in other locations in the code; I fixed that, too.)

src/app/pages/OpenWalletPage/Features/ImportAccountsSelectionModal/index.tsx
```diff
@@ -64,25 +71,39 @@ export function ImportAccountsSelectionModal(props: ImportAccountsSelectionModal
...
+
+  const onPrev = () => {
+    dispatch(importAccountsActions.setPage(pageNum - 1))
+  }
+
+  const canGoNext = !!accounts.length && pageNum !== numberOfPages - 1
+
+  const onNext = () => {
+    dispatch(importAccountsActions.setPage(pageNum + 1))
+    if (props.type === 'ledger') {
```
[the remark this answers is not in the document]
csillag: Ah OK. I fixed that. (Along with several other locations of the code when the literal was present.)

src/app/state/importaccounts/saga.ts
```diff
@@ -59,6 +64,7 @@ function* enumerateAccountsFromMnemonic(action: PayloadAction<string>) {
     }
     yield* setStep(ImportAccountsStep.Done)
     yield* put(importAccountsActions.accountsListed(wallets))
+    yield* put(importAccountsActions.setPage(0))
```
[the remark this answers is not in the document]
csillag: The reason for this is that the saga that will fetch the balances is triggered by the `setPage` action.

src/app/components/Toolbar/Features/AccountSelector/index.tsx
```diff
@@ -78,9 +84,7 @@ export const Account = memo((props: AccountProps) => {
           <Box align="start" flex="grow" direction="row">
             {walletTypes[props.type]} {props.details && <Text size="small">({props.details})</Text>}
           </Box>
-          <Box>
-            <AmountFormatter amount={props.balance} />
-          </Box>
+          <Box>{props.balance !== undefined ? <AmountFormatter amount={props.balance} /> : <Spinner />}</Box>
```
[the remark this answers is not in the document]
csillag: Right. Fixed. (And we don't even need the `!`, just simpliy `balance`.)

src/app/state/importaccounts/index.ts
```diff
@@ -3,7 +3,15 @@ import { ErrorPayload } from 'types/errors'
...
 import { ImportAccountsListAccount, ImportAccountsState, ImportAccountsStep } from './types'
 
-export const initialState: ImportAccountsState = { accounts: [], showAccountsSelectionModal: false }
+export const initialState: ImportAccountsState = {
+  accounts: [],
+  showAccountsSelectionModal: false,
+  accountsSelectionPageNumber: 0,
+}
+
+interface ImportAccountChange extends Partial<ImportAccountsListAccount> {
```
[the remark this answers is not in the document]
csillag: The original idea was that in theory, we can change any part of the account... but yes, since the only usage is to inject the balance, we can simplify it a lot more.

src/app/state/importaccounts/index.ts
csillag: Yes, thanks, already fixed that.

src/app/state/importaccounts/saga.test.ts
```diff
@@ -21,59 +21,68 @@ describe('importAccounts Sagas', () => {
...
       return expectSaga(importAccountsSaga)
+        .withState({})
         .provide([
           [matchers.call.fn(TransportWebUSB.isSupported), true],
           [matchers.call.fn(TransportWebUSB.create), { close: () => {} }],
           [matchers.call.fn(Ledger.enumerateAccounts), [validAccount]],
           [matchers.call.fn(getBalance), {}],
         ])
-        .dispatch(importAccountsActions.enumerateAccountsFromLedger())
+        .dispatch(importAccountsActions.enumerateFirstAccountsFromLedger())
```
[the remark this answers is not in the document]
csillag: Updated.

src/locales/en/translation.json
```diff
@@ -209,7 +209,10 @@
     },
     "header": "How do you want to open your wallet?",
     "importAccounts": {
+      "next": "Next",
       "openWallets": "Open",
+      "openXWallets": "Open {{ number }} accounts",
```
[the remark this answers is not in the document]
csillag: keys changed.

src/app/state/importaccounts/saga.ts
csillag: .... right. Simplified that.

src/app/pages/OpenWalletPage/Features/ImportAccountsSelectionModal/index.tsx
```diff
@@ -95,10 +116,30 @@ export function ImportAccountsSelectionModal(props: ImportAccountsSelectionModal
             onClick={props.abort}
             disabled={cancelDisabled}
           />
+          {size === 'small' ? (
+            <Button onClick={onPrev} label={'<'} disabled={!canGoPrev} />
+          ) : (
+            <Anchor onClick={onPrev} disabled={!canGoPrev}>
```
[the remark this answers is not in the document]
csillag: I agree that using a button would be both technically easier and more semantically correct, but doesn't it look better as a link? (Again, the screenshot specified as a reference sketch showed a link.)

src/app/pages/OpenWalletPage/Features/ImportAccountsSelectionModal/__tests__/__snapshots__/index.test.tsx.snap
```diff
@@ -372,19 +386,46 @@ exports[`<ImportAccountsSelectionModal  /> should match snapshot 1`] = `
...
                 </button>
                 <div
+                  class="c10"
+                />
+                <button
                   class="c9"
+                  disabled=""
+                  type="button"
+                >
+                  &lt;
```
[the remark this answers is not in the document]
csillag: `a11yTitle` added.

src/app/pages/OpenWalletPage/Features/ImportAccountsSelectionModal/index.tsx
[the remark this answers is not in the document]
csillag: OK, for now I made it a button. We can easily re-style it later if needed.

buberdds:
> Can you show me what do you mean by flickering?
That's a minor issue, but height of an account row is changing and it looks like account list is "jumping" or flickering while going through different states. on init row height is X px, than with a loader it's Y px and with balance loaded height is different again. Maybe we need some fixed height on a wrapper element that holds spinner and label.

csillag:
> horizontal scroll bar in extension

Should be fixed, too

src/app/state/importaccounts/saga.ts
csillag: There are two entry points for Ledger-related actions:
 - When we start the process
 - When we move to a new page.

For those two situations, I think we do need two separate actions.

However, I can rearrange the functionality so that only one of the two actions is connected to a saga.
 - The first action (which is only triggered when we start the process) is only resetting everything
 - The second action gets the new accounts. (This is triggered both from the first action and when going to next page.)

I will do that now.

csillag: Done.

csillag:
Rebased on top of the Ledger permission fix.

> UI is flickering while loading balances

Fixed

src/app/state/importaccounts/saga.ts
csillag: So now there is only one saga for the ledger import, and only one trigger action for that saga. So what happens on the first page is exactly the same as what happens on the second page.

However, the drawback is that every time we want to start the ledger account import, we need to dispatch two actions side by side:
```
dispatch(importAccountsActions.resetLedgerAccountImport())
dispatch(importAccountsActions.enumerateAccountsFromLedger())
```
I think this is error a bit prone; it would probably be better to go back to having two actions (one for starting the import process, and another for continuing on the next page)

src/app/pages/OpenWalletPage/Features/ImportAccountsSelectionModal/index.tsx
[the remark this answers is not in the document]
csillag: I'm fine with either way. (My IDE adds relative paths by default, so that's a bit easier for me.)

src/app/state/importaccounts/saga.ts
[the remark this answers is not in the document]
csillag: ... but we can just as well trigger the balance loading directly, and then there is no need to call the `setPage` action.

src/app/state/importaccounts/saga.ts
csillag: I have moved back to having two different actions for the two different entry points.

csillag:
Further possible improvements:
 * Add new test cases that actually test the paging behavior
 * Make it so that the balance checks are executed in parallel with ledger account generation

buberdds:
Error occur
Open modal -> click next -> click on modal overlay (dark background)
Screenshot 2022-11-09 at 11 49 05

Open modal -> close modal immediately by clicking modal overlay
Screenshot 2022-11-09 at 11 50 32

src/app/state/importaccounts/saga.ts
```diff
@@ -71,33 +78,64 @@ function* enumerateAccountsFromMnemonic(action: PayloadAction<string>) {
...
   let transport: Transport | undefined
   try {
     transport = yield* getUSBTransport()
+    const existingAccounts = yield* select(selectImportAccountsFullList)
+    const start = existingAccounts.length
 
     yield* setStep(ImportAccountsStep.LoadingAccounts)
-    const accounts = yield* call(Ledger.enumerateAccounts, transport)
+    for (let index = start; index < start + accountsPerPage; index++) {
+      const a = (yield* call(Ledger.enumerateAccounts, transport, 1, index))[0]
```
buberdds, line 124: let's use meaningful var names

src/app/state/importaccounts/saga.ts
```diff
@@ -109,7 +147,11 @@ function* enumerateAccountsFromLedger() {
     yield* put(importAccountsActions.operationFailed(payload))
   } finally {
     if (transport) {
-      yield* call([transport, transport.close])
+      try {
+        yield* call([transport, transport.close])
+      } catch (error) {
+        console.log('Error while closing Ledger USB interface:', error)
```
buberdds, line 153: Shouldn't we throw  WalletError?

src/app/components/Toolbar/Features/AccountSelector/index.tsx
```diff
@@ -50,6 +51,10 @@ export const Account = memo((props: AccountProps) => {
     [WalletType.PrivateKey]: t('toolbar.wallets.type.privateKey', 'Private key'),
   }
 
+  const { details } = props
+
+  const index = details ? details.split('/').reverse()[0] : '?'
```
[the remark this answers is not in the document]
csillag: Nice. Thanks.

src/app/state/importaccounts/saga.ts
csillag: Originally I used account, but then I wanted to minimize my changes (compared to the original code) and when back to "a". Going back to "account" now.

src/app/state/importaccounts/saga.ts
csillag: No, I think not, because this happens when we are already trying to close things. Most often this is caused by some other event "closing" the connection before we do, like unplugging the USB. But by this time, we have already handled the error in the above code, this is just some cleanup.

csillag:
The latest change (changing the number of accounts per page) has introduced a cypress test fail. I'll fix that later.

src/app/state/importaccounts/saga.ts
```diff
@@ -71,33 +78,64 @@ function* enumerateAccountsFromMnemonic(action: PayloadAction<string>) {
...
+  }
+}
+
+/**
+ * Enumerate more accounts from Ledger, enough to fill up one page.
+ */
 function* enumerateAccountsFromLedger() {
+  const existingAccounts = yield* select(selectImportAccountsFullList)
+  const pageNumber = yield* select(selectImportAccountsPageNumber)
+  if (existingAccounts.length >= (pageNumber + 1) * accountsPerPage) return
```
lukaw3d, line 114: This seems odd. Add a comment

src/app/state/importaccounts/saga.ts
```diff
@@ -71,33 +78,64 @@ function* enumerateAccountsFromMnemonic(action: PayloadAction<string>) {
...
   let transport: Transport | undefined
   try {
     transport = yield* getUSBTransport()
+    const existingAccounts = yield* select(selectImportAccountsFullList)
+    const start = existingAccounts.length
 
     yield* setStep(ImportAccountsStep.LoadingAccounts)
-    const accounts = yield* call(Ledger.enumerateAccounts, transport)
+    for (let index = start; index < start + accountsPerPage; index++) {
+      const account = (yield* call(Ledger.enumerateAccounts, transport, 1, index))[0]
```
lukaw3d, line 124: This is odd. Refactor enumerateAccounts?

And add a ledger-specific accountsPerPage

src/app/state/importaccounts/saga.test.ts
```diff
@@ -84,7 +95,20 @@ describe('importAccounts Sagas', () => {
...
 
     it('should list accounts', async () => {
+      const expectedAccounts: ImportAccountsListAccount[] = []
+      for (let i = 0; i < accountsPerPage * numberOfAccountPages; i++) {
+        expectedAccounts.push({
+          address: mockAddress,
+          path: [44, 474, i],
+          privateKey: '00',
+          publicKey: '00',
+          selected: !i,
```
lukaw3d, line 105: i === 0

src/app/pages/OpenWalletPage/Features/ImportAccountsSelectionModal/index.tsx
```diff
@@ -78,29 +122,43 @@ export function ImportAccountsSelectionModal(props: ImportAccountsSelectionModal
...
+        <Box direction="row" gap="small" justify="between" pad={{ top: 'medium' }}>
+          <Button
+            disabled={!canGoPrev}
+            label={size === 'small' ? '<' : `< ${t('openWallet.importAccounts.prev', 'Prev')}`}
+            size={'small'}
+            a11yTitle={t('openWallet.importAccounts.prev', 'Prev')}
+            onClick={onPrev}
+          />
+          <Box>
+            Page {importAccounts.accountsSelectionPageNumber + 1} / {numberOfAccountPages}
```
lukaw3d, line 139: Translation

src/app/pages/OpenWalletPage/Features/ImportAccountsSelectionModal/index.tsx
```diff
@@ -44,32 +49,71 @@ function ImportAccountsSelector({ accounts }: ImportAccountsSelectorSelectorProp
...
+    case 1:
+      selectionLabel = t('openWallet.importAccounts.oneAccountSelected', 'One account selected')
+      break
+    default:
+      selectionLabel = t(
+        'openWallet.importAccounts.multipleAccountsSelected',
+        '{{ number }} accounts selected',
+        { number: selectedAccounts.length },
+      )
+  }
```
lukaw3d, line 106: This looks like reimplementation of plurals https://www.i18next.com/translation-function/plurals

src/app/components/Toolbar/Features/AccountSelector/index.tsx
```diff
@@ -50,6 +51,10 @@ export const Account = memo((props: AccountProps) => {
     [WalletType.PrivateKey]: t('toolbar.wallets.type.privateKey', 'Private key'),
   }
 
+  const { details } = props
```
lukaw3d, line 54: Slight preference against extracting `details` into a variable (we still use `({props.details})` below, and we aren't extracting other props)

src/app/components/Toolbar/Features/AccountSelector/index.tsx
```diff
@@ -50,6 +51,10 @@ export const Account = memo((props: AccountProps) => {
...
+
+  const index = details ? details.split('/').at(-1) : '?'
```
lukaw3d, line 56: I'm not a fan of the name `index` outside of loop iterators. accountNumber?

src/app/state/importaccounts/index.ts
```diff
@@ -3,41 +3,65 @@ import { ErrorPayload } from 'types/errors'
...
+  accountsSelectionPageNumber: 0,
+  step: ImportAccountsStep.Idle,
+}
 
 const slice = createSlice({
   name: 'importAccounts',
   initialState,
   reducers: {
-    clear(state, action: PayloadAction<void>) {
+    clear(state, _action: PayloadAction<void>) {
```
lukaw3d, line 17: You can remove the _action parameters :D

src/app/state/importaccounts/saga.ts
```diff
@@ -71,33 +78,64 @@ function* enumerateAccountsFromMnemonic(action: PayloadAction<string>) {
...
+    yield* setStep(ImportAccountsStep.Idle)
+  }
+}
+
+function* ensureAllBalancesArePresentOnCurrentPage() {
+  const existingAccounts = yield* select(selectImportAccountsOnCurrentPage)
+  const accountsWithNoBalance = existingAccounts.filter(a => !a.balance)
+  for (let i = 0; i < accountsWithNoBalance.length; i++) {
+    yield* fork(fetchBalanceForAccount, accountsWithNoBalance[i])
+  }
```
lukaw3d, line 105: Does this intentionally not wait? I'd expect `yield* all([ ... ])`

src/app/pages/OpenWalletPage/Features/ImportAccountsSelectionModal/index.tsx
```diff
@@ -44,32 +49,71 @@ function ImportAccountsSelector({ accounts }: ImportAccountsSelectorSelectorProp
...
+  const cancelDisabled = isBusyImporting && !error
+  const confirmDisabled = isBusyImporting || selectedAccounts.length === 0
+
+  const canGoPrev = !isBusyImporting && !!pageNum && !error
+
+  const onPrev = () => {
+    dispatch(importAccountsActions.setPage(pageNum - 1))
+  }
+
+  const canGoNext = !isBusyImporting && pageNum !== numberOfAccountPages - 1 && !error
```
lukaw3d, line 83:
```ts
const canGoPrev = pageNum > 0 && !isBusyImporting && !error
const canGoNext = pageNum < numberOfPages && !isBusyImporting && !error
```
would be more readable

lukaw3d:
> @buberdds Error occur Open modal -> click next -> click on modal overlay (dark background)
> 
> Open modal -> close modal immediately by clicking modal overlay

We should probably just disable clicking outside modal and <kbd>Esc</kbd>.
That would fix: https://github.com/oasisprotocol/oasis-wallet-web/issues/869

csillag:
Changing the number of accounts per page (from 5 to 4) has been separated in to a separate commit: https://github.com/oasisprotocol/oasis-wallet-web/pull/1124/commits/2d9c918df596d04c94c30d66f9ff67e78b658e3b

src/app/state/importaccounts/saga.ts
csillag: OK, added comment

src/app/state/importaccounts/saga.ts
csillag: > Refactor enumerateAccounts?

Refactored.

> And add a ledger-specific accountsPerPage

Why should it be ledger-specific? (Ledger is not the only limiting factor; we are constrained to 4 acc per page because of the extension sizing issues anyway.) Do we want to introduce different rules for all small screen / big screen / mnemonic / ledger combinations?

src/app/state/importaccounts/saga.test.ts
csillag: ok

src/app/components/Toolbar/Features/AccountSelector/index.tsx
csillag: OK, not extracting in any more. (In this function.)

src/app/components/Toolbar/Features/AccountSelector/index.tsx
csillag: OK, renamed.

src/app/state/importaccounts/index.ts
csillag: I have removed most of them, except this one: 
```
enumerateAccountsFromMnemonic(state, _action: PayloadAction<string>) 
```
... since keeping this is necessary to properly define the type of this action. (The string payload is used by the saga.)

src/app/state/importaccounts/saga.ts
csillag: Re-implemented using `yield* all()`. (The flow was the same both before and after the change.)

src/app/pages/OpenWalletPage/Features/ImportAccountsSelectionModal/index.tsx
csillag: OK, updated. (Except that we need `< numberOfPages - 1`.)

csillag:
> We should probably just disable clicking outside modal and Esc. That would fix: #869

We can do that in a separate PR, right?

lukaw3d:
> > We should probably just disable clicking outside modal and Esc. That would fix: #869
> 
> We can do that in a separate PR, right?

Yeah, just make sure edge cases from this PR don't remain

src/app/pages/OpenWalletPage/Features/ImportAccountsSelectionModal/index.tsx
lukaw3d: We don't have any existing plurals, and i18next-scanner doesn't seem to support it correctly via
```tsx
t('aa', {
  count: count,
  defaultValue_0: 'zero1',
  defaultValue_zero: 'zero2',
  defaultValue_1: 'one1',
  defaultValue_one: 'one2',
  defaultValue: 'plural1 {{count}}',
  defaultValue_plural: 'plural2 {{count}}',
  defaultValue_other: 'plural3 {{count}}',
})
```

I came up with this workaround:
```tsx
  // Workaround for i18next-scanner to pickup plurals correctly, because it is missing
  // defaultValue_zero, defaultValue_one, defaultValue_other here:
  // https://github.com/i18next/i18next-scanner/blob/4687b6a/src/parser.js#L502-L503
  t('aa_zero', 'zero')
  t('aa_one', 'one')

  return <div>{t('aa', 'count {{count}}', { count: count })}</div>
```

src/app/state/importaccounts/saga.ts
lukaw3d: I thought we might be increasing accountsPerPage for mnemonic at some point, and forget about ledger performance issues.

With new decrease, let's add a comment that accountsPerPage needs to be small enough for extension layout and small enough for slow ledgers

src/app/pages/OpenWalletPage/Features/ImportAccountsSelectionModal/index.tsx
csillag: OK, I have implemented it using this workaround.

src/app/state/importaccounts/saga.ts
csillag: OK, comment added.

lukaw3d (commented):
I've split the commits in a more sensible manner:
https://github.com/oasisprotocol/oasis-wallet-web/compare/lw/split-pagination

diff of changes: https://github.com/oasisprotocol/oasis-wallet-web/compare/fda9835..cb3f1e8

lukaw3d:
Updated previews

| Mnemonic | Ledger | Extension layout |
| --- | --- | --- |
| mnemonic  | ledger  | image |

buberdds (approved):
issues reported by me are fixed

csillag:
> I've split the commits in a more sensible manner:

Indeed, it does make more sense that way. (I have replaced the commits of this branch with that one.)

csillag:
> Could we hardcode the accounts window size while populating it with Ledger? Moving clickable buttons are PITA.

I think I have fixed that. (But I can't test with a Ledger right now.)

lukaw3d (approved):

PR #1143, opened 2022-11-11
Update for node v18

csillag:
In the current version of the code
 * The web wallet can be compiled with both node v16 and v18
 * The extension can only be built with v16; with v18, parcel hits [some errors](https://github.com/parcel-bundler/parcel/issues/8005).

This is an upstream bug that has been fixed since we locked our dependency versions.
The situation can be resolved by removing `yarn.lock` and thus updating all our dependencies.

This results in many changes.
One of the changes is that date formatting now uses a different kind of space; I updated the tests for that.

After this changes node v18 works OK (compilation, tests, everything), but **node v14 and v16 no longer works**.
(Actually, the code does work with node v16, but the tests don't, and I didn't feel like debugging it, since node v18 is already the current LTS release...)

So of we do this, everyone working on the wallet must upgrade to node v18.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -11,11 +11,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Use Node.js 14.x
+      - name: Use Node.js 18.x
         uses: actions/setup-node@v3
         id: yarn-cache
         with:
-          node-version: 14.x
+          node-version: 18
           cache: 'yarn'
       - if: steps.yarn-cache.outputs.cache-hit != 'true'
         run: yarn install --frozen-lockfile
@@ -25,10 +25,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Use Node.js 14.x
+      - name: Use Node.js 18.x
         uses: actions/setup-node@v3
         with:
-          node-version: 14.x
+          node-version: 18
           cache: 'yarn'
       - run: yarn install --frozen-lockfile
       - run: yarn lint --max-warnings 0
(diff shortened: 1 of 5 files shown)


csillag:
Please note that the failing tests are because the CI is still configured to use node v14, and our package json demands v18.

csillag:
As an aside, one could probably do a tighter update; ie. instead of removing the whole yarn.lock file, update only a few dependencies selectively... but it's a PITA to figure out exactly which ones, and eventually we need to upgrade all of them anyway.

buberdds:
Can you bump node version in GH actions? `.github` search for `14.x` ?

src/app/components/DateFormatter/__tests__/index.test.tsx
```diff
@@ -5,8 +5,8 @@ import { buildDatetime } from '../../../../../internals/getBuildData'
 
 describe('intlDateTimeFormat', () => {
   it('should format date to human readable format', () => {
-    expect(intlDateTimeFormat(1656932251000)).toEqual('7/4/2022, 10:57:31 AM')
-    expect(intlDateTimeFormat(new Date(Date.UTC(2022, 6, 27, 8, 33, 10)))).toEqual('7/27/2022, 8:33:10 AM')
+    expect(intlDateTimeFormat(1656932251000)).toEqual('7/4/2022, 10:57:31 AM')
+    expect(intlDateTimeFormat(new Date(Date.UTC(2022, 6, 27, 8, 33, 10)))).toEqual('7/27/2022, 8:33:10 AM')
```
lukaw3d, line 9: Perhaps replace with `\u202F` to make it more obvious

csillag: done.

csillag:
There is something wrong with the newer react dependencies...

csillag:
> There is something wrong with the newer react dependencies...

That has been solved, but now I have an issue whereas `Intl.DateTimeFormat` behaves differently locally and on GitHub. (Although node v18.12.x is running at both configurations)

csillag:
Now only the Cypress tests are dead...

buberdds:
Done via https://github.com/oasisprotocol/oasis-wallet-web/pull/1306

Issue #1149, opened 2022-11-14
No new line between commission bounds


matevz:
If the validator sets new commission bounds for future epochs, there is no new line between the existing and the upcoming commission bounds. Screenshot:

![image](https://user-images.githubusercontent.com/83672/201644488-9ff72bb5-8b1c-4cb8-9348-add190e28d66.png)

PR #1150, opened 2022-11-14
Split commission bound entries into rows

buberdds:
#1149

Screenshot 2022-11-14 at 13 48 11

### DIFF
--- a/src/app/pages/StakingPage/Features/CommissionBounds/index.tsx
+++ b/src/app/pages/StakingPage/Features/CommissionBounds/index.tsx
@@ -42,7 +42,7 @@ const CommissionBound = memo((props: CommissionBoundProps) => {
   )
 
   return (
-    <Box>
+    <Box as="li">
       <Text>{component}</Text>
     </Box>
   )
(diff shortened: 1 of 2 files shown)


src/app/pages/StakingPage/Features/CommissionBounds/index.tsx
```diff
@@ -61,7 +61,11 @@ export const CommissionBounds = memo((props: Props) => {
       .slice()
       .sort((a, b) => a.epochStart - b.epochStart)
       .map((b, i) => <CommissionBound bound={b} key={i} />)
-    return <>{items}</>
+    return (
+      <Box as="ul" margin="none" pad="none">
```
matevz, line 65: Is using `Box` the correct thing here or should this be `List` (https://v2.grommet.io/list)? Also not sure about setting margin and pad to `none` instead of using a proper class?

matevz (changes requested):
I can confirm this fixes the issue.

src/app/pages/StakingPage/Features/CommissionBounds/index.tsx
buberdds: I didn't use List component, because it's an overkill for such a case to render simple HTML markup like <ul><li> https://github.com/grommet/grommet/blob/master/src/js/components/List/List.js. It has build in InfiniteScroll etc.

Anyway, added a commit with List if that's our preference here. For no using classes, that's how Grommet API works.

matevz: Fair point. Feel free to use the original commit. Thanks.

matevz (approved):

PR #1154, opened 2022-11-14
Refactor inputs

lukaw3d:
Mostly extracted from https://github.com/oasisprotocol/oasis-wallet-web/pull/975

### DIFF
--- a/src/app/components/MnemonicField/index.tsx
+++ b/src/app/components/MnemonicField/index.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import { Box, FormField, TextArea } from 'grommet'
+import { FormField, TextArea } from 'grommet'
 
 interface Props {
   placeholder?: string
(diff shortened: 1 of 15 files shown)


src/app/components/MnemonicField/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  autoFocus?: boolean
+  value?: string
+  onChange?: (event: React.ChangeEvent<HTMLTextAreaElement>) => void
+  error?: string | false
+}
+
+export function MnemonicField(props: Props) {
+  return (
+    <Box border={false}>
```
buberdds, line 16: no need to use `border={false}`

src/app/components/MnemonicField/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <FormField htmlFor={props.inputElementId} error={props.error ? props.error : ''}>
+        <Box border={false}>
```
buberdds, line 18: no need to use border={false}. Why do you need these Box wrappers anyway?

buberdds (approved):

PR #1171, opened 2022-11-22
Prevent browsers from writing sensitive form inputs to user data

lukaw3d:
Add autocomplete=off to sensitive inputs, textareas, and parent forms (because Chrome doesn't support it on textareas).

Browsers write visible input fields from any website to user data to enable restoring sessions. Browsers exclude inputs with autocomplete="off" from cached form data in the session history (even though they ignore it and still offer autofill from saved passwords).

To ensure this fixes vulnerability:
- Manually checked
  grep --text t.e.s.t.M.n.e.m.o.n.i.c -r ~/.config/google-chrome
  grep --text testMnemonic -r ~/.mozilla/firefox

References:
- https://nvd.nist.gov/vuln/detail/CVE-2022-32969
- https://coinyuppie.com/slow-mist-a-brief-analysis-of-the-metamask-wallet-demonic-vulnerability/
- https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion
- https://nvd.nist.gov/vuln/detail/CVE-2022-0167

PR #1175, opened 2022-11-25
Test that browsers don't write sensitive form inputs to user data

lukaw3d:
E2E tests for https://github.com/oasisprotocol/oasis-wallet-web/pull/1171


src/app/lib/preventSavingInputsToUserData.ts
```diff
@@ -5,17 +5,28 @@
...
  * - https://nvd.nist.gov/vuln/detail/CVE-2022-0167
+ * - https://www.otto-js.com/news/article/spell-jacking-enhanced-spellcheck-features-send-pii-even-passwords
  */
-export const preventSavingInputsToUserData = { autoComplete: 'off' }
+export const preventSavingInputsToUserData = {
+  autoComplete: 'off',
+  autoCapitalize: 'off',
+  autoCorrect: 'off',
+  spellCheck: false,
+}
```
lukaw3d, line 32: Moved into https://github.com/oasisprotocol/oasis-wallet-web/pull/1181

pro-wh (approved):
good test

playwright/utils/expectShell.ts
```diff
@@ -0,0 +1,12 @@
+import { execSync } from 'child_process'
+import { expect } from '@playwright/test'
+
+export function expectShell(cmd: string) {
+  return expect.poll(() => {
```
pro-wh, line 5: actually does this `poll` make sense in the expect-it-to-say-not-found case? would it say exit code 1 right away? do we actually want to wait for something?

lukaw3d: Ah, annoyingly good point. I don't see a way to prevent flakiness and false positives here then.

I will make it one-shot. If expect-found becomes flaky after that, then I can add a delay or flush somehow.

PR #1180, opened 2022-11-29
Validate mnemonic words in MnemonicGrid

lukaw3d:
Quick improvement since https://github.com/oasisprotocol/oasis-wallet-web/pull/1181 disables spell-checking mnemonic

| Before 1181 | After 1181 + this |
| --- | --- |
|  localhost_3000_open-wallet_mnemonic (1) |  localhost_3000_open-wallet_mnemonic |


buberdds (approved):

pro-wh (approved):
oh good, this might help people who used a wrong word that's coincidentally a real word

did we support importing mnemonics from other languages? if so, those users will have to ignore this

lukaw3d:
:tada: we didn't support other languages; and bip39 now discourages them https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#wordlists

Fun fact: Japanese mnemonics don't use a normal space, so even our word splitting wouldn't work
image

Issue #1118, opened 2022-10-27
Display version tag in the footer when a tagged version is built


tjanez:
Currently, the build process isn't aware of versioned tags. E.g., even if the built is being made from `v1.3` tag, it will show:
![Screenshot from 2022-10-27 16-58-21](https://user-images.githubusercontent.com/2518544/198328005-1a26aa8f-a394-4e96-86f0-acfbbdd04a1a.png)

It would be nice if we made the build process check if we are currently on a versioned tag and display it appropriately. E.g.:
![Screenshot from 2022-10-27 17-04-18](https://user-images.githubusercontent.com/2518544/198328287-b518414f-21b8-4b32-a1f8-c11d7a8d713a.png)

lukaw3d:
Quick notes:
- if there isn't a tag, probably show the old version
- could use `git describe --exact-match --tags` to print `v1.3`
  or throw `fatal: no tag exactly matches '62737915364f1833e57f832676680ea537ef5320'`
- tag would link to e.g. https://github.com/oasisprotocol/oasis-wallet-web/releases/tag/v1.3

PR #1192, opened 2022-12-05
Add version tag in footer

lubej:
## Issue

Closes https://github.com/oasisprotocol/oasis-wallet-web/issues/1118

## Solution

Latest tag(version) should be displayed in the footer, linking to the matching tag in releases.

- `git describe --tags --abbrev=0` gets the latest single tag in commit history - possible to add `--exact-match`, to show only tags on current commit, but will break test builds;
- env variable `REACT_APP_BUILD_VERSION` added for `vx.x` tag;

## Caveats

`git describe --tags --abbrev=0` will throw an error, if there are no tags added - which is probably what we want anyway.


src/app/components/Footer/index.tsx
```diff
@@ -8,6 +8,8 @@ import { backend, BackendAPIs } from 'vendors/backend'
 import { MobileFooterNavigation, mobileFooterNavigationHeight } from '../MobileFooterNavigation'
 
 const githubLink = 'https://github.com/oasisprotocol/oasis-wallet-web/'
+const githubReleaseLink = (tag: string) =>
+  tag && tag !== 'v0.0' ? `${githubLink}releases/tag/${tag}` : `${githubLink}releases`
```
buberdds, line 12: I guess we don't need `tag &&` check if we expect to always have a string here. Other than that why do we care about `v0.0` value specifically? is `v0.1` valid for example?

src/app/components/Footer/index.tsx
```diff
@@ -54,27 +56,35 @@ export const Footer = memo(() => {
...
+        process.env.REACT_APP_BUILD_SHA &&
+        process.env.REACT_APP_BUILD_VERSION && (
+          <Text size="small" textAlign="center">
+            <Trans
+              i18nKey="footer.version"
+              t={t}
+              components={{
+                ReleaseLink: (
+                  <Anchor
+                    href={githubReleaseLink(process.env.REACT_APP_BUILD_VERSION)}
```
buberdds, line 69: Shouldn't we open external links in a new tab with `rel="noopener"` etc?

buberdds: Another issue not related to this PR, but are these links actually work in extension?

.github/workflows/build-test.yaml
```diff
@@ -25,6 +25,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
buberdds, line 29: Can you make commit message for this change more descriptive? What consequences this change has, does it impact build time?

src/app/components/Footer/index.tsx
lubej: The motivation behind this was, that `v0.0` was set as default in `.env`. So that we don't have a broken link, I just redirected to `https://github.com/oasisprotocol/oasis-wallet-web/releases`. Right now I just set the default version to the first tag in this repository, which is kind of awkward, since this code can't be moved.

.github/workflows/build-test.yaml
lubej: Right, what this line does is, it fetches all history for all branches and tags. The default `fetch-depth: 1` only fetches the first commit. So in order to get the latest tag, I need a more extensive history. I could limit it to few 100s, but I believe that this change shouldn't take more than a 1s.

src/app/components/Footer/index.tsx
lubej: One case to consider is to try out `noreferrer` in the wallet selected case. Where if the existing links are sending the referer, we might be actually exposing user's public key.

buberdds (approved):
Let's merge two last commits into one something like "Open footer links in a new tab".

optional: in terms of guards we could either check tag format in UI or maybe just add comment in gh action next to fetch-depth that it's needed to render footer correctly.

PR #1206, opened 2022-12-10
Cleanup

lukaw3d:


csillag:
Basically LGTM, just one question:

What's the deal with removing babel-plugin-istanbul, then adding it back, and then removing again?
(Can't we squash those changes?)

csillag (approved):
Reviewed. LGTM.

(I think we should clean up the commit history about the coverage plugin, but it's not that important.)

lukaw3d:
I thought babel-plugin-istanbul was unused. Locally I get the coverage even without it, but not on CI :shrug: will revert

PR #1253, opened 2023-01-19
Switch from babel-jest to ts-jest with isolatedModules

lukaw3d:


internals/jest/jest.base-config.js
```diff
@@ -1,7 +1,7 @@
 // @ts-check
 // https://jestjs.io/docs/configuration#defaults
 
-/** @type {import('@jest/types').Config.InitialOptions} */
+/** @type {import('ts-jest').JestConfigWithTsJest} */
```
buberdds, line 4: why we switch here to ts-jest and keep @jest/type in all three other configs?

buberdds (approved):

internals/jest/jest.base-config.js
lukaw3d: Ah, I should switch them all for consistency

PR #1321, opened 2023-03-15
Fix connecting to ledger by opening a normal window, not headless popup

lukaw3d:
Related to https://github.com/oasisprotocol/oasis-wallet-ext/pull/302 and https://github.com/oasisprotocol/oasis-wallet-ext/pull/301

| Before | After |
| --- | --- |
| before  |  after |

### DIFF
--- a/playwright/package.json
+++ b/playwright/package.json
@@ -5,6 +5,6 @@
     "test": "playwright test"
   },
   "devDependencies": {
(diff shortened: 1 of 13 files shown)


buberdds (approved):
Behaves a bit differently than # 301, but we can always adjust this later based on Don's feedback.

PR #1327, opened 2023-03-16
Increase default gas and fee amount to fix Cipher withdrawals

buberdds:
mirroring https://github.com/oasisprotocol/oasis-wallet-ext/pull/296


src/config.ts
```diff
@@ -154,5 +161,15 @@ export const paraTimesConfig: ParaTimesConfig = {
...
 
+export const getRuntimeConfig = (runtimeId: string) => {
+  const runtimeConfig = Object.values(paraTimesConfig).find(
+    config => config.mainnet.runtimeId === runtimeId || config.testnet.runtimeId === runtimeId,
+  )
+  if (!runtimeConfig) {
+    throw new Error(`Runtime config not found for runtimeId: ${runtimeId}`)
+  }
+  return runtimeConfig
+}
```
lukaw3d, line 172: This is yucky. We shouldn't copy it into the good codebase 

We should pass a fuller runtime configuration from
https://github.com/oasisprotocol/oasis-wallet-web/blob/eb4639f437b8c51364dc7b477c2c8d51b57d1851/src/app/state/paratimes/saga.ts#L75-L79
through
- submitParaTimeTransaction
- prepareParatimeTransfer
- OasisTransaction.buildParaTimeTransfer

lukaw3d (approved):

PR #1343, opened 2023-03-21
Refactor selecting network initially to avoid sync timing issues

lukaw3d:
Related to https://github.com/oasisprotocol/oasis-wallet-web/issues/792#issuecomment-1269273318

Reverts part of c125984a415bd1a3fd9dd61c9ce6f2634e8ef8eb and re-fixes https://github.com/oasisprotocol/oasis-wallet-web/pull/975#discussion_r1012328608

Instead of waiting for tabs to be synced before selecting network initially:
- initialize selecting network without syncing that action
- using takeLatest cancel initialization task if tabs sync during selectNetwork


buberdds (approved):
not sure why but you have to bump some snapshots

PR #1352, opened 2023-03-27
ParaTimes: cancel transfer instead of navigating back to ParaTime selection

buberdds:
Currently user can navigate back to ParaTime selector step which can lead to weird experience as we keep data for next steps.

sample: fill in Cipher recipient go back to the first step, select emerald, click next and we have Oasis addr in recipient input where we need eth addr.

### DIFF
--- a/src/app/pages/ParaTimesPage/TransactionRecipient/__tests__/index.test.tsx
+++ b/src/app/pages/ParaTimesPage/TransactionRecipient/__tests__/index.test.tsx
@@ -166,33 +166,17 @@ describe('<TransactionRecipient />', () => {
     expect(navigateToAmount).toHaveBeenCalled()
   })
 
-  it('should navigate back to deposit', async () => {
-    const navigateToDeposit = jest.fn()
-    jest.mocked(useParaTimesNavigation).mockReturnValue({
-      ...mockUseParaTimesNavigationResult,
-      navigateToDeposit,
-    })
-    render(<TransactionRecipient />)
-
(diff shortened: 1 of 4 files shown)


src/app/pages/ParaTimesPage/TransactionRecipient/__tests__/index.test.tsx
```diff
@@ -166,33 +166,17 @@ describe('<TransactionRecipient />', () => {
...
-      ...mockUseParaTimesNavigationResult,
-      navigateToWithdraw,
-    })
     render(<TransactionRecipient />)
 
-    await userEvent.click(screen.getByRole('button', { name: 'Previous' }))
+    await userEvent.click(screen.getByRole('button', { name: 'Cancel transfer' }))
 
-    expect(navigateToWithdraw).toHaveBeenCalled()
+    expect(clearTransactionForm).toHaveBeenCalled()
```
lukaw3d, line 180: could we instead have a test more like

> sample: fill in Cipher recipient go back to the first step, select emerald, click next and we have Oasis addr in recipient input where we need eth addr.

buberdds: That belongs to e2e test more. I can add one and comment it out temporary. Which one is now preferable in Wallet: cypress or playwright  ?

lukaw3d: playwright

lukaw3d (approved):

PR #1362, opened 2023-03-29
Add failing test for privateToEthAddress

lukaw3d:
Related to https://github.com/oasisprotocol/explorer/commit/dbdabe6c56bdfc0866e4a86908a65bda551c72c4

Apparently none of our existing jest tests touch multiple imports from `@ethereumjs/util` that check `instanceof Uint8Array` internally. Otherwise we would have odd failures. Added a skipped test as a reminder


csillag (approved):
LGTM

PR #1377, opened 2023-04-03
Workaround around Google Translate issue

buberdds:
https://github.com/oasisprotocol/oasis-wallet-web/issues/1354


buberdds:
That's not accurate. It is fixing fatal error, but some nodes won't be updated after async actions.

PR #1399, opened 2023-04-14
Reduce bundle size

lukaw3d:
Before:
``` sh
$ yarn build
...
build/index.8cfd98b9.js                               ⚠️  6.07 MB    16.71s
Done in 28.17s.
```

After:
``` sh
$ yarn build
...
build/index.073d82ed.js                               ⚠️  4.11 MB    11.88s
Done in 20.34s.
```


lukaw3d:
I also tried replacing `@oasisprotocol/client-rt` with `@oasisprotocol/client-rt/dist/*` but it isn't worth the hassle - it only saved ~50kb. And same for `@ethereumjs/util`.

buberdds (approved):

PR #1438, opened 2023-05-02
Refactor translations to reference button labels

lukaw3d:


matevz (approved):
Good idea. I tested this with the new Slovenian translation and works as expected.

Issue #1453, opened 2023-05-12
Importing mnemonic on phone freezes the UI for some seconds


lukaw3d:
Related to https://github.com/oasisprotocol/oasis-wallet-web/pull/1124

Code https://github.com/oasisprotocol/oasis-wallet-web/blob/85ec1b6d196dacf8b3a64f5bc05a12e528d0d1e8/src/app/state/importaccounts/saga.ts#L44-L71

When opening from mnemonic we pre-derive 48 accounts (so we can have similar paginating code between Ledger and mnemonic, and so we don't need to keep mnemonic in redux). This appears as if the button didn't work and UI is stuck for multiple seconds.

PR #1454, opened 2023-05-12
Prevent freezing UI rendering while pre-deriving accounts from mnemonic

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/1453

### DIFF
--- a/src/app/state/importaccounts/saga.ts
+++ b/src/app/state/importaccounts/saga.ts
@@ -4,7 +4,7 @@ import * as oasis from '@oasisprotocol/client'
 import { hex2uint, publicKeyToAddress, uint2hex } from 'app/lib/helpers'
 import { Ledger, LedgerSigner } from 'app/lib/ledger'
 import { OasisTransaction } from 'app/lib/transaction'
-import { all, call, fork, put, select, takeEvery } from 'typed-redux-saga'
+import { all, call, delay, fork, put, select, takeEvery } from 'typed-redux-saga'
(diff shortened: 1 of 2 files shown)


buberdds (approved):

src/app/state/importaccounts/saga.ts
```diff
@@ -66,6 +66,7 @@ function* enumerateAccountsFromMnemonic(action: PayloadAction<string>) {
 
   try {
     yield* setStep(ImportAccountsStep.LoadingAccounts)
+    // Pre-derive all pages of accounts so we don't need to store mnemonic in redux.
     for (let i = 0; i < accountsPerPage * numberOfAccountPages; i++) {
       const signer = yield* call(oasis.hdkey.HDKey.getAccountSigner, mnemonic, i)
```
pro-wh, line 71: does the scheme allow us to do the expensive part once instead of 48 times?

lukaw3d: I didn't think of that :D 

I currently doesn't https://github.com/oasisprotocol/oasis-sdk/blob/8039a75d7fe42dd8021317424cee073e60435f84/client-sdk/ts-web/core/src/hdkey.ts#L33-L35. it's fine

pro-wh: https://github.com/oasisprotocol/oasis-sdk/issues/1365 filed

Issue #1110, opened 2022-10-26
Show total balance from API in account selector


lukaw3d:
Currently we only show available balance from gRPC

https://github.com/oasisprotocol/oasis-wallet-web/blob/9719ee006b0fd95266a7f5c47329d3b3beb391c2/src/app/pages/OpenWalletPage/Features/ImportAccountsSelectionModal/index.tsx#L32

https://github.com/oasisprotocol/oasis-wallet-web/blob/550f447ff6ddaa213bff80c57db01f6573df5852/src/app/components/Toolbar/Features/AccountSelector/index.tsx#L106

csillag:
Unfortunately the problem is not merely not showing all the data; we don't even load the required data properly.
For example, look at this account:

http://localhost:3000/account/oasis1qpzregxdxcu83yj4zzvadrhysntunpe44u66s4e5

The balance is:

![image](https://user-images.githubusercontent.com/2093792/201243053-13aa2755-c69d-4a6a-8c34-5af01b9bc334.png)

However, when we load the balance for this account, we are running:

``` TypeScript
  const account = yield* call([nic, nic.stakingAccount], {
    height: 0,
    owner: short,
  })
  console.log('Loaded account', JSON.stringify(account, null, '  '))
  const balance = parseRpcBalance(account)
  console.log('loaded balance', JSON.stringify(balance, null, '  '))
  return balance

```

And this gets us:

```
Loaded account {
  "general": {
    "nonce": 1,
    "balance": {
      "0": 2,
      "1": 84,
      "2": 11,
      "3": 228,
      "4": 0
    }
  }
}
loaded balance {
  "available": "10000000000",
  "validator": {
    "escrow": "0",
    "escrow_debonding": "0"
  }
}
```

... so as we can see, the staked part of the balance is missing.

(The same thing happens with both `OasisScan` and `OasisMonitor`.)

lukaw3d:
gRPC doesn't have the needed information, but APIs do

https://api.oasisscan.com/testnet/chain/account/info/oasis1qpzregxdxcu83yj4zzvadrhysntunpe44u66s4e5
https://monitor.oasis.dev/api/testnet/data/accounts/oasis1qpzregxdxcu83yj4zzvadrhysntunpe44u66s4e5

A while ago I added parsing for this from API and measured how outdated the info is: https://github.com/oasisprotocol/oasis-wallet-web/commit/f1a2d867c93db0375f3167d8c08703d9670f853a
[...]

PR #1468, opened 2023-05-18
Show total balance in account selector

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/1110

### DIFF
--- a/src/app/lib/getAccountBalanceWithFallback.ts
+++ b/src/app/lib/getAccountBalanceWithFallback.ts
@@ -1,0 +1,36 @@
+import { addressToPublicKey, parseRpcBalance } from 'app/lib/helpers'
+import { call } from 'typed-redux-saga'
+import { getExplorerAPIs, getOasisNic } from '../state/network/saga'
+import { Account } from '../state/account/types'
+
+function* getBalanceGRPC(address: string) {
+  const nic = yield* call(getOasisNic)
+  const publicKey = yield* call(addressToPublicKey, address)
+  const account = yield* call([nic, nic.stakingAccount], { owner: publicKey, height: 0 })
+  const grpcBalance = parseRpcBalance(account)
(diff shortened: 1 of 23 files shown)


csillag (approved):
Reviewed, LGTM.

One question, though.

As far as I can tell, currently when the API call fails, and we fall back to the gRPC, this is all done silently. So just by looking at the UI, there is no way to tell whether we are seeing the real total balance, or just the portion returned by the gRPC fallback.

I don't like that. I would prefer somehow flagging the accounts where we couldn't get the full balance, and adding a visual indicator (maybe a question mark?) to the UI in this case.

(But this can be done in a separate PR)

src/app/lib/getAccountBalanceWithFallback.ts
```diff
@@ -0,0 +1,36 @@
...
+  const nic = yield* call(getOasisNic)
+  const publicKey = yield* call(addressToPublicKey, address)
+  const account = yield* call([nic, nic.stakingAccount], { owner: publicKey, height: 0 })
+  const grpcBalance = parseRpcBalance(account)
+  return {
+    address,
+    available: grpcBalance.available,
+    delegations: null,
+    debonding: null,
+    total: null,
```
lukaw3d, line 16: > @csillag: As far as I can tell, currently when the API call fails, and we fall back to the gRPC, this is all done silently. So just by looking at the UI, there is no way to tell whether we are seeing the real total balance, or just the portion returned by the gRPC fallback.

This fallback returns null total, rather than `available + 0 + 0` so the displayed balances shouldn't be misleading. But I noticed we don't differentiate loading and null. It just keeps loading
image

lukaw3d: Fixed 
image

PR #1569, opened 2023-08-04
Make playwright tests more reliably wait for concurrent events

lukaw3d:


lukaw3d:
Somehow this is less reliable on CI :shrug:

PR #1583, opened 2023-08-09
Add strict Permissions-Policy header

lukaw3d:
Currently we have `Permissions-Policy: usb=*` deployed (equal to nothing).


internals/getPermissionHeaders.js
```diff
@@ -53,4 +53,42 @@ const getCsp = ({ isExtension } = { isExtension: false }) =>
...
+    camera=("https://global.transak.com" "https://global-stg.transak.com"),
+    cross-origin-isolated=(),
+    display-capture=(),
+    document-domain=(),
+    encrypted-media=(),
+    execution-while-not-rendered=(),
+    execution-while-out-of-viewport=(),
+    fullscreen=(self "https://global.transak.com" "https://global-stg.transak.com"),
+    geolocation=(),
+    gyroscope=("https://api.sardine.ai"),
```
lukaw3d, line 72: This is annoying for maintenance: Sardine is a dependency of Transak. We won't know when Transak changes their dependencies, or if Sardine started requiring geolocation.

Should we loosen accelerometer,ambient-light-sensor,camera,encrypted-media,geolocation,gyroscope,magnetometer,microphone,publickey-credentials-get,web-share to `=*`?

lukaw3d: Although their docs say to put `allow="camera;microphone;fullscreen;payment"` on iframe, so it's already further restricted :shrug:

lubej (approved):

PR #1584, opened 2023-08-09
Test Validator icons pass CSP

lukaw3d:


playwright/tests/validators.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+  await expect(page.getByTestId('account-selector')).toBeVisible()
+  await page.getByRole('link', { name: 'Stake' }).click()
+  await expect(page.getByRole('heading', { name: 'Validators' })).toBeVisible()
+})
+
+test.describe('Validators', () => {
+  test('Content-Security-Policy should allow validator icons', async ({ page, baseURL }) => {
+    expect(baseURL).toBe('http://localhost:5000')
+    expect((await page.request.head('/')).headers()).toHaveProperty('content-security-policy')
+    await expect(page.getByText('Everstake')).toBeVisible()
```
lubej, line 33: This only checks the text, but not the icon itself? Or am I missing something?
Only way to check would be to take the screenshot.

lubej (approved):

playwright/tests/validators.spec.ts
lukaw3d: This waits for validator list to be loaded and `await expectNoErrorsInConsole(page)` above continuously checks there have been no CSP violations so far.

But I'll check if there's an explicit way to check "is image loaded" too!

doperiddle: Thank u

PR #1587, opened 2023-08-10
Test that generated mnemonics contain more than 32 bits of entropy

lukaw3d:


pro-wh:
image

image

pro-wh:
oh wait it's a playwright test

pro-wh:
image


>   31 skipped
  6 passed (18.9s)
Done in 20.11s.

oo it did not run nor flag a failure

lukaw3d:
uuuh something is wrong
image

lukaw3d:
It's an upstream playwright bug

doperiddle:
What is that?

Mark D Dannenbaum

lukaw3d:
Fixed that in https://github.com/oasisprotocol/oasis-wallet-web/pull/1588

pro-wh (approved):
not against this test, but we don't re-seed between each of these 250,000 trials. that mersenne twister implementation would have also generated as many unique mnemonics sequentially, wouldn't it?

lukaw3d:
I'm not sure / I don't have enough background. But this fails, so it probably re-seeds:
```ts
const a = require('@trustwallet/wallet-core'); /* 3.1.0 */
(async () => {
  const wasm = await a.initWasm()
  const generatedMnemonics = new Set([...new Array(250_000)].map(() => wasm.HDWallet.create(256, 'password').mnemonic()))
  console.log(generatedMnemonics.size)
  // 249995 (takes 30min)
})()
```

pro-wh:
https://github.com/trustwallet/wallet-core/blob/3.1.0/wasm/src/Random.cpp#L20

woa it does re-seed. not just for every call into wasm, but for every request for random bytes

PR #1591, opened 2023-08-16
Fix printing CSP: frame-ancestors should only be removed from <meta>

lukaw3d:
CSP in console has been missing frame-ancestors since 16e94b4.


buberdds (commented):
Why we need different CSP rules for CloudFront (main print task usage) and <meta>?

lukaw3d:
Added a comment, it's to fix a warning `'frame-ancestors' is ignored when delivered via <meta>`. Tests checking for CSP errors were picking it up.

We can probably just remove `<meta>` by now as well

buberdds (approved):

PR #1609, opened 2023-08-23
Add security headers to dev server

lukaw3d:


internals/getSecurityHeaders.js
```diff
@@ -10,9 +10,19 @@ const dappFrameAncestors = `
...
 `
+const localnet = `
+  http://localhost:42280
+  http://localhost:9001
+`
 const hmrWebsocket = `
   ws://localhost:2222
 `
+const reactErrorOverlay = `
+  'sha256-RV6I4HWPb71LvA27WVD3cEz8GsJrHlfcM/2X2Q5gV00='
```
buberdds, line 21: What generates this value? Is it going to be the same across react versions?

buberdds (approved):

internals/getSecurityHeaders.js
lukaw3d: It depends on part of the built version of react-error-overlay library (added by parcel). https://github.com/oasisprotocol/oasis-wallet-web/pull/1609/commits/b6fd0c580c2053bb8bf6faced6dbd82b5e78217f is supposed to make sure we notice if we have to change it after random dependency updates

I'll add this to make it more explicitly connected
```ts
expect((await page.request.head('/')).headers()['content-security-policy']).toContain(reactErrorOverlay)
```

lukaw3d: and I'm relying on browser to print the hash. I can't reproduce it
```js
const fs = require('fs')
const crypto = require('crypto')

const file = fs.readFileSync('./node_modules/react-error-overlay/lib/index.js', 'utf-8')
const script = file.slice(
  file.indexOf('/*! For license'),
  file.indexOf('iframeReady()}]);') + 'iframeReady()}]);'.length,
)
console.log(`${script.slice(0, 20)}...${script.slice(-20)}`)
console.log(crypto.createHash('sha256').update(script).digest('base64'))
// c1+FzHlBWznidUzzL/gYMDO8mS0XL4lAvO5NaizuNZs=
// expected RV6I4HWPb71LvA27WVD3cEz8GsJrHlfcM/2X2Q5gV00=
```

lukaw3d: Update: this would work. But it's still terribly ugly:
```js
const fs = require('fs')
const crypto = require('crypto')
const vm = require('vm')

const file = fs.readFileSync('./node_modules/react-error-overlay/lib/index.js', 'utf-8')

const strScript = file.slice(
  file.indexOf("'/*! For license"),
  file.indexOf("iframeReady()}]);'") + "iframeReady()}]);'".length,
)
const script = vm.runInNewContext(strScript)

console.log(`${script.slice(0, 20)}...${script.slice(-20)}`)
console.log(crypto.createHash('sha256').update(script).digest('base64'))
```

PR #1614, opened 2023-08-25
Before transaction validate if ledger device still derives same account

lukaw3d:
This improves the error message for users mixing up multiple ledger devices.

E.g.:
- user has two ledgers with different mnemonics
- imports account1 from ledger1 and saves it
- later
- plugs in ledger2 and tries to transfer from saved account1
- see error:

| Before | After |
| --- | --- |
| Screenshot from 2023-08-26 00-46-45  | Screenshot from 2023-08-26 00-46-50  |

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -135,6 +135,7 @@
     "reclaimedAmount": "Amount to reclaim"
   },
   "errors": {
+    "LedgerDerivedDifferentAccount": "This account does not belong to the currently connected Ledger.",
     "LedgerOasisAppIsNotOpen": "Oasis App on Ledger is closed.",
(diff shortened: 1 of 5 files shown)


buberdds (approved):

src/locales/en/translation.json
```diff
@@ -135,6 +135,7 @@
     "reclaimedAmount": "Amount to reclaim"
   },
   "errors": {
+    "LedgerDerivedDifferentAccount": "This Ledger did not derive this account.",
```
tjanez, line 138: In terms of the copy, would something like this be even clearer and easier to understand?

_This account does not belong to the currently connected Ledger._

tjanez (approved):
Otherwise looks good, great UX improvement!

PR #1621, opened 2023-08-31
Split main workflow into build, lint and test workflows

buberdds:

PR #1638, opened 2023-09-06
Upload build artifacts

buberdds:
Built on top of #1621

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -33,5 +33,17 @@ jobs:
           cache: yarn
       - name: Install dependencies
         run: yarn install --frozen-lockfile
+      - name: Set workflow variables
+        # Id is needed to access output in a next step.
+        id: vars
+        run: |
+          echo "SHORT_SHA=$(git rev-parse --short HEAD)" >> "$GITHUB_OUTPUT"
       - name: Build app
         run: yarn build
+      - name: Upload build artifacts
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-build.yml
```diff
@@ -33,5 +33,17 @@ jobs:
...
+        run: |
+          echo "SHORT_SHA=$(git rev-parse --short HEAD)" >> "$GITHUB_OUTPUT"
       - name: Build app
         run: yarn build
+      - name: Upload build artifacts
+        # Upload build artifacts on push event.
+        if: github.event_name == 'push'
+        uses: actions/upload-artifact@v3
+        with:
+          name: oasis-wallet-web-${{ steps.vars.outputs.SHORT_SHA }}
```
tjanez, line 48: Does it make sense to omit the `-web` part or will we build different artifacts for different form factors (i.e. web, browser extension and mobile app)?

buberdds: For now I will remove `-web` suffix. Long term it depends if we want to have extension build artifacts too (not sure about mobile)

buberdds: According to @lubej we can have mobile build artifacts too, but an issue might be signing with certificate.

tjanez: Ok, in this case, let's add the `-web` suffix.

tjanez (approved):

PR #1639, opened 2023-09-06
Setup gitlint

buberdds:
Built on top of #1621

### DIFF
--- a/.github/workflows/ci-lint.yml
+++ b/.github/workflows/ci-lint.yml
@@ -26,19 +26,42 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          # Checkout pull request HEAD commit instead of merge commit.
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-lint.yml
```diff
@@ -26,19 +26,42 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          # Checkout pull request HEAD commit instead of merge commit.
+          ref: ${{ github.event.pull_request.head.sha }}
+          # Fetch all history so gitlint can check the relevant commits.
+          fetch-depth: '0'
+      - name: Set up Python 3
+        uses: actions/setup-python@v4.2.0
```
tjanez, line 35: I suggest we only pin the major version (i.e. `v4`), similarly to how we pin other GitHub Actions' versions.

```suggestion
        uses: actions/setup-python@v4
```

tjanez (approved):

PR #1668, opened 2023-09-15
Make Contacts available only when profile exists

buberdds:
Option 3 from Slack thread

### DIFF
--- a/src/app/components/Toolbar/Features/Contacts/index.tsx
+++ b/src/app/components/Toolbar/Features/Contacts/index.tsx
@@ -1,37 +1,73 @@
-import { useState, useContext } from 'react'
+import { useState, useContext, ReactNode } from 'react'
 import { useSelector } from 'react-redux'
-import { useTranslation } from 'react-i18next'
+import { Trans, useTranslation } from 'react-i18next'
+import { useNavigate } from 'react-router-dom'
 import { Box } from 'grommet/es6/components/Box'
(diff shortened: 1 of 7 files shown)


src/locales/en/translation.json
```diff
@@ -397,6 +397,7 @@
       "emptyList": "You have no contacts yet.",
       "manage": "Manage Contact",
       "name": "Name",
+      "notAvailable": "To use Contacts please create a profile while opening a wallet.",
```
buberdds, line 400: Not confirmed yet, waiting for Don.

src/app/components/Toolbar/Features/Contacts/index.tsx
```diff
@@ -1,37 +1,53 @@
...
+  </Box>
+)
+
 export const Contacts = () => {
   const { t } = useTranslation()
   const [layerVisibility, setLayerVisibility] = useState(false)
   const contacts = useSelector(selectContactsList)
+  const isLockableOrCloseable = useSelector(selectIsLockableOrCloseable)
   const isMobile = useContext(ResponsiveContext) === 'small'
+  const isAvailable = isLockableOrCloseable === 'lockable'
```
lukaw3d, line 31: I would use the underlaying `selectUnlockedStatus; unlockedStatus === 'unlockedProfile'`
(like persist.createProfile.startPersisting)

lukaw3d (approved):

Issue #1673, opened 2023-09-19
Balance is not refreshed in AccountSelector and ParaTimes (after confirming deposit)


lukaw3d:
Start with 0.2 TEST balance, transfer in 0.1 TEST while ParaTimes are open:
- balance auto-refreshes
- balance under amount field is refreshed
![wallet dev oasis io_account_oasis1qpep5hp635asdzj5elkc360r55vfaah8sg443m9d_paratimes](https://github.com/oasisprotocol/oasis-wallet-web/assets/3758846/877fdc1c-d0ec-4b09-9b53-4feee2cf2281)
 ![wallet dev oasis io_account_oasis1qpep5hp635asdzj5elkc360r55vfaah8sg443m9d_paratimes (1)](https://github.com/oasisprotocol/oasis-wallet-web/assets/3758846/d8508f5c-f9c2-4b97-9be9-b977c457ae0f)

- confirming deposit fails
![wallet dev oasis io_account_oasis1qpep5hp635asdzj5elkc360r55vfaah8sg443m9d_paratimes (2)](https://github.com/oasisprotocol/oasis-wallet-web/assets/3758846/df5bb871-f351-4d79-941e-ab43e037829d)
![wallet dev oasis io_account_oasis1qpep5hp635asdzj5elkc360r55vfaah8sg443m9d_paratimes (3)](https://github.com/oasisprotocol/oasis-wallet-web/assets/3758846/4c562319-54f4-463f-9b10-7e0e0e20164a)

- opening account selector shows old 0.2 TEST balance
![wallet dev oasis io_account_oasis1qpep5hp635asdzj5elkc360r55vfaah8sg443m9d_paratimes (4)](https://github.com/oasisprotocol/oasis-wallet-web/assets/3758846/8a6d05b1-6d9a-4fa6-a2c4-3fef1e6f84ad)

lukaw3d:
Related to https://github.com/oasisprotocol/oasis-wallet-web/issues/1465

PR #1692, opened 2023-09-28
Fix continuously refreshing balances in wallet slice

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/1673

Fixes making transactions with the refreshed balance and displays new balance in AccountSelector.

### DIFF
--- a/playwright/tests/refreshing-balance.spec.ts
+++ b/playwright/tests/refreshing-balance.spec.ts
@@ -32,3 +32,14 @@ test('Account selector should refresh balances on network change', async ({ page
   await expect(page.getByTestId('account-choice')).toContainText('456.0')
   await page.getByRole('button', { name: /Select/ }).click()
 })
+
+test('Accounts page should continuously refresh balance', async ({ page }) => {
+  await page.getByTestId('account-selector').click()
+  await expect(page.getByTestId('account-balance-summary')).toContainText('123.0')
+  await expect(page.getByTestId('account-choice')).toContainText('123.0')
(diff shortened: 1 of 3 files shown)


playwright/tests/refreshing-balance.spec.ts
```diff
@@ -32,3 +32,14 @@ test('Account selector should refresh balances on network change', async ({ page
   await expect(page.getByTestId('account-choice')).toContainText('456.0')
   await page.getByRole('button', { name: /Select/ }).click()
 })
+
+test('Accounts page should continuously refresh balance', async ({ page }) => {
+  await page.getByTestId('account-selector').click()
+  await expect(page.getByTestId('account-balance-summary')).toContainText('123.0')
+  await expect(page.getByTestId('account-choice')).toContainText('123.0')
+  await mockApi(page, 456)
+  await page.waitForRequest('**/chain/account/info/*', { timeout: 60_000 })
```
lukaw3d, line 41: This is a slow test. ACCOUNT_REFETCHING_INTERVAL is 30 seconds. Not sure if we should change it for tests

lukaw3d: I changed the interval based on NODE_ENV, but suspiciously test still fails, exceeding timeout

lukaw3d: Changed to REACT_APP_E2E_TEST

buberdds (approved):

PR #1704, opened 2023-10-09
Export private key re-design

buberdds:


buberdds:
UI change is accepted by Don. 

I took a shortcut compared to an old extension and user does not have to provide a password and a key is revealed after user confirmation (button click) only. @lukaw3d  do you think it's Ok or we want to have input password as an additional "modal step" ?

src/app/components/Toolbar/Features/Account/ManageableAccountDetails.tsx
```diff
@@ -0,0 +1,162 @@
...
+                      }
+                      showTip={t('persist.loginToProfile.showPassword', 'Show password')}
+                      hideTip={t('persist.loginToProfile.hidePassword', 'Hide password')}
+                      width="auto"
+                    />
+                    <Box direction="row" justify="end" margin={{ top: 'medium' }}>
+                      <Button
+                        primary
+                        type="submit"
+                        label={t('toolbar.settings.exportPrivateKey.validate', 'Validate password')}
```
lukaw3d, line 126: I would replace this with "I understand, reveal my private key", and skip a step

src/app/components/Toolbar/Features/Account/ManageableAccountDetails.tsx
```diff
@@ -0,0 +1,162 @@
...
+    }
+  }
+  const hideLayer = () => {
+    dispatch(persistActions.resetPasswordCheckPass())
+    dispatch(persistActions.resetWrongPassword())
+    setPassword('')
+    setAcknowledge(false)
+    setLayerVisibility(false)
+  }
+  const onSubmit = () => dispatch(persistActions.checkPasswordAsync({ currentPassword: password }))
```
lukaw3d, line 53: Error if user without a profile wants to view their private key

src/app/components/Toolbar/Features/Account/ManageableAccountDetails.tsx
buberdds: You mean remove the initial "I understand, reveal my private key" step and just show pass input right away? Ok.

src/app/components/Toolbar/Features/Account/ManageableAccountDetails.tsx
buberdds: unpersisted has it's own flow now

src/app/state/persist/types.ts
```diff
@@ -36,6 +36,7 @@ export interface PersistState {
   loading: boolean
   stringifiedEncryptionKey: undefined | 'skipped' | StringifiedKeyWithSalt
   enteredWrongPassword: boolean
+  passwordCheckPass: boolean
```
lukaw3d, line 39: Is there any other pattern we can use?
I think it's quite ugly that this is in state

and it's risky
- open tab1, export private key, get asked to type in password, ignore
- open tab2, use a future feature that requires passwordCheckPass (or export private key of another account)
- type password in tab2
- tab1 unlocks

src/locales/en/translation.json
```diff
@@ -428,8 +428,13 @@
     "settings": {
       "cancel": "Cancel",
       "contacts": "Contacts",
-      "exportPrivateKey": "Export Private Key",
-      "lock": "Lock",
+      "exportPrivateKey": {
+        "confirm": "I understand, reveal my private key",
+        "copied": "Private key copied.",
+        "hint1": "The private key consists of a string of characters, and owning the private key is equivalent to owning the asset ownership.",
```
lukaw3d, line 434: "owning the asset ownership"
we need to improve this

buberdds: @donouwens, Can you please provide me an updated copy for this section? Current one was copied from an old extension.

src/app/state/persist/types.ts
buberdds: Not sure, I am open to suggestions how to respond to saga execution in UI differently.

I am missing some detail here. Actions are not synced across tabs why it would affect another tab?

lukaw3d: Oh, i wrote it down wrong, and there's only less serious issue
- open tab1, export private key, get asked to type in password, type pass, see private key
- open tab2, export private key for different account, see private key

buberdds: ahh thanks I can finally repro this :+1:

src/locales/en/translation.json
lukaw3d: I like 1.

src/app/state/persist/types.ts
lukaw3d: I still don't like the pattern enough.
Could you merge this without validating password commits, and move those to a separate PR for later?

lukaw3d (approved):

src/app/state/persist/types.ts
buberdds: additional stuff moved to https://github.com/oasisprotocol/oasis-wallet-web/compare/mz/privKeyPasswordCheck?expand=1

PR #1735, opened 2023-10-19
Fix contacts

lukaw3d:
Issue1:
- add two contacts with the same name
- select second one
- see address1 in input

Issue2:
- add contacts
- open wallet in new tab
- see zero contacts

### DIFF
--- a/src/app/pages/AccountPage/Features/SendTransaction/index.tsx
+++ b/src/app/pages/AccountPage/Features/SendTransaction/index.tsx
@@ -85,9 +85,11 @@ export function SendTransaction(props: SendTransactionProps) {
           >
             <TextInput
               id="recipient-id"
-              suggestions={contacts.map(contact => contact.name)}
+              suggestions={contacts.map(contact => ({ label: contact.name, value: contact.address }))}
               onSuggestionSelect={event =>
-                setRecipient(contacts.find(contact => contact.name === event.suggestion)?.address || '')
+                setRecipient(
+                  contacts.find(contact => contact.address === event.suggestion?.value)?.address || '',
+                )
               }
               name="recipient"
               value={recipient}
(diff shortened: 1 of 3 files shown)


src/app/state/persist/syncTabs.ts
```diff
@@ -31,6 +31,7 @@ export function receiveInitialTabSyncState(
     ...prevState,
     theme: initialSyncState.theme,
     wallet: initialSyncState.wallet,
+    contacts: initialSyncState.contacts,
```
buberdds, line 34: As far as I remember this is cover byPlaywright syncTabs. Why it's not failing?

src/app/pages/AccountPage/Features/SendTransaction/index.tsx
```diff
@@ -85,9 +85,11 @@ export function SendTransaction(props: SendTransactionProps) {
           >
             <TextInput
               id="recipient-id"
-              suggestions={contacts.map(contact => contact.name)}
+              suggestions={contacts.map(contact => ({ label: contact.name, value: contact.address }))}
               onSuggestionSelect={event =>
-                setRecipient(contacts.find(contact => contact.name === event.suggestion)?.address || '')
+                setRecipient(
+                  contacts.find(contact => contact.address === event.suggestion?.value)?.address || '',
```
buberdds, line 91: What's a value in adding contacts with the same names? This can be harmful imo. Shouldn't we block that behavior? cc @donouwens

buberdds: Anyway, if  we will need to change behavior I will create ClickUp task. For now I will merge this and make a release just in case we will have a deploy. Thanks Luka!

buberdds (approved):

src/app/state/persist/syncTabs.ts
lukaw3d: Expanded those tests so they would fail https://github.com/oasisprotocol/oasis-wallet-web/pull/1741

PR #1756, opened 2023-11-16
Change the password in test fixtures

lukaw3d:
I'm planning to use the same password for a test with state from old extension. And old extension had a stricter password policy.


buberdds:
something is off with tests `exit code 1: Command failed: grep -E...`

lukaw3d:
I don't know what's happening. I think firefox has some heuristic to scrub passwords from session file. I'll try with Abcd1234& instead

buberdds (approved):

PR #1769, opened 2023-11-22
Add Ionic Capacitor

lubej:
Folders android and ios were auto generated with npx cap add [platform]. [Android platform](https://capacitorjs.com/docs/android#adding-the-android-platform) - example for Android.

```
npx cap add android
npx cap add ios
```

Assets for each platform were generated using [@capacitor/assets](https://github.com/ionic-team/capacitor-assets#usage---custom-mode)


buberdds:
Before adding all these MBs to repo we should run squoosh or some other optimizer. 
sample splash png: 2.6MB -> (98% down) 38kB

package.json
```diff
@@ -57,7 +59,9 @@
   },
   "dependencies": {
     "@capacitor-community/bluetooth-le": "3.0.0",
+    "@capacitor/android": "5.0.4",
     "@capacitor/core": "5.0.4",
+    "@capacitor/ios": "5.0.4",
```
buberdds, line 64: can you bump to 5.6.0 ? would be cool to update core as well.

android/gradlew.bat
```diff
@@ -0,0 +1,92 @@
+@rem
```
buberdds, line 1: Do we need files for Win?

android/gradle/wrapper/gradle-wrapper.jar
buberdds, line 1: no idea what this compiled file contains, also no idea if that's common to store compiled Java in repo.

android/app/src/test/java/com/getcapacitor/myapp/ExampleUnitTest.java
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Example local unit test, which will execute on the development machine (host).
+ *
+ * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
+ */
+public class ExampleUnitTest {
+
+    @Test
+    public void addition_isCorrect() throws Exception {
+        assertEquals(4, 2 + 2);
```
buberdds, line 16: good that we have some test coverage here :trollface:

ios/App/Podfile.lock
```diff
@@ -0,0 +1,28 @@
...
+    :path: "../../node_modules/@capacitor/ios"
+
+SPEC CHECKSUMS:
+  Capacitor: d3d4463573438b9fa65326d1f3549da6f4c21634
+  CapacitorCommunityBluetoothLe: f2cab9f83b1ac5568c6450f39ad6c8553c64ef16
+  CapacitorCordova: b1fe6bf1f36974a8e4a9044b342d22d49c0996d6
+
+PODFILE CHECKSUM: 47998e176d73fb1dcd46c2cd81d1fab8a9f392d8
+
+COCOAPODS: 1.12.1
```
buberdds, line 28: Do we have readme with requirements (like cacaopods on Mac etc) how to setup and build all these files ?

android/app/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
+# You can control the set of applied configuration files using the
+# proguardFiles setting in build.gradle.
+#
+# For more details, see
+#   http://developer.android.com/guide/developing/tools/proguard.html
+
+# If your project uses WebView with JS, uncomment the following
```
buberdds, line 8: should we uncomment something here?

buberdds:
Can you provide steps how these files were generated - sorry I did not look into ionic since weeks and I remember nothing. thx

lubej:
> Can you provide requirements and steps how these files were generated - sorry I did not look into ionic since weeks and I remember nothing. thx

So all files inside `android` and `ios` folders are autogenerated. 

`ionic cordova platform add android` 
`ionic cordova platform rm ios`

And I wouldn't change those files, I know there is some useless stuff in it - addressing some comments in the PR.

android/app/proguard-rules.pro
lubej: No

ios/App/Podfile.lock
lubej: `ionic cordova build ios` all taken care of by Capacitor

android/gradlew.bat
lubej: I wouldn't touch any of autogenerated scripts. Just to spend ages fixing it, if something breaks in the future.

package.json
lubej: No, I expect before this gets merged, there will be at least 10 new version. I know this version works with current version of https://github.com/oasisprotocol/ionic-ledger-hw-transport-ble. And currently in this PR there is no real way to test on real devices, as Bluetooth UI is still missing - I was planning to add it after this PR is merged.

android/gradle/wrapper/gradle-wrapper.jar
lubej: I would say that it is there for a reason. And I wouldn't touch autogenerated files.

lukaw3d:
What was the reason for including these autogenerated files into the repo again? Is it because of unavoidable modifications we need to make to them later on?

lubej:
> What was the reason for including these autogenerated files into the repo again? Is it because of unavoidable modifications we need to make to them later on?

Yeah. Even though we could in theory just commit diffs, and overwrite generated files on `postinstall` script. But i am not sure if different version of ionic will generate the same files.

android/app/src/test/java/com/getcapacitor/myapp/ExampleUnitTest.java
lubej: I would still rather keep this, just in case there is no confusion if we need to generate android/ios platform from scratch, and wonder where did this file came from disappear. But if you are strongly against it, I can delete it as well.

lukaw3d (approved):

PR #1771, opened 2023-11-24
Add utils for persisted state migration from old extension

lukaw3d:
Functions for migrating state from https://github.com/oasisprotocol/oasis-wallet-ext


src/utils/walletExtensionV0.ts
```diff
@@ -0,0 +1,260 @@
...
+
+export type EncryptedData = EncryptedString<
+  // Array with exactly one element
+  [
+    {
+      mnemonic: EncryptedString<string> // string contains 12 or 24 words
+      currentAddress: string
+      accounts: Array<
+        | {
+            type: 'WALLET_INSIDE' // Mnemonic
```
lubej, line 22: Would be nice to have type as an enum, as it can easily be mistyped.

lubej (approved):
Overall looks good to me, would wait for @buberdds approval, as he is more familiar with the codebase.

buberdds (approved):

PR #1790, opened 2023-12-05
Restyle V0 extension migration UI

lukaw3d:
| Before | After |
| --- | --- |
| Migrate-from-V0-extension-persisted-state-to-valid-RootState-1-chromium-linux | Migrate-from-V0-extension-persisted-state-to-valid-RootState-1-chromium-linux |
| Migrate-from-V0-extension-persisted-state-to-valid-RootState-2-chromium-linux | Migrate-from-V0-extension-persisted-state-to-valid-RootState-2-chromium-linux |
| My-Accounts-tab-should-get-a-private-key-1-chromium-linux | My-Accounts-tab-should-get-a-private-key-1-chromium-linux |

### Mocks:
image

PR #1791, opened 2023-12-05
Add accessibility labels to menus

lukaw3d:


src/app/components/LanguageSelect/index.tsx
```diff
@@ -61,6 +61,7 @@ export const LanguageMenu = () => {
         }}
         items={languageOptions}
         fill
+        a11yTitle="Language"
```
buberdds, line 64: Why a label is not handled by i18n?

buberdds (approved):

src/app/components/LanguageSelect/index.tsx
buberdds: I have found a comment in the old PR "I think we want to keep this one international. Otherwise if someone gets into French, they can't get out"

lukaw3d: Yeah, we're inconsistent about it tho :shrug:

lukaw3d: Don says it's common to translate it

buberdds:
note: PR was merged with Change Log file only. Changes are in https://github.com/oasisprotocol/oasis-wallet-web/pull/1790

PR #1792, opened 2023-12-06
Update react-error-overlay CSP hash

lukaw3d:
Broken since 9c7da1ac4954089abee86f3a4900fb36f0d7401b

Note: react-error-overlay only appears during development, and csp-react-error-overlay.spec.ts doesn't test it on CI. Running it locally detects it as broken tho.

### DIFF
--- a/internals/getSecurityHeaders.js
+++ b/internals/getSecurityHeaders.js
@@ -17,6 +17,7 @@ const localnet = `
 const hmrWebsocket = `
   ws://localhost:2222
 `
+// If this changes csp-react-error-overlay.spec.ts will print a new sha in an error in csp-react-error-overlay.spec.ts.
 const reactErrorOverlay = `'sha256-yt+SNVxRkIi6H6yb7ndFuZM1esMX9esg3UpRHaTsyVk='`
 const hmrScripts = `
   'unsafe-eval'
--- a/internals/getSecurityHeaders.js
+++ b/internals/getSecurityHeaders.js
@@ -17,7 +17,7 @@ const localnet = `
 const hmrWebsocket = `
   ws://localhost:2222
 `
-const reactErrorOverlay = `'sha256-RV6I4HWPb71LvA27WVD3cEz8GsJrHlfcM/2X2Q5gV00='`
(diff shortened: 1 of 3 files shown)


internals/getSecurityHeaders.js
```diff
@@ -17,7 +17,7 @@ const localnet = `
 const hmrWebsocket = `
   ws://localhost:2222
 `
-const reactErrorOverlay = `'sha256-RV6I4HWPb71LvA27WVD3cEz8GsJrHlfcM/2X2Q5gV00='`
+const reactErrorOverlay = `'sha256-yt+SNVxRkIi6H6yb7ndFuZM1esMX9esg3UpRHaTsyVk='`
```
buberdds, line 20: Can you add a comment where we can find this ID so everyone is aware of it and also when it changes?

buberdds (approved):

internals/getSecurityHeaders.js
lukaw3d: Related: https://github.com/oasisprotocol/oasis-wallet-web/pull/1609#discussion_r1305548494

I copy it from chrome console, it's printed on any hot-module-reload when it's broken.

Modified the test so it would now print:
```js
  1) [chromium] › csp-react-error-overlay.spec.ts:5:5 › Dev Content-Security-Policy should allow react-error-overlay 

    Error: expect(received).not.toMatchObject(expected)

    Expected: not {"type": "error"}
    Received:     {"args": [], "location": {"columnNumber": 0, "lineNumber": 2769, "url": "http://localhost:3000/index.2b04b272.js"}, "text": "Refused to execute inline script because it violates the following Content Security Policy directive: \"script-src 'self' 'sha256-RV6I4HWPb71LvA27WVD3cEz8GsJrHlfcM/2X2Q5gV00=' 'unsafe-eval' 'report-sample'\". Either the 'unsafe-inline' keyword, a hash ('sha256-yt+SNVxRkIi6H6yb7ndFuZM1esMX9esg3UpRHaTsyVk='), or a nonce ('nonce-...') is required to enable inline execution.
    ", "type": "error"}

       at ../utils/expectNoErrorsInConsole.ts:37

      35 |         text: message.text(),
      36 |         type: message.type(),
    > 37 |       }).not.toMatchObject({
         |              ^
      38 |         type: 'error',
      39 |       })
      40 |     }

```

it's hard to make a prettier output

PR #1794, opened 2023-12-07
Force user to generate new mnemonic to reduce confusion when revisiting

lukaw3d:
This should make it clearer when reopening https://wallet.oasis.io/create-wallet that it's showing a new mnemonic

| Before | After |
| --- | --- |
|  localhost_3000_create-wallet (1) |  localhost_3000_create-wallet |


src/app/pages/CreateWalletPage/index.tsx
```diff
@@ -110,12 +107,12 @@ export function CreateWalletPage(props: CreateWalletProps) {
       )}
       <Grid gap="small" pad="small" columns={size === 'small' ? ['auto'] : ['2fr', '2fr']}>
         <Box background="background-front" style={blurMnemonicInFirefox}>
-          <MnemonicGrid mnemonic={mnemonic} />
+          <MnemonicGrid mnemonic={mnemonic.length > 0 ? mnemonic : '_'.repeat(24 - 1).split('_')} />
```
buberdds, line 110: I would replace repeat and split with  `new Array(24).fill('')`

PR #1808, opened 2023-12-18
Cleanup leftover for disabling sourcemaps in react-scripts

lukaw3d:


.env.production
```diff
@@ -1,5 +1,3 @@
-GENERATE_SOURCEMAP=false
```
buberdds, line 1: ah it's not used anymore

buberdds (approved):

PR #1809, opened 2023-12-19
Show number of invalid private keys in V0 extension migration step

lukaw3d:


src/locales/en/translation.json
```diff
@@ -242,7 +242,8 @@
   "migrateV0Extension": {
     "backupInvalidPrivateKeys": {
       "confirm": "I’ve safely stored my private keys",
-      "description": "Some of your private keys have typos and won’t be stored by the new wallet extension. Please make sure to copy them and store them elsewhere before proceeding.",
+      "description": "{{count}} of your private keys have typos and won’t be stored by the new wallet extension. Please make sure to copy them and store them elsewhere before proceeding, so you can correct them later.",
```
buberdds, line 245: here we want singular form (keys-key, them-it etc), right?

buberdds (approved):

src/locales/en/translation.json
lukaw3d: we can add that too :shrug: 1 sec

PR #1814, opened 2023-12-20
Explicitly mark each action and slice for syncing and persisting

lukaw3d:
This should detect: https://github.com/oasisprotocol/oasis-wallet-web/pull/1735/files#r1366593848

I think it was one of the main pitfalls with syncing and persisting


buberdds (approved):
test failures are not related to PR

PR #1817, opened 2023-12-21
i18n: Add German (62%) and Spanish (29%) partial translations

lukaw3d:


matevz (approved):
We can merge this and hopefully trigger German and Spanish users to improve the translation.

PR #1739, opened 2023-10-20
Add Bluetooth Ledger state

lubej:

PR #1841, opened 2024-02-07
Ionic Bluetooth UI

lubej:
## Solution

This PR is build upon https://github.com/oasisprotocol/oasis-wallet-web/pull/1739, https://github.com/oasisprotocol/oasis-wallet-web/pull/1784 and https://github.com/oasisprotocol/oasis-wallet-web/pull/1769. It adds additional steps, for user to choose either USB or BLE Ledger device. BLE Ledger device is currently only enabled and supported on mobile devices(through Ionic), this may be subject to change in the future.

## Links

Android test:
https://github.com/oasisprotocol/oasis-wallet-web/assets/9722540/6e0542f4-f92e-4788-aaba-ed9c6ba009cb


lubej:
Couple of things I noticed:
- "Opening Ledger through USB", needs to be replaced with either "Opening Ledger through Bluetooth", or just "Opening Ledger/Connecting with Ledger device";
- Modal height when listing Bluetooth device overflows for some reason;

lubej:
@donouwens mentioned really confusing wording on instructions upon connectioning Ledger device.

Should be for Bluetooth:

1. Connect your Ledger to this device using Bluetooth
2. Make sure your Ledger is paired with this device
3. Open the Oasis app on your Ledger

Should be for USB:

1. Connect your Ledger to this device via USB
2. Close the Ledger Live app on your computer
3. Open the Oasis app on your Ledger

src/app/state/importaccounts/saga.test.ts
```diff
@@ -84,6 +84,40 @@ describe('importAccounts Sagas', () => {
         .silentRun(50)
     })
 
+    it('should list ble devices', async () => {
```
buberdds, line 87: this test exists already / copy paste from L~53

src/app/state/importaccounts/saga.test.ts
```diff
@@ -84,6 +84,40 @@ describe('importAccounts Sagas', () => {
...
+        .provide([
+          [matchers.call.fn(BleTransport.isSupported), true],
+          [matchers.call.fn(BleTransport.list), bleDevices],
+        ])
+        .dispatch(importAccountsActions.enumerateDevicesFromBleLedger)
+        .put.like({ action: { payload: bleDevices } })
+        .silentRun(50)
+    })
+
+    it('should handle unsupported ble', async () => {
```
buberdds, line 112: this test exists already / copy paste L~78

src/app/pages/OpenWalletPage/Features/FromLedger/__tests__/index.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  useSelector: jest.fn(),
+  useDispatch: jest.fn(),
+}))
+
+jest.mock('react-router-dom', () => ({
+  ...jest.requireActual('react-router-dom'),
+  useNavigate: () => jest.fn(),
+}))
+
+jest.mock('react-router-dom', () => ({
```
buberdds, line 16: it is mocked already a few lines above

src/app/pages/OpenWalletPage/Features/FromLedger/__tests__/index.test.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from 'react'
```
buberdds, line 1: not needed

src/app/pages/OpenWalletPage/Features/FromLedger/__tests__/index.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { MemoryRouter } from 'react-router-dom'
+
+jest.mock('react-redux', () => ({
+  useSelector: jest.fn(),
+  useDispatch: jest.fn(),
+}))
+
+jest.mock('react-router-dom', () => ({
+  ...jest.requireActual('react-router-dom'),
+  useNavigate: () => jest.fn(),
```
buberdds, line 13: Why you are assigning a mock function here?

playwright/tests/ledger.spec.ts
```diff
@@ -6,7 +6,7 @@ test.describe('Ledger', () => {
     expect((await page.request.head('/')).headers()).toHaveProperty('permissions-policy')
     await expectNoErrorsInConsole(page)
 
-    await page.goto('/open-wallet/ledger')
```
buberdds, line 9: Playwright test should keep user flow. let's bring back this line -> find and click button -> expect /usb to be loaded

src/app/pages/OpenWalletPage/Features/FromLedger/__tests__/index.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+}))
+
+jest.mock('../../../../../lib/ledger', () => ({
+  ...jest.requireActual('../../../../../lib/ledger'),
+  // Throws BLE not supported
+  canAccessBle: () => jest.fn().mockReturnValue(false),
```
buberdds, line 24: this should mimic promise mockResolvedValue

src/app/pages/OpenWalletPage/Features/FromLedger/__tests__/index.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    await act(() => Promise.resolve())
+
+    expect(container).toMatchSnapshot()
+
+    expect(screen.getByText('errors.usbTransportNotSupported')).toBeInTheDocument()
+    expect(screen.getByText('errors.bluetoothTransportNotSupported')).toBeInTheDocument()
+  })
+
+  it('should render component with an ledger access button', async () => {
```
buberdds, line 46: This test is not needed you can move assertions to prev test imo because in current setup these buttons are always in a view. What would be useful is check their disable attr based on mocked value.

src/app/pages/OpenWalletPage/Features/FromLedger/__tests__/index.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+    renderComponent()
+
+    await act(() => Promise.resolve())
```
buberdds, line 49: act not needed in this test

src/app/pages/OpenWalletPage/Features/FromLedger/__tests__/index.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+    <MemoryRouter>
+      <FromLedger />
+    </MemoryRouter>,
+  )
+
+describe('<FromLedger />', () => {
+  it('should render component', async () => {
+    const { container } = renderComponent()
+
+    await act(() => Promise.resolve())
```
buberdds, line 38: Promise.resolve is useless here, but in general you don't want to use act in this case. If you want to test async changes use await waitFor(() => expect(...)

src/app/pages/OpenWalletPage/Features/FromLedger/__tests__/index.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    expect(container).toMatchSnapshot()
```
buberdds, line 40: I don't see any value in this snapshot. Doesn't matter what styles or i18n keys are. It's better to test when a label or button state changes.

src/app/pages/OpenWalletPage/Features/FromBleLedger/__tests__/index.test.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react'
+import { render } from '@testing-library/react'
+import { FromBleLedger } from '..'
+
+jest.mock('react-redux', () => ({
+  useSelector: jest.fn(),
+  useDispatch: jest.fn(),
+}))
+
+describe('<FromLedger />', () => {
```
buberdds, line 10: FromBleLedger

src/app/pages/OpenWalletPage/Features/FromBleLedger/__tests__/index.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+  it('should render component', () => {
+    const { container } = render(<FromBleLedger />)
+
+    expect(container).toMatchSnapshot()
```
buberdds, line 14: This test creates 2xx lines where non of a line cover logic and just checks i18n keys. If we are interested in labels we can clear PR by using screen API. What test should do is rather test conditions and redux actions if needed.

src/app/pages/OpenWalletPage/Features/FromBleLedger/__tests__/index.test.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react'
```
buberdds, line 1: not needed

buberdds (commented):
I have just looked at tests

android/app/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" xmlns:tools="http://schemas.android.com/tools">
```
buberdds, line 2: 1) Why it wasn't auto generated before when we use tools?
2) Can you add a note in tools that we are targeting Android 12 and above? not sure how clear  "s" is for all of us.

lubej: Not sure why it needs to be clear? This file likely won't change, and I don't see a benefit of adding more information to it - it is Android specific configuration.

Motivation for why this was added is here(plugin we are using for Bluetooth connection): https://github.com/capacitor-community/bluetooth-le?tab=readme-ov-file#optional-android-12-bluetooth-permissions 

And since the schema for `tools` were missing, and the build failed, I added the necessary schema.

src/app/pages/OpenWalletPage/Features/FromLedger/__tests__/index.test.tsx
lubej: Honestly don't remember anymore - this also goes for some other changes. Will need to look into it in more detail, what were the decisions behind the reasoning.

playwright/tests/ledger.spec.ts
lubej: Keeping as is, due to this error in CI:
image

And I don't want to implementy a hacky way to mock Bluetooth functionality in E2E tests.

extension/src/popup/routes.tsx
```diff
@@ -21,4 +21,8 @@ export const routes: RouteObject[] = [
     path: 'open-wallet/connect-device',
     element: <ConnectDevicePage />,
   },
+  {
+    path: 'open-wallet/ledger',
```
buberdds, line 25: Do we have this path defined in two places now? Here and in common routes? If that is true we have to move other one away from commons. Other option is to remove this entry and use common routes only, but move disabled logic to one of validation function of FormLedger

src/app/pages/OpenWalletPage/Features/FromLedger/__tests__/index.test.tsx
buberdds: I think it's because you are wrapping jest.fn() with arrow function 
` canAccessBle: jest.fn().mockResolvedValue(false),` should work

src/app/pages/OpenWalletPage/Features/FromLedger/__tests__/index.test.tsx
```diff
@@ -1,16 +1,43 @@
...
+    </MemoryRouter>,
+  )
+
 describe('<FromLedger />', () => {
-  it('should render component', () => {
-    const { container } = render(<FromLedger />)
+  it('should render component in disabled state', async () => {
+    renderComponent()
+
+    await waitFor(async () => {
```
buberdds, line 27: async and two awaits below not needed

src/app/pages/OpenWalletPage/webextension.tsx
```diff
@@ -2,9 +2,16 @@ import React from 'react'
...
 export function OpenWalletPageWebExtension() {
   const href = useHref('connect-device')
 
   return <SelectOpenMethod webExtensionLedgerAccess={() => openLedgerAccessPopup(href)} />
 }
+
+export function FromLedgerWebExtension() {
+  const href = useHref('/open-wallet/connect-device')
+
+  return <FromLedger webExtensionLedgerAccess={() => openLedgerAccessPopup(href)} disableBluetoothLedger />
```
buberdds, line 16: not sure we need this `FromLedgerWebExtension` wrapper and `disableBluetoothLedger` prop in `FromLedger`
it can be replaced by runtime check in a component or whatever function is responsible for validating access to ble 
```
import { runtimeIs } from 'config'
const isExtension = runtimeIs === 'extension'
```

src/app/pages/OpenWalletPage/Features/FromUsbLedger/index.tsx
```diff
@@ -0,0 +1,56 @@
+import { importAccountsActions } from 'app/state/importaccounts'
+import { Box } from 'grommet/es6/components/Box'
+import { Button } from 'grommet/es6/components/Button'
+import { Heading } from 'grommet/es6/components/Heading'
+import React from 'react'
```
buberdds, line 5: import not needed

src/app/pages/OpenWalletPage/Features/FromLedger/index.tsx
```diff
@@ -1,56 +1,93 @@
...
 import { useTranslation } from 'react-i18next'
-import { useDispatch, useSelector } from 'react-redux'
-import { ImportAccountsSelectionModal } from 'app/pages/OpenWalletPage/Features/ImportAccountsSelectionModal'
-import { selectShowAccountsSelectionModal } from 'app/state/importaccounts/selectors'
-import { Header } from 'app/components/Header'
-import { WalletType } from 'app/state/wallet/types'
 
-export function FromLedger() {
+type SelectOpenMethodProps = {
+  webExtensionLedgerAccess?: () => void
```
buberdds, line 11: this name could be adjusted as it's logic is bound to usb

src/app/pages/OpenWalletPage/Features/ListBleLedgerDevicesModal/index.tsx
```diff
@@ -0,0 +1,131 @@
+import React from 'react'
```
buberdds, line 1: not needed

src/app/pages/OpenWalletPage/Features/ListBleLedgerDevicesModal/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const error = useSelector(selectError)
+  const step = useSelector(selectImportAccountsStep)
+
+  const isBusyImporting = step !== ImportAccountsStep.Idle
+  const cancelDisabled = isBusyImporting && !error
+  const confirmDisabled = isBusyImporting || !selectedBleDevice
+
+  return (
+    <ResponsiveLayer onEsc={props.abort} onClickOutside={props.abort} modal background="background-front">
+      <Box width="800px" pad="medium">
```
buberdds, line 82: Is it responsive with this value?

src/app/pages/OpenWalletPage/Features/ListBleLedgerDevicesModal/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+    </Box>
+  )
+}
+
+interface ListBleLedgerDevicesModalProps {
+  abort: () => void
+  next: () => void
+}
+
+export function ListBleLedgerDevicesModal(props: ListBleLedgerDevicesModalProps) {
```
buberdds, line 69: so the difference is that we use native modal for usb and custom one for ble?

buberdds:
There is one bug compared to master branch which affect extension

master
Ext Open Wallet -> Grant access -> Connect Ledger device -> Connect device -> Modal success -> Popup reloaded with Ledger accounts

PR: 
Ext Open Wallet -> Grant access -> Connect Ledger device -> Connect device -> Modal success -> Popup not reloaded

src/app/pages/OpenWalletPage/Features/ListBleLedgerDevicesModal/index.tsx
lubej: Yes, works as expected.

src/app/pages/OpenWalletPage/Features/ListBleLedgerDevicesModal/index.tsx
lubej: Yes, sort of. Listing is not native, connect to a single device is. This will probably be reworked for desktop support - but the API is still not officially released yet(at least the last time I checked).

extension/src/popup/routes.tsx
lubej: Did try to refactor this, but seems to brake a lot of things 😓 The way I expect this to work is that it overrides the path from the common route(last entry over existing one).

The way I see it, it is a clear separation, as web extension has different entry point. And it original idea, of adding routes, should stay the same?

@lukaw3d can you chip in here?

lubej:
> There is one bug compared to master branch which affects extension
> 
> master Ext Open Wallet -> Grant access -> Connect Ledger device -> Connect device -> Modal success -> Popup reloaded with Ledger accounts
> 
> PR: Ext Open Wallet -> Grant access -> Connect Ledger device -> Connect device -> Modal success -> Popup not reloaded

Can you try again? And make a video of it. As I was unable to reproduce.

buberdds:
master 
connect Ledger
Screenshot from 2024-04-26 08-04-04

list Ledger accounts
Screenshot from 2024-04-26 08-04-31

side branch
- nothing happens in popup after successful connection
Screenshot from 2024-04-26 08-05-03

make sure to to turn off / turn on extension while switching branches

extension/src/popup/routes.tsx
buberdds: How it can brake things? `'open-wallet/ledger'` defined in both routes.tsx files not in commons, or keeping route only in commons. Now with removed disabled"Somtehing" prop we go with commons approach. It means this entry here and FromLedgerWebExtension is not needed anymore, right?

src/app/pages/OpenWalletPage/Features/FromLedger/__tests__/index.test.tsx
buberdds: This minor issue got lost in comments too

extension/src/popup/routes.tsx
lubej: Ah I see now what you mean. The way I tried it was something like this:

```typescript
export function FromLedgerDynamic() {
  const href = useHref('/open-wallet/connect-device')
  const isExtension = runtimeIs === 'extension'

  const props = isExtension ? { webExtensionUSBLedgerAccess: () => openLedgerAccessPopup(href) } : {}

  return <FromLedger {...props} />
}
```

And import it in common only. And almost all unit tests started failing.

src/app/pages/OpenWalletPage/Features/FromLedger/__tests__/index.test.tsx
lubej: Right 👍 fixed

lubej:
> There is one bug compared to master branch which affects extension
> 
> master Ext Open Wallet -> Grant access -> Connect Ledger device -> Connect device -> Modal success -> Popup reloaded with Ledger accounts
> 
> PR: Ext Open Wallet -> Grant access -> Connect Ledger device -> Connect device -> Modal success -> Popup not reloaded

@buberdds this has been fixed now, even though the fix is not really sophisticated. The issue was, that the enumerating modal is not available on the `FromLedger` component. I know that there was a discussion to show both options on Extension as well (Bluetooth & USB) which breaks the functionality. We can either directly redirect to USB Ledger screen, which may cause problems in the future, if the Bluetooth functionality is ever added to desktop. Make the USB Ledger enumeration modal global. Or leave it this way, with redirect to USB Ledger. Which after writing everything down, seems to me like a good enough option.

buberdds:
I am still having issues with using this:

1) desktop 
I hit a case where nothing happens. So if it's only working with capacitor why we are enabling it on desktop?

Screenshot from 2024-04-30 08-04-12

depends on browser or browser's settings we are not showing hint about disabled Ledger button anymore (this affects mobile too when user has BLE disabled)

Screenshot from 2024-04-30 09-13-33

disabled BLE button label - my laptop has BLE enabled, so this message fits mobile devices only   

Screenshot from 2024-04-30 09-17-33

2) extension 
Ledger account list popup has extra scrollbars, which was fixed recently and was one of a blocker for a release

Screenshot from 2024-04-30 08-34-18

mew extension popup Ledger options do not make sense - grant Ledger access and disabled Ledger button next to it  

Screenshot from 2024-04-30 08-06-38

- mobile  (apk installed on mobile device)
not longer able to list Ledger devices ?

Screenshot_20240430-093031
Screenshot_20240430-092953

lubej:
@buberdds 
> Desktop - I hit a case where nothing happens. So if it's only working with capacitor why we are enabling it on desktop?

Yes, so currently there is no official support for desktop - due to Web Bluetooth API still being in experimental mode. This has been changed now. Previously it was only disabled in extension. Now it checks if the app is running on native Capacitor platform(Android or iOS)

> Desktop - depends on browser or browser's settings we are not showing hint about disabled Ledger button anymore (this affects mobile too when user has BLE disabled)

After our sync, I dropped the check. So now Ledger button is always enabled on this screen. And the unsupported messages will appear in the next screen. Also updated Bluetooth and WebUSB messages, to make it more generic

> Extension - Ledger account list popup has extra scrollbars, which was fixed recently and was one of a blocker for a release

After rebase I was unable to reproduce, not sure if this was fixed in the meantime. Can you confirm it works for you as well?

> Extension - new extension popup Ledger options do not make sense - grant Ledger access and disabled Ledger button next to it

As per our sync discussion, we want to make it consistent. Also, there might be Bluetooth support coming to extension, when Web Bluetooth API is no longer in experimental mode.

> Mobile (apk installed on mobile device) - no longer able to list Ledger devices?

I was unable to reproduce. Did your Ledger maybe run out of battery? Or something similar? I see that you have your Ledger paired. Can you try again, and write the exact steps to reproduce?

buberdds:
Extension Ledger views are broken

Screenshot from 2024-05-08 09-27-25
Screenshot from 2024-05-08 09-27-03

buberdds (approved):
will add clickup task to explore modal unification (usb /ble )

PR #1845, opened 2024-02-12
Include security headers in the build and release

lukaw3d:


buberdds (approved):
Is this meant to be used by devops somewhere in TF scripts?

lukaw3d:
Possibly

But also to make releases actually self contained. When trying to deploy before:
- you had to download the release zip
- find the corresponding `release` job, and read the `build` output to find the headers in stdout
  (or risk inconsistency with checking out the released commit and run build locally)

PR #1848, opened 2024-02-12
Update release notes

lukaw3d:


docs/release-process.md
```diff
@@ -2,6 +2,18 @@
...
+- ```sh
+  pip3 install --upgrade \
+    https://github.com/oasisprotocol/towncrier/archive/oasis-master.tar.gz
+  ```
+
+- `yarn changelog`, commit "Assemble changes for ___ release", and merge PR
+- `yarn release-tag`
+  ([.github/workflows/release.yml](/.github/workflows/release.yml) then creates
+  a github release)
+
```
lukaw3d, line 16: My *actual* steps were:
- yarn changelog, commit, merge
- yarn release-tag
- edit release to mark it pre-release
- ask to deploy the pre-release to staging
- ask to deploy to production
- edit release to mark it latest release

Am I doing it wrong and is `yarn release-tag` supposed to come after deployment?

buberdds: I think you are doing it correctly. `release-tag` triggers release workflow which is needed for deployment. If pre-release is your default flow you prob want to add [prerelease](https://github.com/softprops/action-gh-release?tab=readme-ov-file#-customizing) to https://github.com/oasisprotocol/oasis-wallet-web/blob/master/.github/workflows/release.yml

docs/release-process.md
```diff
@@ -2,6 +2,18 @@
 
 ## Web wallet
 
+Make a release:
+
+- ```sh
```
buberdds, line 7: id this shell cmd needed in a flow ?

buberdds (approved):

docs/release-process.md
tjanez: Luka, in your flow, you are using the criterion of "is it deployed to production at wallet.oasis.io" as a criterion for release (vs. pre-release).

It will be a bit more complex soon when we also build the browser extension from the same code base.
Then the question will become: Is the criteria for the release "is it deployed to production at wallet.oasis.io" AND "is it published at chrome web store"?
Or is it better to untie release/pre-release from these deployment/publish criteria and simply mark the latest version tag as the latest release?

docs/release-process.md
lukaw3d: It's different from the official towncrier, so i think its good to point it out whereever its used. Would be better if changelog.js checked if correct towncrier exists and printed "pip install", but that affects multiple repos

docs/release-process.md
lukaw3d: Annoying :/ if we untie: not ideal if we mark a release while we're still checking staging

PR #1857, opened 2024-03-27
Show ticker hint in inputs

lukaw3d:
| Before | After |
| --- | --- |
| image  |  image |

### DIFF
--- a/.changelog/1857.trivial.md
+++ b/.changelog/1857.trivial.md
@@ -1,0 +1,1 @@
+Show ticker hint in inputs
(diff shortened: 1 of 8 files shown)


buberdds (approved):
snapshots are failing, would be cool to see a mocked ticker in at least one snapshot (if that does not require test refactors)

ideas:
- increase right padding in inputs so there is more space between ticker and native controls 
Screenshot from 2024-03-27 17-47-16
- use default text color for tickers. Currently it looks like a link especially in a light theme

lukaw3d:
Annoying:
- Grommet input has a fixed space available, so padding didn't help. I reduced the font size instead.
- Switched to grey color for now, to match other tickers. Might switch all tickers to blue/green in the future
image

PR #1869, opened 2024-04-04
Update profile deletion text

lukaw3d:
| Before | After |
| --- | --- |
| localhost_3000_ (2) | localhost_3000_ |
| localhost_3000_ (3) | localhost_3000_ (1) |

Don agrees

### DIFF
--- a/src/app/components/Persist/DeleteProfileButton.tsx
+++ b/src/app/components/Persist/DeleteProfileButton.tsx
@@ -12,9 +12,10 @@ import { Header } from 'app/components/Header'
 
 interface DeleteProfileButtonProps {
   prominent?: boolean
+  variant: 'voluntary' | 'forgot-password'
 }
 
-export function DeleteProfileButton({ prominent }: DeleteProfileButtonProps) {
+export function DeleteProfileButton({ prominent, variant }: DeleteProfileButtonProps) {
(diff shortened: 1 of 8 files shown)


src/app/components/Persist/DeleteProfileButton.tsx
```diff
@@ -33,7 +33,7 @@ export function DeleteProfileButton({ prominent }: DeleteProfileButtonProps) {
     <>
       <Button
         color="status-error"
-        label={t('persist.loginToProfile.deleteProfile.button', 'Delete profile')}
+        label={t('persist.loginToProfile.deleteProfile.button', 'Forgot password?')}
```
buberdds, line 36: Can we take into account the second context where button is used? When user has an open profile we will show "Forget password?" in Profile tab.
Screenshot from 2024-04-05 09-06-38

buberdds (approved):

src/app/components/Persist/DeleteProfileButton.tsx
lukaw3d: ah :/

lukaw3d: Checking this variant with Don: 
image

PR #1870, opened 2024-04-05
Rename to ROSE Wallet

lukaw3d:
| Before | After |
| --- | --- |
| localhost_3000_account_oasis1qz78ap0456g2rk7j6rmtvasc9v2kjhz2s58qgj90(extension360x600) | localhost_3000_(extension360x600) |
| localhost_3000_account_oasis1qz78ap0456g2rk7j6rmtvasc9v2kjhz2s58qgj90 (1) | localhost_3000_ |
| localhost_3000_account_oasis1qz78ap0456g2rk7j6rmtvasc9v2kjhz2s58qgj90 | localhost_3000_ (1) |

### DIFF
--- a/.changelog/1870.breaking.md
+++ b/.changelog/1870.breaking.md
@@ -1,0 +1,1 @@
+Rename to ROSE Wallet
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -45,5 +45,5 @@ jobs:
         if: github.event_name == 'push'
         uses: actions/upload-artifact@v4
         with:
-          name: oasis-wallet-web-${{ steps.vars.outputs.SHORT_SHA }}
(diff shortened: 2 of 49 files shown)


lukaw3d:
cc @donouwens

lubej (approved):
I am just giving LGTM, but would rather see if @buberdds looks at it as well, as he is more familiar with the project.

Mobile part looks good so far, need to check if anything breaks(it shouldn't), but we need to replace the icon + splash as it contains references to the old naming. @donouwens

lukaw3d:
> but we need to replace the icon + splash as it contains references to the old naming

Good point, I wouldn't notice.

I'll rather merge now and fix later - deadline

lubej:
> > but we need to replace the icon + splash as it contains references to the old naming
> 
> Good point, I wouldn't notice.
> 
> I'll rather merge now and fix later - deadline

Sure, no problem. It is not live anyway.

PR #1879, opened 2024-04-09
Rename to Oasis ROSE Wallet, except in sidebar

lukaw3d:


.changelog/1879.breaking.md
```diff
@@ -0,0 +1 @@
+Rename to ROSE Wallet
```
buberdds, line 1: - Rename to Oasis ROSE Wallet
- why it's breaking? we do not plan to bump major version with this update, right?

lukaw3d: Im not sure. Process then?

buberdds:
We should take into account meta tags like title or og:title and use full product name there.

lukaw3d:
> We should take into account meta tags like title or og:title and use full product name there.

aw, i liked the tab name:
image

currently google would look like this:
image

currently slack would look like this:
image

@tjanez should we change these to Oasis ROSE Wallet?

buberdds (approved):
from code perspective it looks good. Product wise I would wait for Tadej.

.changelog/1879.breaking.md
tjanez: I would classify it as a feature and not breaking.

We will bump major version when we release extension and web from the same codebase.

tjanez:
> @tjanez should we change these to Oasis ROSE Wallet?

Yes, change these to Oasis ROSE Wallet.

PR #1900, opened 2024-04-20
Update and deduplicate sub-dependencies using `npx yarn-deduplicate yarn.lock`

lukaw3d:
Assigning a few extra people, in case you want to do the same in repos you maintain

`rm -rf node_modules/ && yarn install`
before: 35.19 seconds , 735 MB
after: 9.28 seconds , 573 MB

### DIFF
--- a/.changelog/1900.internal.md
+++ b/.changelog/1900.internal.md
@@ -1,0 +1,3 @@
+Update and deduplicate sub-dependencies using `npx yarn-deduplicate yarn.lock`
+
+This reduces the size of node_modules and speeds up installing dependencies.
--- a/src/app/components/AddEscrowForm/__tests__/__snapshots__/index.test.tsx.snap
+++ b/src/app/components/AddEscrowForm/__tests__/__snapshots__/index.test.tsx.snap
@@ -107,9 +107,12 @@ exports[`<AddEscrowForm /> should match snapshot 1`] = `
   box-shadow: none;
 }
 
-.c8:focus:not(:focus-visible) > circle,.c8:focus:not(:focus-visible) > ellipse,
-.c8:focus:not(:focus-visible) > line,.c8:focus:not(:focus-visible) > path,
-.c8:focus:not(:focus-visible) > polygon,.c8:focus:not(:focus-visible) > polyline,
+.c8:focus:not(:focus-visible) > circle,
+.c8:focus:not(:focus-visible) > ellipse,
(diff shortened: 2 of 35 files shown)


lubej (approved):
LGTM, but why did this PR affect snapshot files formatting?

lukaw3d:
jest snapshots -> jest-styled-components -> @adobe/css-tools upgraded from 4.0.1 to 4.3.1 (difference starts at 4.0.2)

lubej:
> jest snapshots -> jest-styled-components -> @adobe/css-tools upgraded from 4.0.1 to 4.3.1 (difference starts at 4.0.2)

`yarn-deduplicate` is a bit misleading(naming) then, as it updates the dependencies as well.

buberdds:
If we blindly update deps with "highest" deduplication strategy please at least mention this in a changelog. For me it is unexpected that we bump redux in this PR for example.

matevz:
NPM hell never seem to stop amuse me. Why even having yarn.lock files, if it gets overridden by yarn-deduplicate? We can just hope that there are no bugs in the sub-sub-sub-package A caused by the common carrot version dependency in another sub-sub-package B, because A didn't test it.

lukaw3d:
@matevz Lock file is just for reproducibility. We can delete it and reinstall from just our package.json, and that would deduplicate too, but also update all transitive dependencies to their latest valid versions.

Each package still specifies which subdependecy versions it supports, e.g.
https://github.com/testing-library/jest-dom/blob/v6.1.5/package.json#L82-L83 "@adobe/css-tools": "^4.3.1",
https://github.com/styled-components/jest-styled-components/blob/v7.2.0/package.json#L56-L57 "@adobe/css-tools": "^4.0.1"
deduplicated to 4.3.1 (it wouldn't deduplicate if something needed "4.0.2" or "^3.1.1")

But yes, even these updates within supported ranges can add bugs

lukaw3d:
Updated changelog and commit message to:
```
Update and deduplicate sub-dependencies using `npx yarn-deduplicate yarn.lock`

This reduces the size of node_modules and speeds up installing dependencies.
```

matevz:
> But yes, even these updates within supported ranges can add bugs

That's exactly what I fear yes. Mostly the carrot versioning is used in `package.json` and the package maintainers typically run tests on exact dependencies inside their yarn.lock file. Deduplication effectively bumps all of those and we will encounter bugs before the original package maintainers will.

PR #1903, opened 2024-04-24
Add active state to mobile navigation buttons

lukaw3d:
| Before | After |
| --- | --- |
|  localhost_3000_account_oasis1qrtyn2q78jv6plrmexrsrv4dh89wv4n49udtg2km(extension375x600) |  localhost_3000_account_oasis1qrtyn2q78jv6plrmexrsrv4dh89wv4n49udtg2km_fiat(extension375x600) |

### DIFF
--- a/src/app/components/MobileFooterNavigation/index.tsx
+++ b/src/app/components/MobileFooterNavigation/index.tsx
@@ -25,13 +25,19 @@ const StyledMobileFooterNavigation = styled.nav`
   right: 0;
   height: ${mobileFooterNavigationHeight};
   display: flex;
-  align-items: center;
-  justify-content: space-evenly;
+  align-items: stretch;
+  justify-content: space-between;
   box-shadow: ${({ theme }) =>
     `0px 0px ${theme.global?.borderSize?.xsmall} ${normalizeColor('background-front-border', theme)}`};
   border-top: ${({ theme }) =>
     `solid ${theme.global?.borderSize?.xsmall} ${normalizeColor('background-contrast', theme)}`};
   flex-direction: row;
+`
+
+const StyledNavLink = styled(NavLink)`
+  &:hover {
+    background-color: ${({ theme }) => normalizeColor('background-contrast', theme)};
+  }
 `
 
 export interface MobileFooterNavigationProps {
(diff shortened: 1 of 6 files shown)


src/app/components/MobileFooterNavigation/index.tsx
```diff
@@ -23,15 +25,27 @@ const StyledMobileFooterNavigation = styled.nav`
...
   flex-direction: row;
 `
 
+const StyledNavLink = styled(NavLink)`
+  // Make items mostly-equal width
+  flex-grow: 1;
+  flex-basis: 0;
+  &:first-of-type {
+    flex-grow: 1.1;
+  }
```
lubej, line 43: Not a fan of this, would rather see a rename from "ROSE Wallet" -> "Wallet" to avoid this.

buberdds:
Hey @donouwens, is that blue background (ROSE Wallet button) in dark theme correct? For me it does not match any other color used in a layout.

Screenshot from 2024-04-25 11-13-14

src/app/components/MobileFooterNavigation/index.tsx
```diff
@@ -23,15 +25,27 @@ const StyledMobileFooterNavigation = styled.nav`
...
   border-top: ${({ theme }) =>
     `solid ${theme.global?.borderSize?.xsmall} ${normalizeColor('background-contrast', theme)}`};
   flex-direction: row;
 `
 
+const StyledNavLink = styled(NavLink)`
+  // Make items mostly-equal width
+  flex-grow: 1;
+  flex-basis: 0;
+  &:first-of-type {
```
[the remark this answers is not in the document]
lubej: To make it 110% width compared to other items, due to text probably cutting out. But this only works on the extension width size. It would fail on tablet size I guess - not sure about when the mobile tabs are displayed exactly.

src/app/components/MobileFooterNavigation/index.tsx
lukaw3d: I think we are forced to keep "ROSE Wallet"

src/app/components/MobileFooterNavigation/index.tsx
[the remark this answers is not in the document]
lukaw3d: > what was a reason to change only first item? mimic white-space: nowrap;?

Yeah but it already seems to fit into one line now, so can remove

buberdds (approved):

buberdds:
eslint/prettier formatting

src/app/components/MobileFooterNavigation/index.tsx
```diff
@@ -23,15 +25,24 @@ const StyledMobileFooterNavigation = styled.nav`
...
+  justify-content: space-between;
   box-shadow: ${({ theme }) =>
     `0px 0px ${theme.global?.borderSize?.xsmall} ${normalizeColor('background-front-border', theme)}`};
   border-top: ${({ theme }) =>
     `solid ${theme.global?.borderSize?.xsmall} ${normalizeColor('background-contrast', theme)}`};
   flex-direction: row;
 `
 
+const StyledNavLink = styled(NavLink)`
+  // Make items mostly-equal width
```
lubej, line 38: This comment has no context now

lukaw3d: should be "Make items equal width" now

lubej: I think comment is not needed at all, as it is straight forward from the code.

lukaw3d: `flex-basis: 0;` seemed unintuitive to me :/ I wouldn't need a comment if it was `grid-template-columns: 1fr 1fr 1fr 1fr;`

PR #1904, opened 2024-04-24
Add logo to login form

lukaw3d:
| Before | After |
| --- | --- |
|  Screenshot from 2024-04-24 23-09-25 |  Screenshot from 2024-04-24 23-07-14 |

### DIFF
--- a/src/app/components/Persist/LoginModalLayout.tsx
+++ b/src/app/components/Persist/LoginModalLayout.tsx
@@ -20,6 +20,11 @@ export function LoginModalLayout(props: {
       <Navigation />
       <ResponsiveLayer
         modal
+        responsive={false}
+        style={{
+          width: '100%',
+          maxWidth: '500px',
+        }}
         background="background-front"
         onClickOutside={props.onClickOutside}
         onEsc={props.onEsc}
@@ -33,7 +38,7 @@ export function LoginModalLayout(props: {
             />
           </Box>
 
-          <Header level={2} textAlign="center">
+          <Header level={2} fill textAlign="center">
             {props.title}
           </Header>
 
--- a/src/app/components/Persist/LoginModalLayout.tsx
+++ b/src/app/components/Persist/LoginModalLayout.tsx
@@ -1,8 +1,11 @@
 import { Box } from 'grommet/es6/components/Box'
 import { ResponsiveLayer } from '../ResponsiveLayer'
-import React from 'react'
+import React, { useContext } from 'react'
 import { Header } from 'app/components/Header'
 import { Navigation } from '../Sidebar'
+import walletBlueLogotype from '../../../../public/Rose Wallet Logo Blue cropped.svg'
+import walletWhiteLogotype from '../../../../public/Rose Wallet Logo White cropped.svg'
+import { ThemeContext } from 'styled-components'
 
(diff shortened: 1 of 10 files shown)


buberdds (commented):
something is wrong with a layout 

Screenshot from 2024-04-25 10-56-02

lubej:
> something is wrong with a layout
> 
> Screenshot from 2024-04-25 10-56-02

Yes, `max-width` being set on the title and paragraph for some reason.

buberdds:
The reason is Grommet. It is adding max-width to many components out of a box. We will have to use `fill` prop in Paragraph and Header.

lukaw3d:
Fixed (added fill, but also limited modal width, otherwise it gets super wide)

| 1 | 2 | 3 |
| --- | --- | --- |
| localhost_3000_create-wallet  | localhost_3000_create-wallet (1)  |  localhost_3000_create-wallet (2) |

src/app/components/Persist/LoginModalLayout.tsx
```diff
@@ -1,26 +1,43 @@
...
+  const { dark } = useContext<any>(ThemeContext)
+
   return (
     <Box direction="row-responsive" background="background-back" fill style={{ minHeight: '100dvh' }}>
       <Navigation />
       <ResponsiveLayer
         modal
+        responsive={false}
+        style={{
+          width: 'min(100%, 500px)',
```
lubej, line 25: Can we avoid inline styles, and use styled-components instead?

Also don't see how this is different from:

```css
width: 100%
max-width: 500px;
```

Make is much easier to read.

src/app/components/Persist/LoginModalLayout.tsx
```diff
@@ -1,26 +1,43 @@
...
         onClickOutside={props.onClickOutside}
         onEsc={props.onEsc}
       >
         <Box pad="medium">
-          <Header level={2} textAlign="center" margin={{ top: 'medium' }}>
+          <Box align="center" margin={{ vertical: 'large' }}>
+            <img
+              alt="ROSE Wallet logo"
+              src={dark ? walletWhiteLogotype : walletBlueLogotype}
+              style={{ height: '65px' }}
```
lubej, line 36: Avoid inline styles. Also, logo looks a bit large to me. Can we scale it down a bit?

Compared to logo in the navigation bar:
image

lubej (approved):

src/app/components/Persist/LoginModalLayout.tsx
lukaw3d: Inline styles:
I would refactor if it's reusable

max-width:
Oh, yep, that does work
(I started from https://github.com/oasisprotocol/oasis-wallet-web/blob/ab5675f07b50e06ba2511d55bf6c9e7c7c98df20/src/app/pages/ParaTimesPage/ParaTimeContent/index.tsx#L23 but that doesn't work correctly in this case)

src/app/components/Persist/LoginModalLayout.tsx
lukaw3d: I was looking at old extension, but maybe that isn't intentionally that large: 
image

I'll reduce to 60 :shrug:

PR #1905, opened 2024-04-24
Denser tab buttons in settings

lukaw3d:
| Before | After |
| --- | --- |
| Screenshot from 2024-04-24 23-15-55  |  Screenshot from 2024-04-24 23-15-32 |

### DIFF
--- a/extension/src/popup.html
+++ b/extension/src/popup.html
@@ -5,7 +5,7 @@
     <title>ROSE Wallet Extension</title>
   </head>
   <body>
-    <div id="root" style="min-width: 360px; min-height: 600px"></div>
+    <div id="root" style="min-width: 375px; min-height: 600px"></div>
     <script type="module" src="./popup/popup.tsx"></script>
     <!--
       TODO: re-enable after supporting multiple extension popups with persistence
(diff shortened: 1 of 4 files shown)


extension/src/popup.html
```diff
@@ -5,7 +5,7 @@
     <title>ROSE Wallet Extension</title>
   </head>
   <body>
-    <div id="root" style="min-width: 360px; min-height: 600px"></div>
+    <div id="root" style="min-width: 375px; min-height: 600px"></div>
```
buberdds, line 8: Aligning popup width to EN tab labels seems naive. Anyway, if we need more space can we go with current/old extension 450 instead? More langs will look Ok. Maybe we should use "Accounts" label too to save some space. It's quite clear that accounts listed under profile belongs to user.

buberdds (approved):

extension/src/popup.html
lukaw3d: Preview of german:
Screenshot from 2024-04-25 17-33-27

lubej (approved):

extension/src/popup.html
lukaw3d: I'd like a wider extension too, and we can update that very easily, but we'll probably want to support 375px iphones either way

PR #1913, opened 2024-05-03
Improve the text for creating a profile

lukaw3d:
| Before | After |
| --- | --- |
|  before | localhost_3000_create-wallet(extension375x600 + scrollbar)  |

Mocks:
image

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -391,10 +391,11 @@
   },
   "persist": {
     "createProfile": {
-      "choosePassword": "Choose a password",
       "passwordMismatch": "Entered password does not match",
-      "repeatPassword": "Re-enter your password",
-      "startPersisting": "Store private keys locally, protected by a password",
+      "repeatPassword": "Confirm your password",
+      "startPersisting": "Store your private keys locally and protect them with a password by creating a profile.",
+      "startPersistingToggle": "Create a profile",
+      "undependableAcknowledge": "I understand this password and profile do not substitute my mnemonic.",
       "unsupported": "Persistence is not supported in this browser."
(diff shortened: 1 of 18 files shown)


src/locales/en/translation.json
```diff
@@ -391,10 +391,11 @@
   },
   "persist": {
     "createProfile": {
-      "choosePassword": "Choose a password",
       "passwordMismatch": "Entered password does not match",
-      "repeatPassword": "Re-enter your password",
-      "startPersisting": "Store private keys locally, protected by a password",
+      "repeatPassword": "Confirm your password",
+      "startPersisting": "Store your private keys locally and protect them with a password by creating a profile.",
```
lubej, line 396: "Securely store your private keys on your local device, protected by password."?

src/locales/en/translation.json
```diff
@@ -391,10 +391,11 @@
...
+      "startPersistingToggle": "Create a profile",
+      "undependableAcknowledge": "I understand this password and profile do not substitute my mnemonic.",
```
lubej, line 398: I understand that neither password or profile serve as a substitution for my mnemonic.

lubej (commented):
Would be nice if we get someone from marketing(native speaker) to go over the texts.

lubej (approved):

src/locales/en/translation.json
lukaw3d: cc @donouwens do you want to iterate more on it

lukaw3d: (i'm merging in current state just to avoid code conflict, and will revisit)

PR #1915, opened 2024-05-03
Show debondable amount next to debonding input

lukaw3d:
| Before | After |
| --- | --- |
|  image | image  |


src/app/components/ReclaimEscrowForm/index.tsx
```diff
@@ -109,8 +110,14 @@ export const ReclaimEscrowForm = memo((props: Props) => {
             }
             reverse
           />
+          <Box align="end" margin={{ top: 'xsmall' }}>
```
lubej, line 113: Can we later add a feature, upon clicking the amount, it would fill in the amount field for you.

lubej (approved):

src/app/components/ReclaimEscrowForm/index.tsx
lukaw3d: That would be nice

also, should make this and paratimes "Max" button consistent
image

PR #1917, opened 2024-05-04
Hide account derivation path in import dialog

lukaw3d:
| Before | After |
| --- | --- |
| localhost_3000_open-wallet_mnemonic(extension375x600 + scrollbar)   | localhost_3000_create-wallet(extension375x600 + scrollbar) |
| localhost_3000_open-wallet_mnemonic | localhost_3000_open-wallet_mnemonic (1) |

### DIFF
--- a/src/app/components/Toolbar/Features/Account/DerivationFormatter.tsx
+++ b/src/app/components/Toolbar/Features/Account/DerivationFormatter.tsx
@@ -12,7 +12,7 @@ export const DerivationFormatter = (props: DerivationFormatterProps) => {
   const { t } = useTranslation()
   const walletTypes: { [type in WalletType]: string } = {
     [WalletType.UsbLedger]: t('toolbar.wallets.type.usbLedger', 'USB Ledger'),
-    [WalletType.BleLedger]: t('toolbar.wallets.type.bluetoothLedger', 'BLE Ledger'),
+    [WalletType.BleLedger]: t('toolbar.wallets.type.bluetoothLedger', 'Bluetooth Ledger'),
     [WalletType.Mnemonic]: t('toolbar.wallets.type.mnemonic', 'Mnemonic'),
     [WalletType.PrivateKey]: t('toolbar.wallets.type.privateKey', 'Private key'),
   }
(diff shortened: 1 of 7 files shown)


src/app/components/Toolbar/Features/Account/DerivationFormatter.tsx
```diff
@@ -12,7 +12,7 @@ export const DerivationFormatter = (props: DerivationFormatterProps) => {
   const { t } = useTranslation()
   const walletTypes: { [type in WalletType]: string } = {
     [WalletType.UsbLedger]: t('toolbar.wallets.type.usbLedger', 'USB Ledger'),
-    [WalletType.BleLedger]: t('toolbar.wallets.type.bluetoothLedger', 'BLE Ledger'),
+    [WalletType.BleLedger]: t('toolbar.wallets.type.bluetoothLedger', 'Bluetooth Ledger'),
```
lubej, line 15: This was exclusively used in the import modal(I think). Due to layout breaking.

src/app/components/Toolbar/Features/Account/ImportableAccount.tsx
```diff
@@ -17,10 +17,6 @@ export const ImportableAccount = ({
       displayBalance={true}
       displayCheckbox={true}
       displayAccountNumber={true}
-      displayDerivation={{
```
lubej, line 20: I kinda liked the derivation path. Any reason it is going away?

lubej (approved):

src/app/components/Toolbar/Features/Account/DerivationFormatter.tsx
lukaw3d: "BLE" was to squeeze in import modal, but it's also displayed in settings > manage account
image

src/app/components/Toolbar/Features/Account/ImportableAccount.tsx
lukaw3d: With Don we were trying to improve illegible "BLE Ledger":
- we tried and liked an icon https://icons.grommet.io/?s=bluetooth instead
- we didn't find the best icon for usb https://icones.js.org/collection/all?s=usb but could also just keep "USB"
- balance position depends on the account type, and longer derivation paths needed fixing
  image

We also don't show it inside account selector, where it would actually add more information:
image

Removing it in import and keeping it in account manage and expanding BLE to Bluetooth was simplest. But I like derivation too :/

lukaw3d:
Replaced by https://github.com/oasisprotocol/oasis-wallet-web/pull/1989

Issue #1868, opened 2024-04-04
Trim 0x from the Ethereum Private key


matevz:
In the Withdraw from ParaTime screen, when entering the hex-encoded private key, trim the leading `0x` if provided. It is not clear when to put or not to put `0x` in front of it, so we should support both.

Related https://github.com/oasisprotocol/cli/issues/106

PR #1923, opened 2024-05-07
Allow to use eth private key that starts with 0x - version 2

buberdds:
Closes https://github.com/oasisprotocol/oasis-wallet-web/issues/1868 

- keep changes limited to TansactionRecipient 
- add `ethPrivateRawKey` field to transaction form


src/app/state/paratimes/types.ts
```diff
@@ -23,7 +23,10 @@ export interface TransactionForm {
   confirmTransferToValidator: boolean
   confirmTransferToForeignAccount: boolean
   defaultFeeAmount: string
+  // compatible with oasisRT.signatureSecp256k1
   ethPrivateKey: string
+  // provided by user and used in form inputs allowing back and forth form navigation
+  ethPrivateRawKey: string
```
lukaw3d, line 29: (I'd name it ethPrivateKeyRaw so it would show up in the searches I did a few months ago, looking for leaks of anything named `privatekey`)

lukaw3d (approved):
Nice!

PR #1927, opened 2024-05-09
Update mobile brand assets

buberdds:
https://app.clickup.com/t/8694cukt5

- adaptive icon sample

Screenshot from 2024-05-09 11-09-49
Screenshot from 2024-05-09 11-09-39


lubej (commented):
Can we commit the changes for `resources` folder? The folder that assets get generated from. So it's easy to reproduce the current results. + did you minify the original icon/splash?

lubej:
The new design does look way cooler, gg @buberdds on suggesting it 👍

buberdds:
> Can we commit the changes for `resources` folder? The folder that assets get generated from. So it's easy to reproduce the current results. + did you minify the original icon/splash?

lol I did not even notice this folder. Why it's not called `assets` like in Capacitor docs?

lubej:
> 1. lol I did not even notice this folder. Why it's not called `assets` like in Capacitor docs?

It does support both folders. https://github.com/ionic-team/capacitor-assets#usage. But I think resources was legacy I guess, because before it was community package(I think).

> 2. I did not minify source files, only generated ones.

Wait, you went through the trouble to minify every generated one? The last I did it, I just minified the ones in resources folder.

> 3. hm I don't like idea of adding 9MB of source image files to the repo. These files are controlled and prepared by designers not us

But if we nuke android/ios due to an upgrade, or something changes in the structure(like new device aspect ration in nay of the platforms). It would be easier, for consistent asset generation. I would just minify them first.

README.md
```diff
@@ -190,9 +190,13 @@ Adding a new language:
...
 1. first add it to Transifex and translate the strings,
 2. create a folder with the new language code in `src/locales`
-and download the translation file there,
+   and download the translation file there,
 3. add the new language to the [list of resources][i18n.ts]
 
+## Mobile app development
+
+[Capacitor and Ionic docs](docs/mobile-development.md)
+
```
lubej, line 199: Once we release the mobile platform, it would be nice to add extensive guide on how to build, prepare OS for both platforms, etc. 👍

buberdds:
> Wait, you went through the trouble to minify every generated one? The last time I did it, I just minified the ones in resources folder.
if the trouble is to run one CLI command than yes. Maybe that's why most of prev assets were much larger

> But if we nuke android/ios due to an upgrade, or something changes in the structure(like new device aspect ration in any of the platforms). It would be easier, for consistent asset generation. I would just minify them first.

For image manipulation like scaling I guess it's better to use original source file not minified, but Ok I will add assets folder.

lubej:
> if the trouble is to run one CLI command than yes. Maybe that's why most of prev assets were much larger

Yes, true, that way it is really simple. Didn't think of that 😅 

> For image manipulation like scaling I guess it's better to use original source file not minified, but Ok I will add assets folder.

True, but it is usually scaled down from the original image anyway. Don't think there is an image larger than the source file. But anyhow, I agree your way is better 👍 I personally would appreciate if the source files are added to the repo.If you strongly disagree we can leave them out.

lubej (approved):

buberdds:
hm cannot merge due to stuck `Post Run satackey/action-docker-layer-caching@v0.0.11`

buberdds:
manually wiped out github caches (100+) for this PR as we were running out of vail cache space. will create issue to remove satackey/action-docker-layer-caching as it is deprecated

PR #1928, opened 2024-05-09
Fix physical back button behavior on Android

lubej:
## Solution

Current behavior of physical back button on Android is unexpected. Back button was closing the app, which is not the expected behavior. This correctly integrates with `react-router-dom` to navigate one page in history back and only closes the app, if there is no history.

## Resources

https://github.com/oasisprotocol/oasis-wallet-web/assets/9722540/448d4168-03c1-4a1f-944d-e19d03e669e7

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
   "dependencies": {
     "@capacitor-community/bluetooth-le": "3.0.0",
     "@capacitor/android": "5.0.4",
+    "@capacitor/app": "5.0.5",
     "@capacitor/core": "5.0.4",
     "@capacitor/ios": "5.0.4",
     "@ethereumjs/util": "9.0.3",
--- a/src/app/components/Ionic/IonicProvider.tsx
+++ b/src/app/components/Ionic/IonicProvider.tsx
@@ -1,0 +1,43 @@
+import { createContext, FC, PropsWithChildren, useEffect } from 'react'
+import { useNavigate } from 'react-router-dom'
+import { Capacitor, type PluginListenerHandle } from '@capacitor/core'
+import { App } from '@capacitor/app'
+
+const IonicContext = createContext<undefined>(undefined)
+
+let backButtonListenerHandle: PluginListenerHandle | undefined
+
+const IonicContextProvider: FC<PropsWithChildren> = ({ children }) => {
+  const navigate = useNavigate()
+
+  useEffect(() => {
+    /**
+     * The back button refers to the physical back button on an Android device and should not be confused
+     * with either the browser back button or ion-back-button.
+     */
+    backButtonListenerHandle = App.addListener('backButton', ({ canGoBack }) => {
+      if (!canGoBack) {
+        App.exitApp()
+      } else {
+        navigate(-1)
+      }
+    })
+
+    return () => {
+      if (backButtonListenerHandle) {
+        backButtonListenerHandle.remove()
+        backButtonListenerHandle = undefined
+      }
+    }
+  }, [navigate])
+
+  return <IonicContext.Provider value={undefined}>{children}</IonicContext.Provider>
+}
+
+export const IonicProvider: FC<PropsWithChildren> = ({ children }) => {
+  if (Capacitor.isNativePlatform()) {
+    return <IonicContextProvider>{children}</IonicContextProvider>
+  }
+
+  return children
+}
--- a/.changelog/1928.bugfix.md
+++ b/.changelog/1928.bugfix.md
@@ -1,0 +1,1 @@
+Fix physical back button behavior on Android
--- a/android/app/capacitor.build.gradle
+++ b/android/app/capacitor.build.gradle
@@ -10,6 +10,7 @@ android {
 apply from: "../capacitor-cordova-android-plugins/cordova.variables.gradle"
 dependencies {
     implementation project(':capacitor-community-bluetooth-le')
+    implementation project(':capacitor-app')
 
 }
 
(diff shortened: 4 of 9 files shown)


lubej:
FYI: Tried with https://ionic-docs-5utg8ms4c-ionic1.vercel.app/docs/v5/developing/hardware-back-button but didn't seem to work for some reason.

src/app/components/Ionic/IonicProvider.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const navigate = useNavigate()
+
+  useEffect(() => {
+    /**
+     * The back button refers to the physical back button on an Android device and should not be confused
+     * with either the browser back button or ion-back-button.
+     */
+    App.addListener('backButton', ({ canGoBack }) => {
+      if (!canGoBack) {
+        App.exitApp()
```
buberdds, line 18: shoudn't we call `minimizeApp` here?

src/app/components/Ionic/IonicProvider.tsx
```diff
@@ -0,0 +1,38 @@
+import { createContext, FC, PropsWithChildren, useEffect } from 'react'
+import { useNavigate } from 'react-router-dom'
+import { Capacitor } from '@capacitor/core'
+import { App } from '@capacitor/app'
+
+const IonicContext = createContext<undefined>(undefined)
```
buberdds, line 6: Why are we using React context for attaching listeners? it's not going to be used with always being undef

src/app/components/Ionic/IonicProvider.tsx
```diff
@@ -0,0 +1,38 @@
+import { createContext, FC, PropsWithChildren, useEffect } from 'react'
+import { useNavigate } from 'react-router-dom'
+import { Capacitor } from '@capacitor/core'
+import { App } from '@capacitor/app'
```
buberdds, line 4: I guess it's too late, but is there an option not to add capacitor bundles to web builds for example?

lubej: It's not too late. I was thinking of dynamically importing @capacitor/app since it is only used in Ionic platform. But will still add to the bundle in the current implementation. But I figured we rather implement some Ionic specific features, and see after how it could be consolidated in the best way. So it doesn't add to bundle size on web and extension. If that sounds fine?

src/app/components/Ionic/IonicProvider.tsx
lubej: I figured there will be some Ionic state specific variables coming in the future, and we will need Context anyway. Like device theme probably - to minimize entanglement with web and extension platform. I can convert it to component, but I think we will run into issues sooner than later.

src/app/components/Ionic/IonicProvider.tsx
lubej: Not sure exactly what minimizeApp does. But exitApp actually minimizes the app, and it behaves like any other app on Android at least. This was taken from example anyhow.

src/app/components/Ionic/IonicProvider.tsx
```diff
@@ -0,0 +1,38 @@
...
+    App.addListener('backButton', ({ canGoBack }) => {
+      if (!canGoBack) {
+        App.exitApp()
+      } else {
+        navigate(-1)
+      }
+    })
+
+    return () => {
+      App.removeAllListeners()
```
lukaw3d, line 25: App.addListener returns a way to remove that specific listener

lukaw3d (approved):

src/app/components/Ionic/IonicProvider.tsx
lubej: Nice, yes I was wondering how to do that. Will fix, so it only removes `backButton` listener.

package.json
```diff
@@ -60,6 +60,7 @@
   "dependencies": {
     "@capacitor-community/bluetooth-le": "3.0.0",
     "@capacitor/android": "5.0.4",
+    "@capacitor/app": "6.0.0",
```
lubej, line 63: This version of `@capacitor/app` is wrong. Should be closer to 5.0.4 version(other `@capacitor` packages), to not brake anything. And then a major upgrade can be done to Capacitor v6.

buberdds (approved):

src/app/components/Ionic/IonicProvider.tsx
```diff
@@ -0,0 +1,43 @@
...
+        navigate(-1)
+      }
+    })
+
+    return () => {
+      if (backButtonListenerHandle) {
+        backButtonListenerHandle.remove()
+        backButtonListenerHandle = undefined
+      }
+    }
```
lukaw3d, line 31: Is this different from
```tsx
    const backButtonListenerHandle = App.addListener('backButton', ({ canGoBack }) => {
      if (!canGoBack) {
        App.exitApp()
      } else {
        navigate(-1)
      }
    })

    return () => {
      backButtonListenerHandle.remove()
    }
```

lubej: It's not. I wanted to do it because of StrictMode(running re-render x2 in dev mode), but I should have skipped attaching the listener, if it is already defined.
Anyway inline const is better. Will update with follow up PR.

lubej: Fixed in https://github.com/oasisprotocol/oasis-wallet-web/pull/1932

PR #1931, opened 2024-05-13
Enable system option in theme dropdown

buberdds:
https://app.clickup.com/t/8694h237j

- using system theme on init is already implemented
- based on UX requirements PR adds system option to Profile dropdown. This setting is not avail in homepage (theme selector when account is not open) 

Looking at ClickUp not sure we need something more. We can bind listeners to respond to system theme changes, but I personally don't care about such feature.

### DIFF
--- a/.changelog/1931.trivial.md
+++ b/.changelog/1931.trivial.md
@@ -1,0 +1,1 @@
+Enable system option in theme dropdown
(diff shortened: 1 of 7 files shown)


lubej (approved):
LGTM

Issue #1925, opened 2024-05-08
Create Profile form validation


buberdds:
When required form inputs are empty we should show some feedback when user clicks submit button. Currently, nothing happens.  
![Screenshot from 2024-05-08 12-40-03](https://github.com/oasisprotocol/oasis-wallet-web/assets/891392/4e420013-e3e5-409d-82c3-bf58ef7fc13c)

lukaw3d:
Related to https://github.com/oasisprotocol/oasis-wallet-web/pull/1913
https://github.com/oasisprotocol/oasis-wallet-web/blob/545a5f1bba7d24df715d8ca1013335a48f385c8a/src/styles/theme/ThemeProvider.tsx#L278-L286

Without customized colors:
![Peek 2024-05-11 03-17](https://github.com/oasisprotocol/oasis-wallet-web/assets/3758846/c3a89cc7-f979-4acb-888a-9b732b0973ac)

Or move `required` from CheckBox to FormField and customize the error message:
![image](https://github.com/oasisprotocol/oasis-wallet-web/assets/3758846/c66f9921-4da4-4483-852b-247e300d84d2)

lukaw3d:
If checkbox is not wrapped in FormField it also shows focus:
![image](https://github.com/oasisprotocol/oasis-wallet-web/assets/3758846/48bdeed4-b3ec-4943-981a-e630a82ec698)

PR #1936, opened 2024-05-15
Add validation message to required checkbox in ChoosePasswordFields

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/1925

(also improved checkbox focus state in case we don't have validation messages in some component)

| Before | After |
| --- | --- |
|    Screenshot from 2024-05-15 02-35-06   |   Screenshot from 2024-05-15 02-33-33     |


src/app/components/Persist/ChoosePasswordFields.tsx
```diff
@@ -69,7 +72,6 @@ export function ChoosePasswordFields() {
           >
             <CheckBox
               name="profileStorageUndependableAcknowledge"
-              required
```
buberdds, line 72: I would remove this prop in `ChoosePasswordInputFields` password1 field too so validation is consistent.

buberdds (approved):

src/app/components/Persist/ChoosePasswordFields.tsx
buberdds: as long as it works everywhere. looks like it's used in more than one place.

lukaw3d: :+1: removed

PR #1939, opened 2024-05-15
Make creating an account the default flow

lukaw3d:


lubej (approved):

lubej:
Seems to break quite a lot of tests.

PR #1940, opened 2024-05-16
Prevent app content from being captured in screenshots or apps view

buberdds:
https://developer.android.com/reference/android/view/WindowManager.LayoutParams#FLAG_SECURE

switching apps
Screenshot_20240516-101313

screenshot taken
Screenshot_20240516-101325


lubej (approved):
Nice 👍 

Can we make a follow up task for iOS, so we don't forget about it?

Issue #1910, opened 2024-04-30
Update footer text


lukaw3d:
Motivation: many users may not have github accounts.

Current:
![image](https://github.com/oasisprotocol/oasis-wallet-web/assets/3758846/c81a9719-423e-4e6d-a771-c2842ed64625)

New design:
![image](https://github.com/oasisprotocol/oasis-wallet-web/assets/3758846/867150f7-81db-4928-a66f-6ba04efa606a)

```
ROSE Wallet is open source and powered by 
Oasis Scan API & Oasis gRPC

Terms and Conditions | Version: 1.11.0 [hide details]
(commit: 1b92a77) built at 4/22/2024, 6:08:23 PM

We’d love your feedback at wallet@oasisprotocol.org
```

lukaw3d:
TODO: Move it into settings

PR #1941, opened 2024-05-19
Update footer text

rochelle-norman:
Closes [#1910 ](https://github.com/oasisprotocol/oasis-wallet-web/issues/1910)

### DIFF
--- a/src/app/components/Footer/index.tsx
+++ b/src/app/components/Footer/index.tsx
@@ -2,7 +2,7 @@ import { Anchor } from 'grommet/es6/components/Anchor'
 import { Box } from 'grommet/es6/components/Box'
 import { Text } from 'grommet/es6/components/Text'
 import { ResponsiveContext } from 'grommet/es6/contexts/ResponsiveContext'
-import React, { memo } from 'react'
+import React, { memo, useState } from 'react'
 import { useSelector } from 'react-redux'
 import { Trans, useTranslation } from 'react-i18next'
 import { selectHasAccounts } from 'app/state/wallet/selectors'
@@ -11,9 +11,13 @@ import { backend } from 'vendors/backend'
 import { BackendAPIs } from 'config'
 import { MobileFooterNavigation } from '../MobileFooterNavigation'
 import { mobileFooterNavigationHeight } from '../../../styles/theme/elementSizes'
+import { Button } from 'grommet/es6/components/Button'
 
 const githubLink = 'https://github.com/oasisprotocol/oasis-wallet-web/'
 const githubReleaseLink = (tag: string) => `${githubLink}releases/tag/${tag}`
+const buildTime =
+  process.env.REACT_APP_BUILD_DATETIME &&
+  `${intlDateTimeFormat(Number(process.env.REACT_APP_BUILD_DATETIME))}`
 
 function NoReleaseLink() {
   return <>-</>
@@ -23,6 +27,11 @@ export const Footer = memo(() => {
   const walletHasAccounts = useSelector(selectHasAccounts)
   const isMobile = React.useContext(ResponsiveContext) === 'small'
   const { t } = useTranslation()
+  const [showCommitLink, setShowCommitLink] = useState(false)
+
+  const toggleCommitLink = () => {
+    setShowCommitLink(!showCommitLink)
+  }
 
   const backendToLabel = {
     [BackendAPIs.OasisMonitor]: t(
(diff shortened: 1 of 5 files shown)


src/app/components/Footer/index.tsx
```diff
@@ -84,24 +122,15 @@ export const Footer = memo(() => {
...
-              ),
-            }}
-            defaults="Version: <ReleaseLink/> (commit: <CommitLink/>) built at {{buildTime}}"
-            values={{
-              buildTime: intlDateTimeFormat(Number(process.env.REACT_APP_BUILD_DATETIME)),
+              CommitLink: <CommitLink />,
             }}
+            defaults="Version: <ReleaseLink/> <CommitLink/>"
           />
-          {poweredByLabel && <Box align="center">{poweredByLabel}</Box>}
```
buberdds, line 101: Linters should throw an error because `poweredByLabel` is defined, but not it is not used.

src/app/components/Footer/index.tsx
```diff
@@ -53,23 +87,27 @@ export const Footer = memo(() => {
...
-          i18nKey="footer.terms"
-          t={t}
-          components={{
-            TermsLink: (
-              <Anchor href="https://wallet.oasis.io/t-c.pdf" target="_blank" rel="noopener noreferrer" />
-            ),
-          }}
-          defaults="<TermsLink>Terms and Conditions</TermsLink>"
+          defaults="ROSE Wallet is fully <GithubLink>open source</GithubLink> - and powered by
+          Oasis Scan API & Oasis gRPC"
```
buberdds, line 91: Oh and we prob want to keep dynamic "backend label" too. Like we plan to support Nexus in the future so we can switch between backend APIs when one is having issues.

src/app/components/Footer/index.tsx
```diff
@@ -23,6 +26,37 @@ export const Footer = memo(() => {
   const walletHasAccounts = useSelector(selectHasAccounts)
   const isMobile = React.useContext(ResponsiveContext) === 'small'
   const { t } = useTranslation()
+  const [showCommitLink, setShowCommitLink] = useState(false)
+
+  const toggleCommitLink = () => {
+    setShowCommitLink(!showCommitLink)
+  }
+
+  function CommitLink() {
```
buberdds, line 35: I would keep it inline as it's not reusable.

src/locales/en/translation.json
```diff
@@ -188,13 +188,13 @@
     "thirdPartyDisclaimer": "This service is provided by an external party. Oasis* does not carry any responsibility. All fees charged by Transak."
   },
   "footer": {
-    "github": "ROSE Wallet is fully <GithubLink>open source</GithubLink> - Feedback and issues are appreciated!",
+    "github": "ROSE Wallet is fully <GithubLink>open source</GithubLink> - and powered by\n          Oasis Scan API & Oasis gRPC",
```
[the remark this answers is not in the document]
rochelle-norman: Thanks, you are right this doesn't need changing as I have reverted it back to the `poweredByLabel`

src/app/components/Footer/index.tsx
rochelle-norman: Good point, I have moved it inline

buberdds:
Nice work btw, thank you. I will add three more comments, so if you are running out of time or patience we can combine forces.

src/app/components/Footer/index.tsx
```diff
@@ -85,23 +103,46 @@ export const Footer = memo(() => {
...
               ),
               CommitLink: (
-                <Anchor
-                  href={`${githubLink}commit/${process.env.REACT_APP_BUILD_SHA}`}
-                  label={process.env.REACT_APP_BUILD_SHA.substring(0, 7)}
-                  target="_blank"
-                  rel="noopener noreferrer"
-                />
+                <span>
+                  <Anchor
```
buberdds, line 107: Accessibility wise, this element should be a button (Grommet Button)

src/app/components/Footer/index.tsx
```diff
@@ -85,23 +103,46 @@ export const Footer = memo(() => {
...
           />
-          {poweredByLabel && <Box align="center">{poweredByLabel}</Box>}
         </Text>
       )}
-
+      <Text size="small" textAlign="center" margin={{ top: 'small' }}>
+        <Trans
+          i18nKey="footer.feedback"
+          t={t}
+          defaults="We’d love your feedback at wallet@oasisprotocol.org"
```
buberdds, line 143: wallet@oasisprotocol.org should be mailto: link

buberdds:
When you run `yarn test` you will see that some tests will need some tweaks. 
For updating snapshots run `yarn jest -u`
Other one will fail because link with hash is not longer visible. Either simulate click using `import userEvent from '@testing-library/user-event'` or remove the test (we can refactor them in the future)

src/app/components/Footer/index.tsx
rochelle-norman: Good point, I have updated it. Not sure 100% if the props are correct though.

src/app/components/Footer/index.tsx
rochelle-norman: Thanks, I have updated it.

buberdds (approved):

buberdds:
Thanks!

PR #1943, opened 2024-05-21
Add a delay before user can finish V0 migration to encourage reading

lukaw3d:
[screen-recording.webm](https://github.com/oasisprotocol/oasis-wallet-web/assets/3758846/bf25e9c8-7798-432b-a8b1-c48b3ca921f8)

### DIFF
--- a/src/app/components/Persist/MigrateV0StateForm.tsx
+++ b/src/app/components/Persist/MigrateV0StateForm.tsx
@@ -23,6 +23,7 @@ import { themeActions } from '../../../styles/theme/slice'
 import { RevealOverlayButton } from '../RevealOverlayButton'
 import { PrivateKeyFormatter } from '../PrivateKeyFormatter'
 import { PasswordWrongError } from '../../../types/errors'
+import { CountdownButton } from '../CountdownButton'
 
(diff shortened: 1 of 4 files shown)


src/app/components/Persist/MigrateV0StateForm.tsx
```diff
@@ -155,14 +156,14 @@ export function MigrateV0StateForm() {
             />
           </FormField>
           {migratingV0State.invalidPrivateKeys.length > 0 ? (
-            <Button
+            <CountdownButton
```
buberdds, line 159: Do we have multiple submit buttons to handle different labels?

buberdds (approved):

src/app/components/Persist/MigrateV0StateForm.tsx
lukaw3d: yeah - ideally these would trigger different handlers too, but we are using a single Form onSubmit. And could be styled differently, but invalid keys are should be so rare that it doesn't matter

lukaw3d: so they are incidentally only different in labels

PR #1945, opened 2024-05-21
Add Android update screen

lubej:
## Solution

Adds an update screen, when update is available on mobile platforms(currently only implemented for Android platform). Where clicking on "Update now" takes the user to the app store page.

## Resources

localhost_3000_(iPhone 12 Pro)


lubej:
Note to self: Does update check run after user enters the wallet again. It should check only once, and then maybe once per day?

lubej:
> Note to self: Does update check run after user enters the wallet again. It should check only once, and then maybe once per day?

Current implementation checks for an update only on app launch. Which should be sufficient for current case, as the update can be skipped anyway. When mandatory update feature start to roll out, we should revisit the topic on what time interval the update check should be implemented.

src/app/components/Ionic/providers/IonicProvider.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const ionicProviderInitialState: IonicProviderState = {
+  updateAvailability: UpdateAvailability.NOT_INITIALIZED,
+  error: null,
+}
+
+export const IonicContextProvider: FC<PropsWithChildren> = ({ children }) => {
+  const [state, setState] = useState<IonicProviderState>({ ...ionicProviderInitialState })
+
+  const { checkForUpdateAvailability } = useIonicRequiresUpdate(state, setState)
```
lukaw3d, line 15: Sending changes into setState like this is hard to trace

lukaw3d (approved):

src/app/components/Ionic/providers/IonicProvider.tsx
lubej: Yes I agree. But making a redux slice just for this seemed like an overkill.

lukaw3d: ew i didn't mean that :D just move useState into useIonicRequiresUpdate and return state out

lubej: Right, my bad 😄 

Here is the follow up for it: https://github.com/oasisprotocol/oasis-wallet-web/pull/1969

PR #1951, opened 2024-05-24
Allow quotes when typing 'delete'

lukaw3d:
since our instructions are `To confirm and proceed, please type 'delete' below.`

### DIFF
--- a/src/app/components/DeleteInputForm/index.tsx
+++ b/src/app/components/DeleteInputForm/index.tsx
@@ -21,7 +21,9 @@ export function DeleteInputForm({ children, onCancel, onConfirm }: DeleteInputFo
       <FormField
         name="type_delete"
         validate={(value: string | undefined) =>
-          !value || value.toLowerCase() !== t('deleteForm.confirmationKeyword', 'delete').toLowerCase()
+          !value ||
+          value.toLowerCase().replace(/['"“”]/g, '') !==
+            t('deleteForm.confirmationKeyword', 'delete').toLowerCase()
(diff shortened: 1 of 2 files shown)


src/app/components/DeleteInputForm/index.tsx
```diff
@@ -21,7 +21,9 @@ export function DeleteInputForm({ children, onCancel, onConfirm }: DeleteInputFo
       <FormField
         name="type_delete"
         validate={(value: string | undefined) =>
-          !value || value.toLowerCase() !== t('deleteForm.confirmationKeyword', 'delete').toLowerCase()
+          !value ||
+          value.toLowerCase().replace(/['"“”]/g, '') !==
```
lubej, line 25: Doesn't this enable you to write `'''''''''delete'''''''''`?

lubej (approved):

src/app/components/DeleteInputForm/index.tsx
lukaw3d: yeah :shrug: still seems fine

lubej: Also `'''''d'e''l'e''t'e'''''` would be possible. But I guess it doesn't matter in the end.

lubej: I would personally rather see we remove apostrophes entirely. And mark the text different(italic, bold), that is supposed to be written down in confirmation dialog.

lukaw3d: good idea
https://github.com/oasisprotocol/oasis-wallet-web/pull/1953

Issue #1942, opened 2024-05-20
Fix updating transactions list


lukaw3d:
Current behavior:
- make a new transaction and request balance and transactions from oasisscan
- request balance from oasisscan every 30 seconds
- if balance changes, request transactions from oasisscan
- request balance from oasisscan every 30 seconds
- if balance changes, request transactions from oasisscan
- request balance from oasisscan every 30 seconds
- if balance changes, request transactions from oasisscan
...

Issue:
Immediately after making a transaction oasisscan responds with old balance and old transactions. After 30 seconds oasisscan responds with changed balance, and then with **old transactions**. After 30 seconds balance is not changed, so we do not try to get new transactions. Wallet is stuck showing old transactions, missing the new transaction.

Solutions:
- maybe just add 15 second delay to fetching transactions after balance changes (delay might be different on testnet)
- fetch transactions every time. Adds overhead to oasisscan backend
- ignore balance change until transactions change

lukaw3d:
Another issue: debonding transactions don't change "available" balance so this is ignored https://github.com/oasisprotocol/oasis-wallet-web/blob/dcb01bfacd7eb043022c5be2581daaeae066fbdb/src/app/state/account/saga.ts#L127

lukaw3d:
Adding a 15 second delay is too low
Checked the delay:
- made a transfer
- +9 seconds
- balance updated
- +22 seconds
- transactions updated
```js
setInterval(async () => {
  const acc = await (await fetch('https://api.oasisscan.com/mainnet/chain/account/info/oasis1qz78ap0456g2rk7j6rmtvasc9v2kjhz2s58qgj90')).json()
  const txs = await (await fetch('https://api.oasisscan.com/mainnet/chain/transactions?size=1&address=oasis1qz78ap0456g2rk7j6rmtvasc9v2kjhz2s58qgj90&runtime=true')).json()
  console.log(new Date().toISOString(), acc.data.available, txs.data.list[0].txHash)
}, 1000)
```

lukaw3d:
Retried:
- transfer
- +8 seconds
- balance and transactions updated

and on testnet
- transfer
- +6 sec
- balance updated
- +31 sec 
- transactions updated

lukaw3d:
Fixed by https://github.com/oasisprotocol/oasis-wallet-web/pull/1954

Issue #1770, opened 2023-11-24
mockApi doesn't work in extension.spec.ts


lukaw3d:
`context.route` only finds requests to `chrome-extension://`

`context.backgroundPages..route` doesn't help either
```tsx
  await context.backgroundPages()[0].route('https:**', async route => {
    console.log(route.request().url())
    route.continue()
  })
```

lukaw3d:
Mocking works since migrating to mv3 and moving requests from bgpage into popup #2120

PR #1954, opened 2024-05-27
Pending transactions

lubej:
## Description

This PR addresses https://github.com/oasisprotocol/oasis-wallet-web/issues/1942. So the initial proposition was that we should decrease the time to fetch transactions. Which in my opinion does not solve the core issue, of not showing pending transaction. So this PR introduces a list of pending transactions, which should mitigate original issue in some way.

I wanted to introduce polling for each of the transactions, that would refresh the state of each pending transaction on each new block. But that introduces a state, where balance could be out of sync with the actual transaction history - also it would need a synchronous clock - which is also a hustle to implement, as the right approach IMO would be to refresh the balance/pending txs on each new block. `oasis.client.NodeInternal` function `consensusWatchBlocks`, would be ideal for that case, but does not seems to work, either there is an issue with TS client implementation or the grpc node does not support the method.

## Resources 

image


lubej:
This PR is still in draft state, requested a review to validate the idea.

src/app/state/transaction/saga.ts
```diff
@@ -201,6 +201,40 @@ export function* doTransaction(action: PayloadAction<TransactionPayload>) {
...
+
+    const transaction: Transaction = {
+      status: undefined,
+      fee: undefined,
+      level: undefined,
+      to: undefined,
+      round: undefined,
+      runtimeId: undefined,
+      runtimeName: undefined,
+      timestamp: undefined,
```
lukaw3d, line 215: I would save a timestamp

src/app/state/transaction/types.ts
```diff
@@ -56,6 +56,15 @@ export interface TransactionState {
...
   preview?: TransactionPreview
+  // TODO: Should pendingTransactions be persisted?
+  /**
+   * Keeps track of pending transactions.
+   *
+   * @typedef {Object} pendingTransactions
+   * @property {Object<string, Transaction[]>} [pendingTransactions] - A dictionary where the keys are addresses
+   *    and the value is array of pending transactions.
+   */
+  pendingTransactions: Record<string, Transaction[]>
```
lukaw3d, line 67: Should probably be in AccountState

src/app/state/transaction/types.ts
```diff
@@ -56,6 +56,15 @@ export interface TransactionState {
   success: boolean
   error?: ErrorPayload
   preview?: TransactionPreview
+  // TODO: Should pendingTransactions be persisted?
```
lukaw3d, line 59: I wouldn't, and I wouldn't even keep them if user switches accounts

I'd just show "has pending transactions" warning after reload by checking account nonce and nonces in latest transactions (but would need to fetch tx details for each transaction in the list when we get them from oasisscan)

src/app/state/transaction/saga.ts
```diff
@@ -201,6 +201,40 @@ export function* doTransaction(action: PayloadAction<TransactionPayload>) {
...
+          }
+        : {}),
+      ...(action.payload.type === 'reclaimEscrow'
+        ? {
+            to: action.payload.validator,
+          }
+        : {}),
+    }
+
+    yield* put(transactionActions.addPendingTransaction({ transaction, from: activeWallet.address }))
```
lukaw3d, line 237: paratime transactions will need handling elsewhere

src/app/state/transaction/index.ts
```diff
@@ -60,6 +62,30 @@ export const transactionSlice = createSlice({
...
+        payload: { from, transaction },
+      } = action
+
+      if (!state.pendingTransactions[from]) {
+        state.pendingTransactions[from] = [transaction]
+      } else {
+        state.pendingTransactions[from] = [transaction, ...state.pendingTransactions[from]]
+      }
+    },
+    removePendingTransaction(state, action: PayloadAction<PendingTransactionPayload>) {
```
lukaw3d, line 77: Instead of consensusWatchBlocks: just remove pending transactions when their hash appears in transactionsLoaded

src/app/components/Transaction/index.tsx
```diff
@@ -479,12 +481,33 @@ export function Transaction(props: TransactionProps) {
...
+            {(() => {
+              switch (transaction.status) {
+                case true:
+                  return <span>{t('account.transaction.successful', 'Successful')}</span>
+                case false:
+                  return <span>{t('account.transaction.failed', 'Failed')}</span>
+                case undefined:
+                default:
+                  /* TODO: Add translation */
+                  return <span>Pending</span>
```
lukaw3d, line 508: This could add confusion
https://github.com/oasisprotocol/explorer/blob/3d82e749424e6d9001041948a8892c126d54dcab/src/oasis-nexus/generated/api.ts#L918-L921
undefined success in nexus api means unknown / confidential

src/app/state/transaction/saga.ts
lubej: You mean `new Date()`? If I understood correctly timestamp is a time when transaction was mined. Which would differ from the pending tx preview, so I would rather skip it.

src/app/state/transaction/types.ts
lubej: Will move, but not sure if it breaks migration or anything else.

src/app/components/AlertBox/index.tsx
```diff
@@ -12,7 +14,8 @@ interface Props {
   center?: boolean
   /** Example `icon={<Info size="20px" />}` */
   icon?: React.ReactNode
-  children: React.ReactNode
+  children?: React.ReactNode
+  contents?: React.ReactNode
```
lubej, line 18: TODO: Refactor AlertBox, so that `title` prop uses existing styling/structure. And `children` should "replace" the whole contents of the AlertBox.,

src/vendors/oasisscan.ts
```diff
@@ -50,7 +53,16 @@ export function getOasisscanAPIs(url: string | 'https://api.oasisscan.com/mainne
     return data
   }
 
-  async function getTransactionsList(params: { accountId: string; limit: number }): Promise<Transaction[]> {
+  async function getTransaction({ hash }: { hash: string }) {
```
lubej, line 56: TODO: Add cache mechanism, as transaction won't ever change once we already fetch it.

src/app/components/Transaction/index.tsx
lubej: Refactored status from `boolean` to `enum` here https://github.com/oasisprotocol/oasis-wallet-web/pull/1970

src/app/components/AlertBox/index.tsx
lubej: Opened here: https://github.com/oasisprotocol/oasis-wallet-web/pull/1975

src/vendors/oasisscan.ts
lubej: https://github.com/oasisprotocol/oasis-wallet-web/pull/1976

src/utils/__fixtures__/test-inputs.ts
```diff
@@ -49,6 +49,13 @@ export const privateKeyUnlockedState = {
     transactions: [],
     transactionsError: undefined,
     loading: false,
+    pendingTransactions: {
+      local: [],
+      testnet: [],
+      mainnet: [],
+    },
+    // TODO: Would expect this nonce to be 0?
+    nonce: '1',
```
lubej, line 58: @lukaw3d some unexpected behaviour occurred here, due to migration. I would expect this nonce to be '0', as it is probably missing in original state. (I added nonce to the Account interface in this PR) 

Also bellow, I would expect wallet to not include nonce at all - as it breaks the interface.

lubej: Ah I see now that `nonce = 1` comes from mocked route here `**/chain/account/info/*`

src/app/pages/AccountPage/Features/TransactionHistory/index.tsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-restricted-syntax */
```
lukaw3d, line 1: danger

src/utils/__fixtures__/test-inputs.ts
```diff
@@ -426,4 +448,4 @@ export const walletExtensionV0UnlockedState = {
     }),
     enteredWrongPassword: false,
   },
-} satisfies RootState
+} as RootState
```
lukaw3d, line 451: This reduces type safety

lubej: It does, but refers to my question here https://github.com/oasisprotocol/oasis-wallet-web/pull/1954#discussion_r1645389582

How come the migration now adds nonce to wallet/s, the only change I made was to the account interface? And saved the nonce from account fetch to the account object. Do you maybe know from the top of your head? Before I make a deep dive.

src/app/components/Transaction/index.tsx
```diff
@@ -461,9 +461,11 @@ export function Transaction(props: TransactionProps) {
               </Box>
```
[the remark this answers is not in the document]
lubej: Apparently detailed tx API, does not return amount field correctly for `staking.Allow`. In raw form the amount is recorded in `amount_change` which does not translate to `amount` field. Nice catch 👍

src/app/pages/AccountPage/Features/TransactionHistory/index.tsx
lubej: It was complaining for no good reason, so that's why I added it. Added "block" disable now, as `eslint-disable-next-line` was not picking it up.

src/utils/__fixtures__/test-inputs.ts
lukaw3d: walletExtensionV0UnlockedState: migrating-v0-ext-persisted-state.spec.ts is fetching real data https://github.com/oasisprotocol/oasis-wallet-web/issues/1770

privateKeyUnlockedState: it is fetching mocked data

lukaw3d: Oh - why is it in wallet/s. I'm not entirely sure, but related to https://github.com/oasisprotocol/oasis-wallet-web/commit/abe01239ff358026b2c6df2c5160cb69c9a5afe8. We aren't removing extra fields when passing balances between account and wallet: https://github.com/oasisprotocol/oasis-wallet-web/blob/8ca2795a923a3469d0f69f371f84535e2cf971f9/src/app/state/wallet/saga.ts#L72

lubej: I am not sure why we even store balances in the wallet/s? As source of truth is on the blockchain, and values may not be correct.

src/app/pages/AccountPage/Features/TransactionHistory/index.tsx
```diff
@@ -11,39 +11,95 @@ import { useTranslation } from 'react-i18next'
...
+              </span>
+            }
+          >
+            {t(
+              'account.summary.someTxsInPendingState',
+              'Some transactions are currently in a pending state.',
+            )}
+          </AlertBox>
+          {pendingTransactionComponents.length && (
+            <Box gap="medium" data-testid="pending-txs" margin={{ top: 'small' }}>
```
lukaw3d, line 91: This prints "0" when hasUnknownPendingTransactions
image

Related https://github.com/oasisprotocol/explorer/issues/259

```suggestion
          {pendingTransactionComponents.length > 0 && (
            <Box gap="medium" data-testid="pending-txs" margin={{ top: 'small' }}>
```

src/app/pages/AccountPage/Features/TransactionHistory/index.tsx
lukaw3d: Ah, sorry it wasted your time. I'll fix linter later

src/utils/__fixtures__/test-inputs.ts
lukaw3d: Just legacy
After loading stored wallets we update their balances through selectNetwork -> refreshOnNetworkChange -> fetchWallet

src/app/components/Transaction/index.tsx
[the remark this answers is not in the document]
lukaw3d: Were any other transaction methods affected?

src/app/state/account/selectors.ts
```diff
@@ -12,3 +13,44 @@ export const selectAccountAddress = createSelector([selectAccount], account => a
...
+        return bigNonce > acc ? bigNonce : acc
+      },
+      undefined as bigint | undefined,
+    )
+
+    if (!maxNonceFromTxs) {
+      return BigInt(accountNonce) > 0n
+    }
+
+    return BigInt(maxNonceFromTxs) + BigInt(1) < BigInt(accountNonce)
```
lukaw3d, line 54: https://7ad67e19.oasis-wallet.pages.dev/account/oasis1qpw4ryz88wfe9mmy2sqpkevlzp4ua8ec7udxfqts
incorrectly always shows "Some transactions are currently in a pending state"

src/app/pages/AccountPage/Features/TransactionHistory/index.tsx
```diff
@@ -11,39 +11,95 @@ import { useTranslation } from 'react-i18next'
...
+    <Box margin="none">
       {transactionsError && (
         <p>
-          {t('account.transaction.loadingError', "Couldn't load transactions.")}{' '}
+          {t('account.transaction.loadingError', `Couldn't load transactions.`)}{' '}
           <ErrorFormatter code={transactionsError.code} message={transactionsError.message} />
         </p>
       )}
+      {/* eslint-disable no-restricted-syntax */}
+      {(!!pendingTransactionComponents.length || hasUnknownPendingTransactions) && (
```
lukaw3d, line 57: It shows pending transactions while loading. I would default to not showing them - it's much more common.

[screen-recording.webm](https://github.com/oasisprotocol/oasis-wallet-web/assets/3758846/eb169067-4b7e-49a2-b2c5-50072159ddd4)

src/app/state/account/saga.ts
```diff
@@ -126,13 +149,14 @@ export function* fetchingOnAccountPage() {
...
           if (
             staleBalances.available !== refreshedAccount.available ||
             staleBalances.delegations !== refreshedAccount.delegations ||
-            staleBalances.debonding !== refreshedAccount.debonding
+            staleBalances.debonding !== refreshedAccount.debonding ||
+            hasPendingTxs
           ) {
             // Wait for oasisscan to update transactions (it updates balances faster)
-            yield* delay(TRANSACTIONS_UPDATE_DELAY)
             yield* call(fetchAccount, startAction)
```
lukaw3d, line 160: Did you try if this works?

src/app/state/account/selectors.ts
lubej: Nice catch! Fixed here https://github.com/oasisprotocol/oasis-wallet-web/pull/1954/commits/5e2036b61d2f0e070dc60a51ff34f2ff7fa872bf the issue was that 0 nonce was not correctly parsed from tx.

src/app/state/account/saga.ts
lubej: Yes, there is no need for delay anymore because of `hasPendingTxs`. Now we know when account is out of sync with transaction history, and keep making requests until nonce will match.

lubej: Unless you can think of a case, where that would not work?

lukaw3d: ah, it was my edge case.
I made transfer to myself, bypassing "Cannot send to yourself" error:
- (on chain: made a transaction, incremented nonce, didn't change balance)
- frontend adds transaction to selectPendingTransactionForAccount
- account nonce isn't updated so hasAccountUnknownPendingTransactions is false
- ACCOUNT_REFETCHING_INTERVAL: fetches account, but detects no balance change and no pending txs
- ACCOUNT_REFETCHING_INTERVAL: fetches account, but detects no balance change and no pending txs

Transfer to another account:
- (on chain: made a transaction, incremented nonce, changed balance)
- frontend adds transaction to selectPendingTransactionForAccount
- account nonce isn't updated so hasAccountUnknownPendingTransactions is false
- ACCOUNT_REFETCHING_INTERVAL: fetches account, **detects balance change**:
  - saves new account info into state (updated balance and **updated nonce**)
  - fetches and saves transactions but they aren't changed
- ACCOUNT_REFETCHING_INTERVAL: fetches account, no balance change, **has pending txs** (updated nonce)
  - fetches and **saves transactions**

lubej:
@lukaw3d for the `decryptWithPasswordV0` what would be the best way to make it pass?

Should I strip the nonces, or should the encrypted state be updated? And how can I update the encrypted state?

lukaw3d (approved):

src/app/state/account/saga.ts
```diff
@@ -47,12 +48,33 @@ export function* fetchAccount(action: PayloadAction<string>) {
...
         try {
-          const transactions = yield* call(getTransactionsList, {
+          const transactions: Transaction[] = yield* call(getTransactionsList, {
             accountId: address,
             limit: TRANSACTIONS_LIMIT,
           })
-          yield* put(accountActions.transactionsLoaded(transactions))
+
+          const detailedTransactions = yield* call(() =>
+            Promise.allSettled(transactions.map(({ hash }) => getTransaction({ hash }))),
```
lukaw3d, line 60: This tries to incorrectly fetch paratime transactions too. See requests in:
https://wallet.dev.oasis.io/account/oasis1qry5d5zxs58w9deu5ara5zaamhkzqtg99s00ls90

lubej: Right. Is there any better way to determine if the tx is a runtime tx -> https://github.com/oasisprotocol/oasis-wallet-web/pull/1999?

Issue #1955, opened 2024-05-27
Disable input number scrolling changes


lubej:
Current behavior:

After you set a value, for example in amount field, accidental hovering over the field and scrolling to navigate can accidentally change value in the input field:

https://www.loom.com/share/51cda8c571d948b5b215f3632dbe4f7c

Proposed behavior:

Disable wheel events on input number type fields.

PR #1956, opened 2024-05-28
Prevent value change on number input wheel scroll

lubej:
Resolves https://github.com/oasisprotocol/oasis-wallet-web/issues/1955

### DIFF
--- a/src/app/lib/__tests__/usePreventChangeOnNumberInputScroll.test.tsx
+++ b/src/app/lib/__tests__/usePreventChangeOnNumberInputScroll.test.tsx
@@ -11,17 +11,15 @@ describe('usePreventChangeOnNumberInputScroll', () => {
     const { getByRole } = render(<TextInputFormItem />)
 
     const numberInput = getByRole('input')
-
-    numberInput.blur = jest.fn()
-    const focusMock = jest.spyOn(numberInput, 'focus')
+    fireEvent.change(numberInput, { target: { value: '123' } })
 
     act(() => {
       fireEvent.focus(numberInput)
       fireEvent.wheel(numberInput)
     })
 
-    expect(numberInput.blur).toHaveBeenCalled()
-    await waitFor(() => expect(focusMock).toHaveBeenCalled())
+    waitFor(() => expect(numberInput).toHaveFocus())
+    expect(numberInput).toHaveValue(123)
   })
 
   it('Should not focus on input that is not focused', async () => {
@@ -29,14 +27,10 @@ describe('usePreventChangeOnNumberInputScroll', () => {
 
     const numberInput = getByRole('input')
 
-    numberInput.blur = jest.fn()
-    const focusMock = jest.spyOn(numberInput, 'focus')
-
     act(() => {
       fireEvent.wheel(numberInput)
     })
 
-    expect(numberInput.blur).toHaveBeenCalled()
-    await waitFor(() => expect(focusMock).not.toHaveBeenCalled())
+    expect(numberInput).not.toHaveFocus()
   })
 })
--- a/src/app/lib/__tests__/usePreventChangeOnNumberInputScroll.test.tsx
+++ b/src/app/lib/__tests__/usePreventChangeOnNumberInputScroll.test.tsx
@@ -1,0 +1,42 @@
+import { render, fireEvent, act, waitFor } from '@testing-library/react'
+import { TextInput } from 'grommet/es6/components/TextInput'
+import { usePreventChangeOnNumberInputScroll } from '../usePreventChangeOnNumberInputScroll'
+
+const TextInputFormItem = () => (
+  <TextInput role="input" type="number" {...usePreventChangeOnNumberInputScroll()} />
+)
+
+describe('usePreventChangeOnNumberInputScroll', () => {
+  it('Should focus on input that is already in focus', async () => {
+    const { getByRole } = render(<TextInputFormItem />)
+
+    const numberInput = getByRole('input')
+
+    numberInput.blur = jest.fn()
+    const focusMock = jest.spyOn(numberInput, 'focus')
+
+    act(() => {
+      fireEvent.focus(numberInput)
+      fireEvent.wheel(numberInput)
+    })
+
+    expect(numberInput.blur).toHaveBeenCalled()
+    await waitFor(() => expect(focusMock).toHaveBeenCalled())
+  })
+
+  it('Should not focus on input that is not focused', async () => {
+    const { getByRole } = render(<TextInputFormItem />)
+
+    const numberInput = getByRole('input')
+
+    numberInput.blur = jest.fn()
+    const focusMock = jest.spyOn(numberInput, 'focus')
+
+    act(() => {
+      fireEvent.wheel(numberInput)
+    })
+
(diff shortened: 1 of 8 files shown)


src/app/lib/__tests__/preventChangeOnNumberInputScroll.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+  numberInput.blur = jest.fn()
+  const focusMock = jest.spyOn(numberInput, 'focus')
+
+  act(() => {
+    fireEvent.wheel(numberInput)
+  })
+
+  expect(numberInput.blur).toHaveBeenCalled()
+  await waitFor(() => expect(focusMock).toHaveBeenCalled())
+})
```
lukaw3d, line 21: This tests implementation details instead of the behavior

src/app/lib/preventChangeOnNumberInputScroll.ts
```diff
@@ -0,0 +1,19 @@
...
+
+      // Prevents input value change
+      target.blur()
+      // Prevents scrolling
+      e.stopPropagation()
+
+      // Focus after blur
+      setTimeout(() => {
+        target.focus()
+      }, 0)
```
lukaw3d, line 16: Unintended sideffect: input becomes focused when scrolling over it

src/app/lib/__tests__/preventChangeOnNumberInputScroll.test.tsx
lubej: Right, the right implementation would be:
- set value to x
- trigger wheel event scroll
- check if value equal to x

But the issue is that jest probably mocks wheel event, at least I couldn't make it work. So I settled on this - which would check if the implementation calls went through for the on wheel event.

src/app/lib/preventChangeOnNumberInputScroll.ts
```diff
@@ -0,0 +1,19 @@
+import { TextInputProps } from 'grommet/es6/components/TextInput'
+
+export const preventChangeOnNumberInputScroll = (): Pick<TextInputProps, 'onWheel'> => {
+  return {
+    onWheel: e => {
```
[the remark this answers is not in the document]
lubej: https://github.com/oasisprotocol/oasis-wallet-web/pull/1964

I added `.preventDefault()` implementation with global listener. It behaves differently. Since focus is captured, you won't be able to scroll on top of the input[type="number"] field. Which makes it a bit awkward IMO. The current implementation will let you scroll the whole page while hovering over input[type="number"] field.

src/app/lib/preventChangeOnNumberInputScroll.ts
lubej: So I checked, wheel event does not get triggered on mobile. So I added an implementation of global listener here https://github.com/oasisprotocol/oasis-wallet-web/pull/1964.

I would expect the input[type="field"] to not lose focus, while scrolling on it. But you can try global listener implementation with `preventDefault()` that would prevent scrolling at all - if that implementation is more intuitive for you. Or we could simply remove `target.focus()`, but that way it would always lose focus, when scrolling over the input[type="number"] field.

src/app/lib/__tests__/preventChangeOnNumberInputScroll.test.tsx
lukaw3d: here's an e2e test in playwright that fails on master:
```ts
  const input = page.getByPlaceholder('Enter an amount')
  await input.click()
  await input.fill('1111')
  const inputPosition = await input.boundingBox()
  await page.mouse.move(
    inputPosition!.x + inputPosition!.width / 2,
    inputPosition!.y + inputPosition!.height / 2,
  )
  await page.mouse.wheel(0, 10)
  await page.mouse.wheel(0, 10)
  await page.mouse.wheel(0, 10)
  await expect(input).toHaveValue('1111')
```

src/app/lib/preventChangeOnNumberInputScroll.ts
lubej: Ok, this has been resolved now, if input is focused, it will keep focus. Otherwise focus won't be captured on un-focused field. But feel like implementation is far from ideal, and I think I would rather go with https://github.com/oasisprotocol/oasis-wallet-web/pull/1964 for simplicity sake - even though it does not work the same as this implementation.

src/app/lib/usePreventChangeOnNumberInputScroll.ts
```diff
@@ -0,0 +1,47 @@
...
+      onFocus?.(e)
+
+      isFocused.current = true
+    },
+    onMouseOut: e => {
+      onMouseOut?.(e)
+
+      if (document.activeElement === inputRef?.current) {
+        isFocused.current = false
+      }
```
lukaw3d, line 44: Is there a performance reason for doing this instead of `onWheel: document.activeElement !== target`?

src/app/lib/usePreventChangeOnNumberInputScroll.ts
```diff
@@ -0,0 +1,47 @@
...
+    },
+  }
+}
```
lukaw3d, line 47: This implementation loses focus, quite often
Peek 2024-05-31 23-27

lukaw3d: Fix and simplify with
```suggestion
export const usePreventChangeOnNumberInputScroll = (): Pick<TextInputProps, 'onWheel'> => {
  return {
    onWheel: e => {
      const target = e.target as HTMLElement
      if (target !== document.activeElement) return

      // Prevents input value change
      target.blur()
      // Focus after blur
      setTimeout(() => {
        target.focus()
      }, 0)
    },
  }
}
```

lubej: Yeah, looked into detail and `onMouseOut` does not work perfectly. Initially I used `onBlur` event, which should be the optimal way to check if focus is lost. But since I was blurring the element inside `onWheel` to prevent value change, it did not work as expected.

Either way, your solution works. The only issue I have is an unit test, where after wheel event the `activeElement` becomes `body` elements. But seems to work fine in browser environment.

Thanks for taking the time, to look for better solution 👍

src/app/lib/__tests__/usePreventChangeOnNumberInputScroll.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+    const numberInput = getByRole('input')
+    fireEvent.change(numberInput, { target: { value: '123' } })
+
+    act(() => {
+      fireEvent.focus(numberInput)
+      fireEvent.wheel(numberInput)
+    })
+
+    // Focuses on body after wheel change, where in actual browser focuses back on the input
+    // expect(numberInput).toHaveFocus()
```
lukaw3d, line 22: This sounds like it should work. Does it just need to wait a bit?

lukaw3d (approved):

src/app/lib/__tests__/usePreventChangeOnNumberInputScroll.test.tsx
lubej: Yes, good point.

PR #1957, opened 2024-05-28
Upgrade Capacitor to v6

buberdds:
- Gradle Plugin 8.2 is required so minimum Android Studio version is Hedgehog https://developer.android.com/studio/releases#android_gradle_plugin_and_android_studio_compatibility
- Capacitor 6 sets a deployment target of iOS 13 and Android 14 (SDK 34)

APK seems Ok after updates. I am not fully able to test Ledger though. That part was not working for me earlier too (I can select BLE device, but cannot list accounts - shows USB Transport error).

### DIFF
--- a/android/variables.gradle
+++ b/android/variables.gradle
@@ -1,14 +1,14 @@
 ext {
     minSdkVersion = 22
-    compileSdkVersion = 33
-    targetSdkVersion = 33
-    androidxActivityVersion = '1.7.0'
+    compileSdkVersion = 34
+    targetSdkVersion = 34
+    androidxActivityVersion = '1.8.0'
     androidxAppCompatVersion = '1.6.1'
     androidxCoordinatorLayoutVersion = '1.2.0'
-    androidxCoreVersion = '1.10.0'
-    androidxFragmentVersion = '1.5.6'
-    coreSplashScreenVersion = '1.0.0'
-    androidxWebkitVersion = '1.6.1'
+    androidxCoreVersion = '1.12.0'
+    androidxFragmentVersion = '1.6.2'
+    coreSplashScreenVersion = '1.0.1'
+    androidxWebkitVersion = '1.9.0'
     junitVersion = '4.13.2'
     androidxJunitVersion = '1.1.5'
     androidxEspressoCoreVersion = '3.5.1'
--- a/package.json
+++ b/package.json
@@ -58,18 +58,18 @@
     "report-dir": "cypress-coverage"
   },
   "dependencies": {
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -58,12 +58,13 @@
...
-    "@capacitor/app": "5.0.5",
-    "@capacitor/core": "5.0.4",
-    "@capacitor/ios": "5.0.4",
+    "@capacitor-community/bluetooth-le": "6.0.0",
+    "@capacitor/android": "^6.0.0",
+    "@capacitor/app": "^6.0.0",
+    "@capacitor/core": "^6.0.0",
+    "@capacitor/ios": "^6.0.0",
     "@ethereumjs/util": "9.0.3",
+    "@ledgerhq/hw-transport": "^6.30.1",
```
lubej, line 67: How come this dependency was added?

lubej:
I can confirm connecting & listing accounts works for me for Bluetooth Ledger for this dependency updates. 

@buberdds can you check permissions for the application? And is you Nano X paired with the phone?
I can't think of any other reason, why it wouldn't work.

buberdds:
Yes it is paired otherwise I would not be able to see it on "Select Ledger Device"  BLE list. Maybe I messed with developer settings. Anyway will try using different phone today.

package.json
buberdds: the reason is that `npx cap migrate` internally calls `yarn upgrade` https://github.com/ionic-team/capacitor/blob/main/cli/src/tasks/migrate.ts#L428C45-L428C52. Adding this dep is not the end of a world because we are importing that dependency in our code base, but I don't want to use `yarn upgrade` anywhere. Will do migration "more manually".

lubej (approved):

android/variables.gradle
```diff
@@ -1,14 +1,14 @@
 ext {
     minSdkVersion = 22
-    compileSdkVersion = 33
-    targetSdkVersion = 33
```
lubej, line 4: Saw that in the migration guide, this bump is necessary, but wonder if we can distribute it for lower version. I guess lets get it released for 34, and then try to do a lower version. Or is `minSdkVersion` for that? 🤔

buberdds: after reading about it - yes, that's how it works. You compile using SDK XYZ while still being backward-compatible with the minSdkVersion. In this case it is Android 5.1 or something. We can bump minSdkVersion to exclude old devices. This can be done via Google Play Console too.

lubej: Nice 👍

PR #1964, opened 2024-05-30
Prevent value change on number input wheel scroll globally

lubej:
Addresses https://github.com/oasisprotocol/oasis-wallet-web/issues/1955


lubej:
Closed in favor of https://github.com/oasisprotocol/oasis-wallet-web/pull/1956

PR #1965, opened 2024-05-30
Migrate from grpc.oasis.dev to grpc.oasis.io

lukaw3d:
Heads up: this changes security headers

### DIFF
--- a/internals/getSecurityHeaders.js
+++ b/internals/getSecurityHeaders.js
@@ -45,8 +45,8 @@ const getCsp = ({ isExtension, isDev }) =>
       'self';
     connect-src
       'self'
-      https://grpc.oasis.dev
-      https://testnet.grpc.oasis.dev
+      https://grpc.oasis.io
+      https://testnet.grpc.oasis.io
       https://api.oasisscan.com
       ${isDev ? localnet : ''}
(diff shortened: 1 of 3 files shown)


internals/getSecurityHeaders.js
```diff
@@ -45,10 +45,9 @@ const getCsp = ({ isExtension, isDev }) =>
       'self';
     connect-src
       'self'
-      https://grpc.oasis.dev
-      https://testnet.grpc.oasis.dev
+      https://grpc.oasis.io
+      https://testnet.grpc.oasis.io
```
buberdds, line 49: just to get rid of this value can we bump test-inputs too?

buberdds (approved):

buberdds:
Added notes for devops https://github.com/oasisprotocol/internal-ops/pull/257

internals/getSecurityHeaders.js
lukaw3d: no. If we ever modify our migration code to also use old extension's NETWORK_CONFIG, then user's data will contain grpc.oasis.dev. And if we don't modify it to .io it will get blocked by CSP and tests might catch that mistake

PR #1967, opened 2024-06-01
Quick fix for updating transactions list by adding more delay

lukaw3d:
For https://github.com/oasisprotocol/oasis-wallet-web/issues/1942

### DIFF
--- a/src/app/state/account/saga.ts
+++ b/src/app/state/account/saga.ts
@@ -15,6 +15,7 @@ import { getAccountBalanceWithFallback } from '../../lib/getAccountBalanceWithFa
 import { walletActions } from '../wallet'
 
 const ACCOUNT_REFETCHING_INTERVAL = process.env.REACT_APP_E2E_TEST ? 5 * 1000 : 30 * 1000
+const TRANSACTIONS_UPDATE_DELAY = 35 * 1000 // Measured between 8 and 31 second additional delay after balance updates
 const TRANSACTIONS_LIMIT = 20
 
 export function* fetchAccount(action: PayloadAction<string>) {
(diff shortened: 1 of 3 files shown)


src/app/state/account/saga.ts
```diff
@@ -15,6 +15,7 @@ import { getAccountBalanceWithFallback } from '../../lib/getAccountBalanceWithFa
 import { walletActions } from '../wallet'
 
 const ACCOUNT_REFETCHING_INTERVAL = process.env.REACT_APP_E2E_TEST ? 5 * 1000 : 30 * 1000
+const TRANSACTIONS_UPDATE_DELAY = 35 * 1000 // Measured between 8 and 31 second additional delay after balance updates
```
lubej, line 18: Should there be a different delay on mainnet opposed to testnet? As the issue comment suggests, there is only 8sec delay on mainnet. Why should an user wait x3 the time, when it could have been updated faster?

lubej (approved):

src/app/state/account/saga.ts
lukaw3d: Mainnet often takes 20+ seconds

lubej: Can we adjust that for the mainnet? 15sec is still a lot of time.

src/app/state/account/saga.ts
```diff
@@ -130,6 +131,8 @@ export function* fetchingOnAccountPage() {
             staleBalances.delegations !== refreshedAccount.delegations ||
             staleBalances.debonding !== refreshedAccount.debonding
           ) {
+            // Wait for oasisscan to update transactions (it updates balances faster)
+            yield* delay(TRANSACTIONS_UPDATE_DELAY)
```
lubej, line 135: Even though, thinking about this, this may introduce race condition. In a sense that, if you switch account in the mean time while the delay is in progress. It could refresh the "newly" opened account for no reason. So it should check `state.address !== accountAddress`(should be provided from payload action). Will try to work it out in https://github.com/oasisprotocol/oasis-wallet-web/pull/1954

lukaw3d: can you start with a test for this? I don't think it happens

lubej: Yeah, you are right `delay` gets canceled by `takeLatestCancelable`. Never mind. I though it acts as a setTimeout(without it being canceled).

PR #1975, opened 2024-06-17
Refactor AlertBox

lubej:
Introduce content prop on AlertBox for existing styling/formatting. Replace children prop, to support custom formatting.

### DIFF
--- a/src/app/components/AlertBox/index.tsx
+++ b/src/app/components/AlertBox/index.tsx
@@ -2,18 +2,26 @@ import { Box } from 'grommet/es6/components/Box'
 import { Text } from 'grommet/es6/components/Text'
 import { normalizeColor } from 'grommet/es6/utils'
 import * as React from 'react'
-import { useContext } from 'react'
+import { ReactElement, useContext } from 'react'
 import { ThemeContext } from 'styled-components'
 
 export type AlertBoxStatus = 'error' | 'warning' | 'warning-weak' | 'ok' | 'ok-weak' | 'info'
 
-interface Props {
+type Props = {
   status: AlertBoxStatus
   center?: boolean
-  /** Example `icon={<Info size="20px" />}` */
-  icon?: React.ReactNode
-  children: React.ReactNode
-}
+} & (
+  | {
+      content: React.ReactNode
+      icon?: React.ReactNode
+      children?: undefined
+    }
+  | {
+      children: ReactElement | Iterable<React.ReactNode>
+      content?: undefined
+      icon?: undefined
+    }
+)
 
 const mapStatus = {
   error: {
@@ -61,12 +69,15 @@ export function AlertBox(props: Props) {
       background={{ color: mapStatus[props.status].background }}
       pad={{ horizontal: 'small' }}
     >
-      <Box direction="row" gap="small" align="center" justify={props.center ? 'center' : 'start'}>
-        {props.icon && <Text color={color}>{props.icon}</Text>}
-        <Text weight="bold" size="12px" style={{ marginTop: 10, marginBottom: 10 }}>
-          {props.children}
-        </Text>
-      </Box>
+      {!props.children && props.content && (
+        <Box direction="row" gap="small" align="center" justify={props.center ? 'center' : 'start'}>
+          {props.icon && <Text color={color}>{props.icon}</Text>}
+          <Text weight="bold" size="12px" style={{ marginTop: 10, marginBottom: 10 }}>
+            {props.content}
+          </Text>
+        </Box>
+      )}
+      {props.children}
     </Box>
   )
 }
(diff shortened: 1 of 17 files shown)


src/app/components/AlertBox/index.tsx
```diff
@@ -12,7 +12,7 @@ interface Props {
   center?: boolean
   /** Example `icon={<Info size="20px" />}` */
   icon?: React.ReactNode
-  children: React.ReactNode
+  content: React.ReactNode
```
buberdds, line 15: This prop should be now optional

src/app/components/AlertBox/index.tsx
```diff
@@ -61,12 +61,16 @@ export function AlertBox(props: Props) {
       background={{ color: mapStatus[props.status].background }}
       pad={{ horizontal: 'small' }}
     >
-      <Box direction="row" gap="small" align="center" justify={props.center ? 'center' : 'start'}>
-        {props.icon && <Text color={color}>{props.icon}</Text>}
-        <Text weight="bold" size="12px" style={{ marginTop: 10, marginBottom: 10 }}>
-          {props.children}
-        </Text>
-      </Box>
+      {!props.children && (props.icon || props.content) && (
```
buberdds, line 64: We should remove props.icon from this condition as we allow to render alert box with icon only which is useless.

src/app/components/AlertBox/index.tsx
```diff
@@ -61,12 +61,16 @@ export function AlertBox(props: Props) {
...
+        <Box direction="row" gap="small" align="center" justify={props.center ? 'center' : 'start'}>
+          {props.icon && <Text color={color}>{props.icon}</Text>}
+          <Text weight="bold" size="12px" style={{ marginTop: 10, marginBottom: 10 }}>
+            {props.content}
+          </Text>
+        </Box>
+      )}
+      {/* eslint-disable-next-line no-restricted-syntax */}
```
buberdds, line 72: Why are you disabling lint rule here? you can pass children as a text node <>Foo</> and Google Translate won't work, right?

src/app/components/AlertBox/index.tsx
buberdds, line 64: Would be useful to indicate this relation between content and children via types like 
```
type Props = {
  center?: boolean
  status: AlertBoxStatus
} & (
  | {
      content: React.ReactNode
      icon?: React.ReactNode
      children?: undefined
    }
  | {
      children: React.ReactNode
      content?: undefined
      icon?: undefined
    }
)
```

src/app/components/AlertBox/index.tsx
lubej: These are custom rules, it does not work nicely with "dynamic" content. I marked children as `ReactElement | Iterable<React.ReactNode>`, so no direct `string` is allowed. As for why, Google translate changes the structure of the elements, and that is the reason why text should be wrapped inside some element, to not break layout. As far as I know Google Translate will still work.

src/app/components/AlertBox/index.tsx
```diff
@@ -61,12 +69,16 @@ export function AlertBox(props: Props) {
       background={{ color: mapStatus[props.status].background }}
       pad={{ horizontal: 'small' }}
     >
-      <Box direction="row" gap="small" align="center" justify={props.center ? 'center' : 'start'}>
-        {props.icon && <Text color={color}>{props.icon}</Text>}
-        <Text weight="bold" size="12px" style={{ marginTop: 10, marginBottom: 10 }}>
-          {props.children}
-        </Text>
-      </Box>
+      {!props.children && props.content && (
```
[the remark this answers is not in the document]
lubej: With pending txs I was trying to do a bit more "custom" AlertBox:
https://github.com/oasisprotocol/oasis-wallet-web/blob/62c1711b8b07c90b7f386e6081ba667086048f9a/src/app/pages/AccountPage/Features/TransactionHistory/index.tsx#L61-L88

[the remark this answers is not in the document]
lubej: But yes, I agree on `content` naming. Was thinking of naming it `label` first, but that felt counterintuitive as well, as the `label` doesn't have to be only text.

src/app/components/AlertBox/index.tsx
buberdds: why we can't just write  `{props.children}`. Why we need ` {props.children && <>{props.children}</>}` ?

lubej: Yeah, good point. Now that props are "safeguarded" by TypeScript, should be fine that way.

src/app/components/AlertBox/index.tsx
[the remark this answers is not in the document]
lubej: localhost_3000_account_oasis1qrfwkqg4n3fd3xs8msljdhyuy0rr34289vcntjft

lubej:
Closing in favor of https://github.com/oasisprotocol/oasis-wallet-web/pull/1978

Issue #1972, opened 2024-06-11
Mobile ux issue: if user first opens the app without connection, then it displays RPC error forever


lukaw3d:
Almost the same as https://github.com/oasisprotocol/oasis-wallet-web/issues/1853

User cant refresh the page after they have connection. They have to kill the app

Issue #1880, opened 2024-04-09
E2E test "recover from fatal errors" is flaky (at least locally)


lukaw3d:
```console
luka@luka:oasis-wallet-web/playwright$ xvfb-run yarn test
...
1) [chromium] › extension.spec.ts:76:7 › The extension popup should load › recover from fatal errors 

  Error: locator.click: Target page, context or browser has been closed
  Call log:
    - waiting for getByRole('button', { name: 'Reload extension' })
    -   locator resolved to <button type="button" class="StyledButton-sc-323bzc-0 jG…>…</button>
    - attempting click action
    -   waiting for element to be visible, enabled and stable
    -     element is not stable - waiting...
    -   element is visible, enabled and stable
    -   scrolling into view if needed
    -   done scrolling
    -   performing click action


      97 |       const page = await context.newPage()
      98 |       await page.goto(`${extensionPopupURL}/`)
  >  99 |       await page.getByRole('button', { name: 'Reload extension' }).click()
        |                                                                    ^
    100 |       await page.close()
    101 |     }
    102 |

      at oasis-wallet-web/playwright/tests/extension.spec.ts:99:68

1 failed
  [chromium] › extension.spec.ts:76:7 › The extension popup should load › recover from fatal errors 

luka@luka:oasis-wallet-web/playwright$ xvfb-run yarn test extension
Running 5 tests using 4 workers

  ✓  1 …sion.spec.ts:15:7 › The extension popup should load › should successfully load javascript chunks (3.0s)
  ✓  2 ….ts:21:7 › The extension popup should load › get state from background page through webext-redux (4.8s)
  ✓  3 …extension.spec.ts:45:7 › The extension popup should load › should allow embedded Transak widget (12.2s)
  ✓  4 …tension.spec.ts:33:7 › The extension popup should load › ask for USB permissions in ledger popup (4.0s)
Slow API 551.586ms chrome-extension://iakeicaignipfamnkkfhojgdkokpgofm/rubik-v20-latin-ext_latin-300.a151c08f.woff2
  ✘  5 [chromium] › extension.spec.ts:76:7 › The extension popup should load › recover from fatal errors (4.5s)
Slow API 3359.544ms https://assets-stg.transak.com/static/index-TKnnq8QD.js
[...]

PR #1977, opened 2024-06-19
Always show Reload button on fatal error (for recovering in mobile app)

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/1972
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/1880

### DIFF
--- a/src/app/components/FatalErrorHandler/index.tsx
+++ b/src/app/components/FatalErrorHandler/index.tsx
@@ -104,15 +104,16 @@ export function FatalErrorHandler({ children }: Props) {
             }
           />
         </Box>
-        {isExtension && (
-          <Box align={isMobile ? 'stretch' : 'end'} margin={{ bottom: isMobile ? 'xlarge' : 'large' }}>
-            <BrandButton
-              onClick={() => (window as any).chrome?.runtime?.reload()}
-              icon={<Refresh />}
-              label={t('fatalError.reloadExtension', 'Reload extension')}
-            />
-          </Box>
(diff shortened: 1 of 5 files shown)


src/app/components/FatalErrorHandler/index.tsx
```diff
@@ -104,15 +104,16 @@ export function FatalErrorHandler({ children }: Props) {
...
-              onClick={() => (window as any).chrome?.runtime?.reload()}
-              icon={<Refresh />}
-              label={t('fatalError.reloadExtension', 'Reload extension')}
-            />
-          </Box>
-        )}
+        <Box align={isMobile ? 'stretch' : 'end'} margin={{ bottom: isMobile ? 'xlarge' : 'large' }}>
+          <BrandButton
+            onClick={() => {
+              ;(window as any).chrome?.runtime?.reload()
```
lubej, line 110: Any reason for this semicolon here?

lubej (approved):

src/app/components/FatalErrorHandler/index.tsx
lukaw3d: Prettier automatically adds it. It's meant to prevent concatenating this line with previous one

e.g. if lines were switched and it didn't add it:
```ts
window.location.reload()
(window as any).chrome?.runtime?.reload()
```
executes as `window.location.reload()(window).chrome?.runtime?.reload()`

PR #1978, opened 2024-06-19
Add endSlot prop in AlertBox component

lubej:


src/app/pages/AccountPage/Features/TransactionHistory/index.tsx
```diff
@@ -35,6 +39,27 @@ export function TransactionHistory(props: Props) {
...
+              style={{ verticalAlign: 'middle' }}
+            >
+              <Text size="xsmall">{t('account.summary.viewAccountTxs', 'View account transactions')}</Text>
+              <FormNext color="brand" size="18px" />
+            </Anchor>
+          </span>
+        }
+      >
+        {t('account.summary.someTxsInPendingState', 'Some transactions are currently in a pending state.')}
+      </AlertBox>
```
lubej, line 61: Satisfied my use case in #1954 
image

lubej:
Will remove [Add sample usage - REVERT](https://github.com/oasisprotocol/oasis-wallet-web/pull/1978/commits/e6fac60a1b007bea1a557e2c31f09149537f4434) and update snapshots after initial review.

src/app/components/AlertBox/index.tsx
```diff
@@ -61,11 +62,18 @@ export function AlertBox(props: Props) {
...
     >
-      <Box direction="row" gap="small" align="center" justify={props.center ? 'center' : 'start'}>
-        {props.icon && <Text color={color}>{props.icon}</Text>}
-        <Text weight="bold" size="12px" style={{ marginTop: 10, marginBottom: 10 }}>
-          {props.children}
-        </Text>
+      <Box
+        direction="row"
+        justify={props.justify ?? 'start'}
+        pad={{ horizontal: 'none', vertical: 'xsmall' }}
```
buberdds, line 68: horizontal is not needed

.changelog/1978.internal.md
```diff
@@ -0,0 +1 @@
+Add endSlot prop in AlertBox component
```
buberdds, line 1: `internal` is used for builds, docs, refactor etc. Simple tasks that are reflected in UI are using `trivial` type.

src/app/components/AlertBox/index.tsx
```diff
@@ -61,11 +62,18 @@ export function AlertBox(props: Props) {
...
+      >
+        <Box direction="row" gap="small" align="center">
+          {props.icon && <Text color={color}>{props.icon}</Text>}
+          <Text size="xsmall" weight="bold" style={{ verticalAlign: 'middle' }}>
```
buberdds, line 72: This is not needed as parent flexbox container takes care of alignment, right?

lubej: Since I replaced `size="12px"` with `size="xsmall"`, the later adds `line-height` props. Thats why I wanted to make it centered - otherwise looks weird with different vertical paddings inside alert. It is not ideal, as you would have to do the same on the `endSlot`, but I just disliked that `size` was hardcoded.

buberdds (approved):

PR #1979, opened 2024-06-19
Add transaction detail API cache

lubej:
Adds cache for the following 2 endpoints:
- getTransaction (`/chain/transaction/{hash}`)
- getRuntimeTransactionInfo (`/runtime/transaction/info?id={paraTimeId}&hash={hash}`)

As those responses won't change, once indexed, they are cached through the app session(until user closes the wallet).

### DIFF
--- a/src/vendors/oasisscan.ts
+++ b/src/vendors/oasisscan.ts
@@ -22,6 +22,9 @@ import {
 
 import { throwAPIErrors } from './helpers'
 
+const getTransactionCacheMap: Record<string, Transaction> = {}
+const getRuntimeTransactionInfoCacheMap: Record<string, RuntimeTransactionInfoRow> = {}
+
 export function getOasisscanAPIs(url: string | 'https://api.oasisscan.com/mainnet') {
   const explorerConfig = new Configuration({
     basePath: url,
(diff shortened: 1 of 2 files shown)


src/vendors/oasisscan.ts
```diff
@@ -54,11 +56,20 @@ export function getOasisscanAPIs(url: string | 'https://api.oasisscan.com/mainne
...
+    }
+
     const transaction = await operationsEntity.getTransaction({
       hash,
     })
 
-    const [parsedTx] = parseTransactionsList([transaction.data])
+    const [parsedTx] = parseTransactionsList([transaction.data ?? {}])
+
+    if (transaction.code === 0 && !!transaction.data) {
```
lukaw3d, line 69: Comment that oasisscan returns {"code":0,"data":null} for missing or unprocessed transactions

src/vendors/oasisscan.ts
```diff
@@ -54,11 +56,20 @@ export function getOasisscanAPIs(url: string | 'https://api.oasisscan.com/mainne
   }
 
   async function getTransaction({ hash }: { hash: string }) {
+    if (hash in getTransactionCacheMap) {
+      return getTransactionCacheMap[hash]
```
lukaw3d, line 60: Please add to `getRuntimeTransactionInfo` too

lukaw3d (approved):

src/vendors/oasisscan.ts
```diff
@@ -22,6 +22,8 @@ import {
 
 import { throwAPIErrors } from './helpers'
 
+const getTransactionCacheMap: Record<string, Transaction> = {}
+
 export function getOasisscanAPIs(url: string | 'https://api.oasisscan.com/mainnet') {
```
lukaw3d, line 27: Hm, need to differentiate testnet from mainnet. You can add `url` to be part of the key

lubej: Yes, good point.

PR #2001, opened 2024-07-11
Simplify appending nonces to consensus transactions

lukaw3d:
Refactors #1954 and #1999.

also deduplicated changelog entries


src/vendors/oasisscan.ts
```diff
@@ -108,7 +106,14 @@ export function getOasisscanAPIs(url: string | 'https://api.oasisscan.com/mainne
...
     const list = await Promise.all(
-      transactionsList.data.list.map(async tx => (tx.runtimeId ? getRuntimeTransactionInfo(tx) : tx)),
+      transactionsList.data.list.map(async tx => {
+        if (tx.runtimeId) {
+          return await getRuntimeTransactionInfo(tx)
+        } else {
+          const { nonce } = await getTransaction({ hash: tx.txHash })
+          return { ...tx, nonce }
+        }
+      }),
```
lukaw3d, line 116: One difference from before: this fails if any request fails, unlike allSettled. But I think simplifying it is worth it

src/app/state/account/saga.ts
```diff
@@ -56,35 +56,7 @@ export function* fetchAccount(action: PayloadAction<string>) {
...
-              }
-            }
-
-            return t
-          })
-
-          yield* put(
-            accountActions.transactionsLoaded({ networkType, transactions: transactionsWithUpdatedNonce }),
-          )
+          yield* put(accountActions.transactionsLoaded({ networkType, transactions }))
```
lubej, line 59: Yes, good point, this logic should not be here.

src/vendors/oasisscan.ts
lubej: I kind of liked, that it did not fail before if any detailed tx call fails. As it would still work in most cases, but sure, this way there is no flakiness.

lubej (approved):

PR #2005, opened 2024-07-15
Make Total balance font-size equal to other balance fields

lukaw3d:
| Before | After |
| --- | --- |
| image       |     image |


lukaw3d:
Should use right-aligned numbers later

buberdds (approved):

PR #2019, opened 2024-07-27
Fix theme switcher icon size

lukaw3d:
Incorrectly changed in d373356df291f500ad808c30eaefd49a2fa5c698.

### DIFF
--- a/src/app/components/ThemeSwitcher/index.tsx
+++ b/src/app/components/ThemeSwitcher/index.tsx
@@ -18,10 +18,10 @@ import { getTargetTheme } from 'styles/theme/utils'
 
 interface Props {}
 
-const getThemesIcons = (t: TFunction) => ({
-  light: <Sun aria-label={t('theme.lightMode', 'Light mode')} size="24px" />,
-  dark: <Moon aria-label={t('theme.darkMode', 'Dark mode')} size="24px" />,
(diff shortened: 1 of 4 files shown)


src/app/components/ThemeSwitcher/index.tsx
```diff
@@ -19,9 +19,9 @@ import { getTargetTheme } from 'styles/theme/utils'
 interface Props {}
 
 const getThemesIcons = (t: TFunction) => ({
-  light: <Sun aria-label={t('theme.lightMode', 'Light mode')} size="24px" />,
-  dark: <Moon aria-label={t('theme.darkMode', 'Dark mode')} size="24px" />,
-  system: <System aria-label={t('theme.system', 'System')} size="24px" />,
+  light: <Sun aria-label={t('theme.lightMode', 'Light mode')} />,
```
buberdds, line 22: This works for `ThemeSwitcher`, but not for `ThemeSelect` which still needs `24px`. We have theme selection in a sidebar and in the profile settings modal (Grommet Select). Maybe we can add an optional size arg in `getThemesIcons` ?

buberdds (approved):

src/app/components/ThemeSwitcher/index.tsx
lukaw3d: ooh that's why I changed it last time

PR #2020, opened 2024-07-27
Assemble changes for 1.12.0 release

lukaw3d:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "https://raw.githubusercontent.com/lukaw3d/parcel-bundler-json-schemas/main/package_schema.json",
   "name": "@oasisprotocol/rose-wallet",
-  "version": "1.11.0",
+  "version": "1.12.0",
```
buberdds, line 4: As this is an official new extension release, should't we
- manually switch 2.0.0, applies to manifest.json too
- mention unified code base/new ext release in CHANGELOG

cc @tjanez

tjanez: > 1. As this is an official new extension release, should't we
> 
> * manually switch to 2.0.0, applies to manifest.json too
> * mention unified code base/new ext release in CHANGELOG
>   cc @tjanez

I discussed this with @lukaw3d today and I think we could go step-by-step by first releasing all these changes as version 1.12.0. And after that is out for the users and no issues are uncovered we bump the version to 2.0.0 and also release the new extension to the Chrome Web Store.

> 2. CI fix [Exclude manifest.json from towncrier check #2021](https://github.com/oasisprotocol/oasis-wallet-web/pull/2021)

Thanks for that!

tjanez (approved):

Issue #2030, opened 2024-08-06
Delegations view on some accounts is broken


lukaw3d:
https://wallet.dev.oasis.io/account/oasis1qprtzrg97jk0wxnqkhxwyzy5qys47r7alvfl3fcg/stake/active-delegations
![image](https://github.com/user-attachments/assets/5c3b5a04-f560-4fd6-8714-d6c948bd3953)

lukaw3d:
https://api.oasisscan.com/mainnet/chain/account/delegations?size=500&address=oasis1qprtzrg97jk0wxnqkhxwyzy5qys47r7alvfl3fcg
```json
...
      {
        "validatorAddress": "oasis1qqs5wnxvsk009swtt7ehm5fslxve96kczszwt47s",
        "validatorName": null,
        "icon": null,
        "entityAddress": null,
        "shares": "1549017.970963851",
        "amount": "2024435.120714304",
        "active": true
      },
      {
        "validatorAddress": "oasis1qpwrq93z8s9ytu2hfjtqggc9edgwfadzevs3trvm",
        "validatorName": "Munay Network",
        "icon": "https://s3.amazonaws.com/keybase_processed_uploads/e46b46540baf3a06ac7babf28a06db05_360_360.jpg",
        "entityAddress": null,
        "shares": "1281083.498020046",
        "amount": "1534115.76456032",
        "active": true
      },
      {
        "validatorAddress": null,
        "validatorName": null,
        "icon": null,
        "entityAddress": "oasis1qpz6gtyzec4dwq7vxhvseg9yutavyht36qjsyqm2",
        "shares": "1000",
        "amount": "1000",
        "active": false
      },
      {
        "validatorAddress": null,
        "validatorName": null,
        "icon": null,
        "entityAddress": "oasis1qrcc0c2tppgkc4ue55ykaxe2yznr7zlmkg7vw5a0",
        "shares": "1000",
        "amount": "1000",
        "active": false
      },
      {
        "validatorAddress": null,
        "validatorName": null,
        "icon": null,
        "entityAddress": "oasis1qrct7ry2dcmk4ng2lcskzdv26c09xxmm4ytrlzzd",
        "shares": "1000",
        "amount": "1000",
        "active": false
      },
      {
        "validatorAddress": null,
        "validatorName": null,
        "icon": null,
        "entityAddress": "oasis1qrqz3yx3ujewxpuunlhd43xx3dcxwt90rcajfh3p",
[...]

PR #2031, opened 2024-08-06
Fix display of delegations to non-validators

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/2030

https://wallet.dev.oasis.io/account/oasis1qprtzrg97jk0wxnqkhxwyzy5qys47r7alvfl3fcg/stake/active-delegations

| Before | After |
| --- | --- |
| image | image |

- Fixes selecting non-validators. Previously all "undefined" validators were pre-expanded and crashed when collapsing + re-expanding their details
- Prints delegation's address
- Round delegation amounts in the table. Accurate delegation amount is still displayed when expanding
- (we currently don't have a rounding indicator like https://github.com/oasisprotocol/explorer/blob/1651e1a64a4ae9a7743b3fdc78e1e98aa5d1f5b4/src/app/components/RoundedBalance/__tests__/index.test.tsx#L11-L12)

### DIFF
--- a/src/app/pages/StakingPage/Features/DelegationList/__tests__/__snapshots__/ActiveDelegationList.test.tsx.snap
+++ b/src/app/pages/StakingPage/Features/DelegationList/__tests__/__snapshots__/ActiveDelegationList.test.tsx.snap
@@ -592,7 +592,7 @@ exports[`<ActiveDelegationList  /> should match snapshot 1`] = `
                     class="c0"
                     style="display: inline-flex; white-space: nowrap; font-family: "Roboto mono", monospace; letter-spacing: 0;"
                   >
-                    0.0000001
+                    0.00
                   </div>
                   <span
                     class="c19"
--- a/.changelog/2031.bugfix.md
+++ b/.changelog/2031.bugfix.md
@@ -1,0 +1,1 @@
+Fix display of delegations to non-validators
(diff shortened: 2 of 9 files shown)


src/app/pages/StakingPage/Features/DelegationList/__tests__/__snapshots__/ActiveDelegationList.test.tsx.snap
```diff
@@ -592,7 +592,7 @@ exports[`<ActiveDelegationList  /> should match snapshot 1`] = `
                     class="c0"
                     style="display: inline-flex; white-space: nowrap; font-family: "Roboto mono", monospace; letter-spacing: 0;"
                   >
-                    0.0000001
+                    0.00
```
buberdds, line 595: If it's not longer an accurate value maybe it's worth adding it to expanded details as it is different value than "Total escrow"?

buberdds (approved):

src/app/pages/StakingPage/Features/DelegationList/__tests__/__snapshots__/ActiveDelegationList.test.tsx.snap
lukaw3d: ah it is missing in debonding page; delegation had it under the input field. Added as text now too:
image

PR #2033, opened 2024-08-09
Harden github workflow against injection

lukaw3d:


src/locales/de/translation.json
```diff
@@ -129,10 +129,6 @@
     "switchToMainnet": "Wechsel zu Mainnet."
   },
   "footer": {
-    "poweredBy": {
-      "oasismonitor": "Powered by Oasis Monitor API & Oasis gRPC",
-      "oasisscan": "Powered by Oasis Scan API & Oasis gRPC"
-    },
```
lubej, line 135: Not sure this change belongs to this PR.

lubej (approved):

src/locales/de/translation.json
tjanez: Agreed. Was this an unintentional change?

.changelog/2033.trivial.md
```diff
@@ -0,0 +1 @@
+Harden github workflow against injection
```
tjanez, line 1: This could also deserve a mention in the internal section of the CHANGELOG

tjanez (approved):
Thanks for thinking about security!

Issue #1858, opened 2024-03-27
Clicking on some validator fields does not open the validator


lukaw3d:
![validator expand](https://github.com/oasisprotocol/oasis-wallet-web/assets/3758846/4ad93db9-d6b6-4d91-b64e-9e5ff9157d77)

- clicking on address of unnamed validator: doesn't
- clicking on escrow amount: doesn't
- clicking on name of a validator: expands
- clicking on fee amount: expands

appears to be caused by text vs element inside cell

(maybe react-data-table-component v7 would fix this)

PR #2034, opened 2024-08-12
Allow to use any of validator field to expand details

buberdds:
Closes https://github.com/oasisprotocol/oasis-wallet-web/issues/1858

note: there is special `data-tag="allowRowEvents"` attribute in react-data-table-component that we could add to last nodes. For me, it wasn't reliable with svgs (clicking paths when attr is set in svg) and it would require to pass a prop via multiple generic components to get to NoTranslate for example. It is also weird to bind events with data attrs in react. I am going with `onClick` handler here.

### DIFF
--- a/.changelog/2034.trivial.md
+++ b/.changelog/2034.trivial.md
@@ -1,0 +1,1 @@
+Allow to use any of validator field to expand details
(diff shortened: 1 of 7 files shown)


lukaw3d:
Made an alternative implementation: https://github.com/oasisprotocol/oasis-wallet-web/pull/2040

PR #2040, opened 2024-08-17
Fix expanding table rows on non-plaintext cell

lukaw3d:
Fixes https://github.com/oasisprotocol/oasis-wallet-web/issues/1858

Alternative to https://github.com/oasisprotocol/oasis-wallet-web/pull/2034


buberdds (approved):
this is much better

PR #2041, opened 2024-08-19
Don't destroy Transak iframe every time pending transactions refresh

lukaw3d:


buberdds (approved):
this was not happening before pending txs when we were polling balances already?

lukaw3d:
It was, but less frequently. It's triggered here
https://github.com/oasisprotocol/oasis-wallet-web/blob/296d8bd92692ddd5915008668e0ac6cb2e4dd81d/src/app/state/account/saga.ts#L134-L145

src/app/state/account/saga.ts
```diff
@@ -136,7 +136,7 @@ export function* fetchingOnAccountPage() {
             staleBalances.available !== refreshedAccount.available ||
             staleBalances.delegations !== refreshedAccount.delegations ||
             staleBalances.debonding !== refreshedAccount.debonding ||
-            hasPendingTxs
+            hasPendingTxs // TODO: reduce setLoading if reloaded transactions haven't changed
```
lukaw3d, line 139: Added a comment

PR #2076, opened 2024-10-25
Add support for Nexus API as a backend option for Wallet

buberdds:
- needs https://github.com/oasisprotocol/wallet/pull/2078
- will need follow up PR in the near future similar to https://github.com/oasisprotocol/explorer/pull/1530, but Nexus deployments  are stuck in testnet.prd so it's not urgent
- Testing requires .env update
```
REACT_APP_BACKEND=nexus
```
- before switching to backend other than Oasis Scan v1 we should look at https://github.com/oasisprotocol/wallet/issues/2079 first


src/vendors/nexus.ts
```diff
@@ -0,0 +1,311 @@
...
+        mergedTransactions.map(async transaction => {
+          if (transaction.method === ConsensusTxMethod.StakingReclaimEscrow) {
+            const eventsResponse = await api.consensusEventsGet({
+              limit: 1,
+              txHash: transaction.hash,
+              type: ConsensusEventType.StakingEscrowDebondingStart,
+            })
+            const amount = (eventsResponse.events[0].body as { amount?: number })?.amount
+            if (transaction.body) {
+              ;(transaction.body as { amount?: typeof amount }).amount = amount
```
lukaw3d, line 123: (I'd avoid mutating inside .map)

lukaw3d (approved):

PR #2094, opened 2024-11-12
Refactor how extension requests ledger access (don't rely on state sync)

lukaw3d:
Extracted from https://github.com/oasisprotocol/wallet/pull/2084

Related to https://github.com/oasisprotocol/wallet/pull/2084#discussion_r1830380814

Previous flow:
- '#/open-wallet/ledger' has "Grant access to your Ledger" button
- it opens '#/open-wallet/connect-device' in permanent popup as well as navigates self-closing popup to '#/open-wallet/ledger/usb' (both stay open)
- connect-device popup has "Connect Ledger device" button that requests usb permission, selects device, and triggers listing accounts in redux. That is synced into self-closing popup
- when done in connect-device popup, user tries to continue in self-closing popup, but it closes
- when reopening, and navigating to import from ledger again, it skips some steps because it still has ledger listed accounts in redux
- some user actions will clear that store. Then if user navigates to import from ledger again, it will show "Grant access to your Ledger" and popup again

New flow:
- '#/open-wallet/ledger' has "Grant access to your Ledger" button if it has not been granted permission before
- it opens '#/open-wallet/connect-device' in permanent popup
- connect-device popup has "Connect Ledger device" button that requests usb permission, and selects device. This already gives necessary permissions to self-closing popup
- when done in connect-device popup, user tries to continue in self-closing popup, but it closes
- when reopening, and navigating to import from ledger again, it shows "USB Ledger" button this time (same every time user comes back with already granted permissions)

(I don't have a test ledger device right now. Please check if some of these notes are incorrect)

### DIFF
--- a/.changelog/2094.bugfix.md
+++ b/.changelog/2094.bugfix.md
@@ -1,0 +1,1 @@
+Refactor how extension requests ledger access (don't rely on state sync)
--- a/extension/src/ExtLedgerAccessPopup/ExtLedgerAccessPopup.tsx
+++ b/extension/src/ExtLedgerAccessPopup/ExtLedgerAccessPopup.tsx
@@ -1,6 +1,5 @@
-import React, { useState } from 'react'
+import { useState } from 'react'
 import { useTranslation } from 'react-i18next'
-import { useDispatch } from 'react-redux'
 import { Box } from 'grommet/es6/components/Box'
 import { Button } from 'grommet/es6/components/Button'
 import { Spinner } from 'grommet/es6/components/Spinner'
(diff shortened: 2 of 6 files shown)


buberdds:
The Ledger flow is already not ideal, and this change makes it even less user-friendly. I discussed this with @donouwens, and here are some of his suggestions:
 
In "Grant access to your Ledger" window when user connects a device, nothing happens in popup so it is harder for user to understand what action has been completed.

Can we:
1) Automatically close the "Device connected" success popup after a set number of seconds? (we assume losing focus will also close the popup), or
2) Enhance the messaging by adding a brief explanation about what happened once the device is connected?

lukaw3d:
ah, yeah
Maybe its even possible to prevent the popup1 from closing if we re-focus it before closing persistent popup2 on a timeout

buberdds:
what about message passing instead of state sync ? This way we can keep self-closing popup up to date with user actions

lukaw3d:
I'd still rather avoid that complexity

lukaw3d:
> ah, yeah Maybe its even possible to prevent the popup1 from closing if we re-focus it before closing persistent popup2 on a timeout

It seems impossible to prevent popup1 from closing

PR #2097, opened 2024-12-02
Explain Android app signing key configuration and secret management

buberdds:
Move tutorial from https://github.com/oasisprotocol/wallet/pull/1944 to Wallet mobile docs and add notes about certificate validity flag.

### DIFF
--- a/.changelog/2097.doc.md
+++ b/.changelog/2097.doc.md
@@ -1,0 +1,1 @@
+Explain Android app signing key configuration and secret management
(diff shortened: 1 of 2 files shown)


lukaw3d:
If I understand correctly, we can not change the key after we start publishing. Even if we make our process more secure later, we still have to use the same key.

If so, then storing the key in github secrets is too risky. We should treat it like a domain certificate

lukaw3d:
Oh nevermind. This is just for dev version. We'll use aab + Play App Signing for production

lukaw3d (approved):

PR #2102, opened 2024-12-05
Switch dump validators to Nexus

buberdds:
needed for https://github.com/oasisprotocol/wallet/issues/2086

Full API bindings changes are extracted to https://github.com/oasisprotocol/wallet/pull/2107 (not needed for the next release)


src/app/state/staking/saga.test.ts
```diff
@@ -158,21 +158,65 @@ describe('Staking Sagas', () => {
...
+              self_delegation_balance: '619388791323275',
+              self_delegation_shares: '438908447022946',
+            },
+            in_validator_set: true,
+            media: {
+              name: 'stakefish',
+            },
+            node_id: 'SQZZd1wsWXdFsqswUoh6hZtmzu+ejuSnrGeHtgIBJDo=',
+            rank: 2,
+            start_date: '2024-04-11T03:43:47Z',
```
buberdds, line 186: Changes in API binding can be removed when start_date is removed from mocks, but I think it's worth keeping more valid bindings anyway. With old generated types are mark as Date which is not true, these props are strings.

src/app/state/staking/saga.test.ts
```diff
@@ -207,8 +251,8 @@ describe('Staking Sagas', () => {
               list: parseValidatorsList(
                 getMainnetDumpValidatorsMock.list.map(({ _expectedStatus, ...v }) => ({
```
buberdds, line 252: Test is passing, but I am not 100% sure about this. Test implements logic that is in saga?

lubej (approved):

PR #2118, opened 2025-01-24
Refactor how extension requests ledger access (don't rely on state sync) and close popup

lukaw3d:
Supersedes https://github.com/oasisprotocol/wallet/pull/2094

---

Extracted from https://github.com/oasisprotocol/wallet/pull/2084

Related to https://github.com/oasisprotocol/wallet/pull/2084#discussion_r1830380814

Previous flow:
- '#/open-wallet/ledger' has "Grant access to your Ledger" button
- it opens '#/open-wallet/connect-device' in permanent popup as well as navigates self-closing popup to '#/open-wallet/ledger/usb' (both stay open)
- connect-device popup has "Connect Ledger device" button that requests usb permission, selects device, and triggers listing accounts in redux. That is synced into self-closing popup
- when done in connect-device popup, user tries to continue in self-closing popup, but it closes
- when reopening, and navigating to import from ledger again, it skips some steps because it still has ledger listed accounts in redux
- some user actions will clear that store. Then if user navigates to import from ledger again, it will show "Grant access to your Ledger" and popup again

New flow:
- '#/open-wallet/ledger' has "Grant access to your Ledger" button if it has not been granted permission before
- it opens '#/open-wallet/connect-device' in permanent popup
- connect-device popup has "Connect Ledger device" button that requests usb permission, and selects device. This already gives necessary permissions to self-closing popup
- when done in connect-device popup, user tries to continue in self-closing popup, but it closes
- when reopening, and navigating to import from ledger again, it shows "USB Ledger" button this time (same every time user comes back with already granted permissions)

### DIFF
--- a/src/app/useRouteRedirects.tsx
+++ b/src/app/useRouteRedirects.tsx
@@ -9,7 +9,17 @@ export const useRouteRedirects = () => {
   const navigate = useNavigate()
 
   useEffect(() => {
-    if (address) {
+    const hasRecentlyGrantedUsbAccess =
+      Date.now() - parseInt(window.localStorage.getItem('oasis_wallet_granted_usb_ledger_timestamp') ?? '0') <
+      5 * 60 * 1000
+
+    if (hasRecentlyGrantedUsbAccess) {
+      // When we implement auto-locking, this won't correctly redirect if wallet
+      // gets locked within 5min (useEffect called with address=undefined before unlocking).
+      navigate(`/open-wallet/ledger`)
+      // Delay is needed as workaround for React.StrictMode.
+      setTimeout(() => window.localStorage.removeItem('oasis_wallet_granted_usb_ledger_timestamp'), 1000)
+    } else if (address) {
       navigate(`/account/${address}`)
     }
   }, [address, navigate])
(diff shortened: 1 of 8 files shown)


lukaw3d:
This is based only on the first two commits of https://github.com/oasisprotocol/wallet/pull/2094. Here's the small diff between those two (rebased) and this PR: https://github.com/oasisprotocol/wallet/compare/966dbf312e67564246dd11019ff4d735f9703b7a..f47c85967c540108f2e07d9e5eebaf0e5ac13dbf

src/app/useRouteRedirects.tsx
```diff
@@ -9,7 +9,17 @@ export const useRouteRedirects = () => {
   const navigate = useNavigate()
 
   useEffect(() => {
-    if (address) {
+    const hasRecentlyGrantedUsbAccess =
```
buberdds, line 12: would try `chrome.runtime.sendMessage`  and `chrome.runtime.onMessage.addListener` instead of this logic in hook

buberdds (approved):
translations issues addressed in https://github.com/oasisprotocol/wallet/pull/2119

src/app/useRouteRedirects.tsx
lukaw3d: For that, mv3 flow without service worker would have to be:
- keep ConnectDevice popup open
- click wallet to open default popup
- then default popup can listen to messages

PR #2084, opened 2024-11-02
Prototype manifest v3 extension

lukaw3d:
Closes https://github.com/oasisprotocol/wallet/issues/2073

Prototyped a few options with small increases in complexity:
1. Always open extension wallet in one tab
  - if you close the tab you need to type in the password again
2. Always open extension wallet in one persistent popup
  - if you close the popup you need to type in the password again

3.+4.: Same behavior as manifest v2 - auto-unlock when popup opens
  - difference vs v2: opening extension as a tab is blocked, e.g. middle click account link

PR #2120, opened 2025-01-27
Migrate extension wallet to Manifest V3 architecture

lukaw3d:
Supersedes https://github.com/oasisprotocol/wallet/pull/2084

Built on top of https://github.com/oasisprotocol/wallet/pull/2118

Diff vs #2084: https://github.com/oasisprotocol/wallet/compare/ca34c34d91e928ad545445648bd8152bc459c5b9..76eb7677053b22541c332b21f38237882d2cbaeb


playwright/utils/mockApi.ts
```diff
@@ -8,7 +8,7 @@ import type {
 } from '../../src/vendors/nexus/index'
 import { StringifiedBigInt } from '../../src/types/StringifiedBigInt'
 
-export async function mockApi(context: BrowserContext | Page, balance: StringifiedBigInt) {
+export async function mockApi(context: BrowserContext | Page, balance: number | StringifiedBigInt) {
```
buberdds, line 11: oh I did not update all mockApi usage. this is minor and not related to this PR, but `Account['available']` would be more accurate. can bump it once this is one is merged if that works for you.

.changelog/2120.breaking.md
```diff
@@ -0,0 +1,4 @@
+Migrate extension wallet to Manifest V3 architecture
+
+New limitations: extension users must create a profile while importing a wallet.
```
buberdds, line 3: Expected: 78; Actual: 80

public/manifest.json
```diff
@@ -28,12 +28,16 @@
...
+    "extension_pages": "{{{ EXTENSION_CSP }}}"
+  },
   "background": {
-    "page": "../extension/src/background.html",
-    "persistent": true
+    "service_worker": "service-worker.js"
   },
-  "web_accessible_resources": ["./oasis-xu-frame.html"],
-  "externally_connectable": { "ids": [] }
+  "externally_connectable": { "ids": [] },
```
buberdds, line 38: If we remove this entry error/warning in `chrome://extensions/` will be removed and the end result will be the same (no extensions or apps can connect).

src/app/state/persist/saga.ts
```diff
@@ -211,8 +219,55 @@ function* encryptAndPersistState(action: AnyAction) {
...
+async function readSharedExtMemory() {
+  if (runtimeIs !== 'extension') return
+  const browser = await import('webextension-polyfill')
+  const storage = await browser.storage.session.get('stringifiedEncryptionKey')
+  return storage.stringifiedEncryptionKey as PersistState['stringifiedEncryptionKey']
+}
+
 export function* persistSaga() {
   yield* watchPersistAsync()
+  yield* retainEncryptionKeyBetweenPopupReopenings()
```
buberdds, line 270: Do we need to take into account `unlockedStatus` in this logic? 

For example:
- I am locking my wallet using Profile dropdown -> Profile -> Lock Profile
- Close extension
- Open popup - wallet is open (no UnlockGate)

src/app/state/persist/saga.ts
```diff
@@ -211,8 +219,55 @@ function* encryptAndPersistState(action: AnyAction) {
...
+      }
+    }
+  })
+
+  yield* fork(function* () {
+    const encryptedState = window.localStorage.getItem(
+      STORAGE_FIELD,
+    ) as EncryptedString<PersistedRootState> | null
+    if (!encryptedState) return // Ignore
+    try {
```
lukaw3d, line 242: From michal: does order matter?
does look risky to write before reading the initial state

src/app/state/persist/saga.ts
lukaw3d: Nope, should already have been handled, but chrome has a bug https://github.com/w3c/webextensions/issues/263
so `await browser.storage.session.set({ stringifiedEncryptionKey })` does nothing when undefined

src/app/state/persist/saga.ts
lukaw3d: fixed

lukaw3d: Hm, https://github.com/oasisprotocol/wallet/pull/2120/commits/bfe226e53c00983b6f825e84fc96df3debe5882e breaks v0  migration

lukaw3d: ah I lost one commit during squashing

src/app/state/persist/saga.ts
lukaw3d: Workaround in https://github.com/oasisprotocol/wallet/pull/2120/commits/db49e882699d6e851b46b31274031bd31d1b3f38

buberdds (approved):

PR #2123, opened 2025-02-20
Fix paratime transaction status in the first few seconds

lukaw3d:
Fixes https://github.com/oasisprotocol/wallet/issues/2113

### DIFF
--- a/src/app/state/account/saga.ts
+++ b/src/app/state/account/saga.ts
@@ -90,6 +90,13 @@ export function* refreshAccountOnParaTimeTransaction() {
   while (true) {
     const { payload } = yield* take(transactionActions.paraTimeTransactionSent)
 
+    // Increase the chance to get updated transactions from API.
+    // Note: We can't rely on nonce to detect pending paratime transactions
+    yield* delay(3000)
+    yield* call(refreshAccount, payload)
+    yield* delay(6000)
+    yield* call(refreshAccount, payload)
+    yield* delay(6000)
     yield* call(refreshAccount, payload)
   }
 }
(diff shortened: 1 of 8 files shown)


src/app/state/account/saga.ts
```diff
@@ -90,6 +90,13 @@ export function* refreshAccountOnParaTimeTransaction() {
   while (true) {
     const { payload } = yield* take(transactionActions.paraTimeTransactionSent)
 
+    // Increase the chance to get updated transactions from API.
+    // Note: We can't rely on nonce to detect pending paratime transactions
+    yield* delay(3000)
+    yield* call(refreshAccount, payload)
+    yield* delay(6000)
+    yield* call(refreshAccount, payload)
+    yield* delay(6000)
```
buberdds, line 99: Why we need all these delays? 6000 is to make sure Consensus and ParaTime blocks are finished?

buberdds (approved):
Do we need any additional work based on https://github.com/oasisprotocol/wallet/blob/master/src/app/state/transaction/saga.ts#L240 comment?

lukaw3d:
> Do we need any additional work based on https://github.com/oasisprotocol/wallet/blob/master/src/app/state/transaction/saga.ts#L240 comment?

Yep, but that's bit harder

src/app/state/account/saga.ts
lukaw3d: Yes, and hopefully nexus indexes them quite fast.
During deposit flow: we receive allowance tx after the first delay already, but usually not deposit tx yet

PR #2128, opened 2025-02-25
Show ParaTime transactions in pending transactions section

lukaw3d:
Fixes https://github.com/oasisprotocol/wallet/pull/2123#pullrequestreview-2640345847

image

### DIFF
--- a/src/app/state/transaction/saga.ts
+++ b/src/app/state/transaction/saga.ts
@@ -1,7 +1,13 @@
 import { client, misc } from '@oasisprotocol/client'
 import { Signer } from '@oasisprotocol/client/dist/signature'
 import { PayloadAction } from '@reduxjs/toolkit'
-import { hex2uint, isValidAddress, parseRoseStringToBaseUnitString, uint2bigintString } from 'app/lib/helpers'
+import {
+  hex2uint,
+  isValidAddress,
+  parseRoseStringToBaseUnitString,
+  publicKeyToAddress,
+  uint2bigintString,
+} from 'app/lib/helpers'
 import { LedgerSigner } from 'app/lib/ledger'
 import { OasisTransaction, signerFromEthPrivateKey, signerFromPrivateKey, TW } from 'app/lib/transaction'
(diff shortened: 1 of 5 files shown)


src/app/state/transaction/saga.ts
```diff
@@ -294,13 +325,35 @@ export function* submitParaTimeTransaction(runtime: Runtime, transaction: ParaTi
...
-    fromAddress,
+    fromOasisAddress,
     runtime,
   )
 
   yield* call(OasisTransaction.signParaTime, chainContext, paraTimeTransactionSigner as Signer, rtw)
   yield* call(OasisTransaction.submit, nic, rtw)
   yield* put(transactionActions.paraTimeTransactionSent(transaction.recipient))
+
+  const hash = yield* call(hashSignedTransaction, rtw.unverifiedTransaction as any) // TODO use rtw.hash when added
```
lukaw3d, line 336: @oasisprotocol/client-rt doesn't seem to have a way to get runtime transaction hash. This works at least for deposit and withdraw

cc @pro-wh

buberdds:
Is this working as expected? 
Screenshot from 2025-03-05 11-10-07
Screenshot from 2025-03-05 11-09-52

Pending tx jump between two lists

buberdds (approved):
I think an output of Wed Product meeting was that we are keeping it as is.

lukaw3d:
yep yep

notes from meeting:
- Pending Transactions shows submitted transactions from app's state until API returns matching tx hash on backend
- Activity shows transactions from API, but transactions that go across layers have two steps and API backend returns success=undefined on those transactions for a few seconds https://github.com/oasisprotocol/explorer/issues/528

tjanez:
I would prefer if we would distinguish the copy so it would say "In progress" instead of "Pending" to avoid any potential confusion with the "Pending transactions" section and its "Pending" status.

I think we should first implement it on the Nexus side: https://github.com/oasisprotocol/nexus/issues/924#issuecomment-2710457430.

PR #2139, opened 2025-03-07
Remove server-handler dependency

buberdds:
An alternative to bumping https://github.com/oasisprotocol/wallet/pull/2138

### DIFF
--- a/.changelog/2139.internal.md
+++ b/.changelog/2139.internal.md
@@ -1,0 +1,1 @@
+Replace server-handler dependency with Node http module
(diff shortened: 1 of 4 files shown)


lukaw3d:
Bugs in this would be quite scary. I haven't implemented low level nodejs servers before - do you think this is safe?

buberdds:
It's not something that could be used in prod for sure. Isn't `start:prod` just a dummy task to check a bundle? Currently, I wouldn't say it simulates prod env.  Anyway, just an idea. Let's keep `server-handler` for maintenance reasons.

lukaw3d:
Yes, it can only expose your development machine to malicious websites (like recent Vitest UI vulnerability)

PR #2156, opened 2025-03-24
Improve description of allowance change transactions

lukaw3d:
https://wallet.dev.oasis.io/account/oasis1qppz78n4lpnw9knz7d7jz7vvjrt7qd2c6ypsa7me

| Before | After |
| --- | --- |
|  Screenshot from 2025-03-24 22-26-03      |       Screenshot from 2025-03-24 22-11-40 |

### DIFF
--- a/.changelog/2156.bugfix.md
+++ b/.changelog/2156.bugfix.md
@@ -1,0 +1,1 @@
+Improve description of allowance change transactions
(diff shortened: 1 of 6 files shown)


buberdds:
I read "changed +0.02" as increased by, but that's not how allowance change works, right?

lukaw3d:
It actually works like that

buberdds (approved):

lukaw3d:
From sync: if we fetch transaction's event we could show new allowance value, instead of relative amount_change

PR #2180, opened 2025-05-21
Fix importing images in parcel@2.14

lukaw3d:
Broken since https://github.com/oasisprotocol/wallet/pull/2174


lukaw3d:
aaaaaaaaaaa I forgot this breaks jest. That's why we were import directly

related https://github.com/parcel-bundler/parcel/issues/10167

PR #2188, opened 2025-06-06
Fix flashing wrong layout on init

lukaw3d:
Fixes https://github.com/oasisprotocol/wallet/issues/1863


lukaw3d:
playwright test broke because useRouteRedirects now redirects away from e2e before setIsResponsiveContextInitialized(true) lets it to be rendered. Ignoring this until https://github.com/oasisprotocol/wallet/issues/823

buberdds:
What's an issue with a failing test? It's fine when I run it locally using this side branch.

lukaw3d:
I would just ignore this PR and layout issue until https://github.com/oasisprotocol/wallet/issues/823

PR #2214, opened 2025-11-28
Add "Verify address on Ledger" button in settings

lukaw3d:
Fixes https://github.com/oasisprotocol/wallet/issues/2213

Screenshot from 2025-11-28 07-26-02

### DIFF
--- a/src/app/state/wallet/saga.ts
+++ b/src/app/state/wallet/saga.ts
@@ -1,7 +1,7 @@
 import { PayloadAction } from '@reduxjs/toolkit'
 import { hex2uint, publicKeyToAddress, uint2hex } from 'app/lib/helpers'
 import nacl from 'tweetnacl'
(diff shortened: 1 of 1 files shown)


buberdds (approved):

src/app/state/wallet/saga.ts
```diff
@@ -33,6 +35,8 @@ export function* rootWalletSaga() {
 
   // Reload wallet balances if network changes
   yield* takeEvery(networkActions.networkSelected, refreshOnNetworkChange)
+
+  yield* takeEvery(walletActions.verifyAddressOnLedger, verifyAddressOnLedger)
```
lukaw3d: > when click multiple times at verify button it throws error, but probably users won't do this

oops. Switching this to `takeLeading` to fix it. Sometimes parallel commands to ledger cause it to get stuck, so worth preventing it

PR #2215, opened 2025-11-28
Display unhandled transaction methods as generic not unrecognized

lukaw3d:
Makes https://github.com/oasisprotocol/wallet/issues/2175 less scary


buberdds (approved):
snapshot

PR #2231, opened 2025-12-04
Adjust mobile version code value with Google Play versioning boundaries

buberdds:
Related https://github.com/oasisprotocol/wallet/issues/2227 and https://github.com/oasisprotocol/wallet/pull/2229

```
// 2.5.0
versionCode 20050004163 // master
versionCode 205004164 // branch  
            2100000000 // max 
```

### DIFF
--- a/.changelog/2231.internal.md
+++ b/.changelog/2231.internal.md
@@ -1,0 +1,1 @@
+Update internal mobile build versions
--- a/internals/scripts/changelog.js
+++ b/internals/scripts/changelog.js
@@ -57,9 +57,12 @@ execSync(`towncrier build --version ${version}`, { stdio: 'inherit' })
 const gitCommitCount = execSync('git rev-list --count HEAD', { encoding: 'utf8' }).trim()
 const versionPrefix = version
   .split('.')
-  .map((n, i) => (i > 0 ? n.padStart(3, '0') : n))
+  .map((n, i) => (i > 0 ? n.padStart(2, '0') : n))
   .join('')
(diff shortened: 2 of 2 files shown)


lukaw3d:
we have a problem. The greatest value Google Play allows for versionCode is 2100000000. Both of these are too long

lukaw3d:
do we ever need git count?

lukaw3d:
I would rather drop git count than patch number :shrug:

tjanez:
> we have a problem. The greatest value Google Play allows for versionCode is 2100000000. Both of these are too long

Indeed: https://developer.android.com/studio/publish/versioning. Thanks for checking this!

> do we ever need git count?

I think it will be useful because when I upload builds done on `master` or `stable/*` branch, they must also be unique and increasing.

Perhaps we should just limit ourselves to, e.g.:
"2.5.3" -> "20503"+"4 places git commit number" ("205034161" in this case)

With this, we limit the number of minor and patch versions to 99 (this should be perfectly fine) and the number of git commits to 9999 which should be fine for quite some time.

The scheme also works up to major version 20 which I think should be fine for quite some time.

Thoughts?

buberdds:
> do we ever need git count?

yeah prob we don't need a "dynamic part" with release frequency around 4-5 per year. We also don't use stable branches here, right?
`(major*10_000_000)+(minor*100_000)+(patch*1_000)` would work too (or padded)

If we ever need to crate internal test variants we can bump manually by 1 in PR, but this is just theory (release frequency)

